PR #207, opened 2023-02-03
Fix detail routing.

kialam:


AlexSCorey (commented):
I agree 100% this needs to be, and probably should be done soon, but I also think we should do all the route changes at one time if we can. Also, we seem to have a convention conflict where in Credentials we have a component 'CredentialPage' that handles the routes for details and other tabs.  I'm not bothered with 1 implementation over the other, but I would like consistency.

kialam:
Closing in favor of existing /id/details pattern.

PR #264, opened 2023-03-03
Improve action labels on buttons

keithjgrant:
Add resource names to buttons (eg "Add" becomes "Add Access")

### DIFF
--- a/frontend/awx/access/users/Users.tsx
+++ b/frontend/awx/access/users/Users.tsx
@@ -203,7 +203,7 @@ export function AccessTable(props: { url: string }) {
         variant: ButtonVariant.primary,
         icon: PlusIcon,
         label: t('Add users'),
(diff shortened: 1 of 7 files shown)


frontend/awx/access/users/Users.tsx
```diff
@@ -203,15 +203,15 @@ export function AccessTable(props: { url: string }) {
         variant: ButtonVariant.primary,
         icon: PlusIcon,
         label: t('Add users'),
-        shortLabel: t('Add'),
+        shortLabel: t('Add Access'),
```
vidyanambiar, line 206: The `AccessTable` component is not used since we added the new `ResourceAccessList` component. I think I forgot to clean this one up.

vidyanambiar (approved):
👍

PR #270, opened 2023-03-06
Use labels to indicate admin/auditor users

keithjgrant:
Replaces colored text for user types and uses labels instead

Screenshot 2023-03-06 at 12 44 07 PM


vidyanambiar (approved):
lgtm

PR #275, opened 2023-03-08
Adds aria-label to select dialog and bulk action dialog

AlexSCorey:
As title says:

It also eliminates a console warning about accessibility for those dialogs


jamestalton (approved):
lgtm

PR #280, opened 2023-03-09
Empty state component in PageTable.tsx

ShaiahWren:
Hello Zita,

Thank you for your very helpful input today. I will continue this ticket by adding the EmptyStateError component. 

Changes to the code should show no UI changes - styling remains the same and only improves code quality.

All best,
Shaiah


vidyanambiar (commented):
@ShaiahWren I see some styling changes in the Empty state introduced by this PR. The content width seems smaller and introduces more wrapped text. Also, the icon is not shown. Here's a comparison of the empty state on the Teams page with and without these changes. Is this an intended change in styling? cc: @jamestalton 

Current implementation on `main`
Pasted Graphic 1

PR:
image

PR #281, opened 2023-03-10
Adds baseline e2e tests for organizations

mabashian:
I also removed a duplicate test in the teams e2e test suite.  `edit team` and `team details edit team` were effectively testing the same thing.  cc @vidyanambiar

### DIFF
--- a/cypress/e2e/awx/access/organizations.cy.ts
+++ b/cypress/e2e/awx/access/organizations.cy.ts
@@ -40,4 +40,74 @@ describe('organizations', () => {
     cy.navigateTo(/^Organizations$/, false);
     cy.hasTitle(/^Organizations$/);
   });
+
+  it('create organization', () => {
+    const organizationName = 'E2E Organization ' + randomString(4);
+    cy.navigateTo(/^Organizations$/, false);
+    cy.clickLink(/^Create organization$/);
+    cy.typeByLabel(/^Name$/, organizationName);
+    cy.clickButton(/^Create organization$/);
+    cy.hasTitle(organizationName);
+    // Clean up this organization
+    cy.clickPageAction(/^Delete organization/);
+    cy.get('#confirm').click();
+    cy.clickButton(/^Delete organization/);
+    cy.hasTitle(/^Organizations$/);
+  });
+
+  it('organization details', () => {
+    cy.navigateTo(/^Organizations$/, false);
+    cy.clickRow(organization.name);
+    cy.hasTitle(organization.name);
+    cy.clickButton(/^Details$/);
+    cy.contains('#name', organization.name);
+  });
+
+  it('organization details edit organization', () => {
+    cy.navigateTo(/^Organizations$/, false);
+    cy.clickRow(organization.name);
+    cy.hasTitle(organization.name);
+    cy.clickButton(/^Edit organization$/);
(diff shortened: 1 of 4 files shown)


vidyanambiar (commented):
Not directly related to this PR but it looks like the created organizations are not getting cleaned up in the [after hook](https://github.com/ansible/ansible-ui/pull/281/files#diff-1da165254959ba193f4cd2125f95f3d83cb0d807c47fb3e45a355d355ab78705R18). With each run of the suite a new E2E organization remains behind on the live server. I think the issue is on [this line](https://github.com/ansible/ansible-ui/pull/281/files#diff-1da165254959ba193f4cd2125f95f3d83cb0d807c47fb3e45a355d355ab78705R19). We need to use the `created__gt` query param instead of `created__lt` to delete organizations created in the past 2 hours.

frontend/awx/access/organizations/Organizations.tsx
```diff
@@ -64,8 +65,8 @@ export function Organizations() {
         variant: ButtonVariant.primary,
         icon: PlusIcon,
         label: t('Create organization'),
-        href: RouteObj.CreateOrganization,
-      },
+        onClick: () => navigate(RouteObj.CreateOrganization),
+      } as IPageActionButton,
```
jamestalton, line 69: The href is preferrable as a user can right click and open in new window.  Keith put that in.

cypress/e2e/awx/access/organizations.cy.ts
```diff
@@ -16,7 +16,7 @@ describe('organizations', () => {
     // Sometimes if tests are stopped in the middle, we get left over organizations
     // Cleanup E2E organizations older than 2 hours
     cy.requestGet<ItemsResponse<Organization>>(
-      `/api/v2/organizations/?limit=100&created__lt=${new Date(
+      `/api/v2/organizations/?limit=100&created__gt=${new Date(
```
jamestalton, line 19: wouldn't this get new organizations and not only ones older than 2 hours?
since multiple e2e runs could be going against the same server, thiswould delete the orgs that the other e2e is using

vidyanambiar: Ah I see, my bad! I misinterpreted how that query param was being used because I saw all the new E2E orgs from the `'create organization'` remain in the list after multiple test runs. @mabashian Perhaps we can revert this to `created__lt` but also delete the organization as a final step in the `'create organization'` spec?

cypress/e2e/awx/access/teams.cy.ts
```diff
@@ -87,16 +87,6 @@ describe('teams', () => {
...
-    cy.navigateTo(/^Teams$/, true);
-    cy.clickRow(team.name);
-    cy.clickButton(/^Edit team$/);
-    cy.hasTitle(/^Edit team$/);
-    cy.typeByLabel(/^Name$/, 'a');
-    cy.clickButton(/^Save team$/);
-    cy.hasTitle(`${team.name}a`);
-  });
-
   it('team details', () => {
```
jamestalton, line 90: while this is much like the edit from details page, deleting this gets rid of the coverage on clicking the row edit going to the edit page. in addition, edit from details returns to the team details. edit from row, returns to the teams page.

vidyanambiar (approved):
lgtm!

PR #283, opened 2023-03-10
UI updates to teams list edit action button + E2E tests

vidyanambiar:
1. The "edit" row action is part of a kebab menu in all the lists except "Teams" where it shows up as an icon outside the kebab menu. This PR moves the row action for editing teams into the kebab menu to be consistent with other lists.
2. The dropdown actions that fall under a kebab menu have a static `toggle-kebab` id that is repeated on the page when there are multiple kebab menus (for instance on each row of a list), making the id non-unique. The id was being set in the framework component for the dropdown action. Since the framework component could apply to any type of resource/object, I've replaced the `id` attribute with `className` and updated the tests accordingly.

@mabashian Thanks for catching these issues!

### DIFF
--- a/cypress/e2e/awx/access/teams.cy.ts
+++ b/cypress/e2e/awx/access/teams.cy.ts
@@ -231,7 +231,7 @@ describe('teams', () => {
 
   it('teams table row edit team', () => {
     cy.navigateTo(/^Teams$/, true);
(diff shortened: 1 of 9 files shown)


jamestalton (approved):
lgtm

PR #293, opened 2023-03-14
Brand rework

keithjgrant:
* Replaces logos and icons with the angry potato
* Fixes header logo to use img tag rather than inlined svg


jamestalton (approved):
lgtm

PR #294, opened 2023-03-15
Support form layout control using page form section

jamestalton:
In the case where you want a form input to take up the full space horizontally you can now wrap the component in a PageFormSection with a singleColumn prop.


lgalis (approved):
LGTM

PR #300, opened 2023-03-16
Fix various user form bugs and add basic user e2e test coverage

jamestalton:
- Fixes bug where typing organization into the input would result in a form validation error when submitting the user add form
- Fixes bug where users were unable to specify a first or last name in user add/edit forms
- Fixes bug where password fields were required when submitting the user edit form
- Added e2e test coverage for the following cases:
  - Adding a user
  - Rendering user details
  - Editing user from details
  - Deleting user from details
  - Editing user from list
  - Deleting user from list row action
  - Deleting user from list bulk action

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Ansible UI
 
-The UI for [Ansible](https://www.ansible.com).
+The UI for [Ansible](https://www.ansible.com)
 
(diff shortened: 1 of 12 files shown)


jamestalton (commented):
lgtm

PR #307, opened 2023-03-20
Updates some e2e test descriptions to be more verbose

mabashian:
Per conversations with @akus062381 we want the descriptions of the e2e tests to read more like sentences.  I tried to add a little more context to each so that I was clear exactly what was being tested.


akus062381 (approved):
This is exactly what I was hoping for, awesome!

PR #311, opened 2023-03-20
add default selection support

jamestalton:
Support for a view to have default selection

This is not hoked up to any dialogs yet, this is in preparation for future work.

In the case where you are opening a dialog from say the Teams view,
you create an AWXView for the dialogs table with the currently selected items using useAwxView({ defaultSelections: view.selected })

Or from a PageFormSelect where the current value in the select is passed to a SelectDialog by creating a new AwxView with the current item as the defaultSelections

### DIFF
--- a/framework/PageTable/useTableItems.tsx
+++ b/framework/PageTable/useTableItems.tsx
@@ -90,9 +90,15 @@ export interface ISelected<T extends object> {
   keyFn: (item: T) => string | number;
 }
 
+/** Hook to track selection from a list of items. */
 export function useSelected<T extends object>(
+  /** The items in which selections are being tracked. Used to update the selected items when an item changes.  */
   items: T[],
+
+  /** A function that returns a unique key for each item, used to track selection. */
   keyFn: (item: T) => string | number,
+
+  /** The default items that should be initially selected. */
   defaultSelection?: T[]
 ): ISelected<T> {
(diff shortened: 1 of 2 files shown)


framework/PageTable/useTableItems.tsx
```diff
@@ -92,9 +92,19 @@ export interface ISelected<T extends object> {
 
 export function useSelected<T extends object>(
   items: T[],
-  keyFn: (item: T) => string | number
+  keyFn: (item: T) => string | number,
+  defaultSelection?: T[]
```
AlexSCorey, line 96: Will this function every time the modal is opened?  If so, I  think  `defaultSelection` is a misnomer and should be changed to something like `selectedItems`, or `value`.

AlexSCorey (changes requested):
@jamestalton I checked out this branch locally, and it there is still a disconnect between the form field value, and the selected items in the modal.  See gif.
selectedValues

framework/PageTable/useTableItems.tsx
jamestalton: This PR does not hook up the default selection for the view into any dialogs.
This just adds the support for the view and future PRs can take advantage of that.

vidyanambiar (commented):
Looks good 👍  

It would help to have some documentation added around this and perhaps an example scenario.

vidyanambiar (approved):

AlexSCorey (approved):

PR #316, opened 2023-03-21
Cleanup crud hooks

jamestalton:
| Hook              | Description                                   |
| ----------------- | --------------------------------------------- |
| useGet            | returns data from a url                       |
| useOptions        | returns options from a url                    |
| useGetRequest     | returns a function to get data from a url     |
| useOptionsRequest | returns a function to options data from a url |
| usePostRequest    | returns a function to post data to a url      |
| usePatchRequest   | returns a function to patch data to a url     |
| useDeleteRequest  | returns a function to delete data to a url    |


frontend/common/crud/useGet.tsx
```diff
@@ -28,7 +28,9 @@ export function useGet<T>(url: string, query?: Record<string, string | number |
     url += '?' + new URLSearchParams(normalizedQuery).toString();
   }
 
-  const response = useSWR<T>(url, getRequest);
+  const response = useSWR<T>(url, getRequest, {
+    dedupingInterval: 0,
```
AlexSCorey, line 32: Based on SWR docs it looks like this value is used to prevent duplicate requests within a certain time interval with a default interval being 2000 ms. It seems like this code will allow an immediate duplicate request to occur.  Can you expand on this line, and what we are trying to achieve with it?

frontend/common/Debug.tsx
```diff
@@ -5,8 +5,8 @@ import { PageHeader } from '../../framework';
 import { useBulkActionDialog } from '../../framework/BulkActionDialog';
```
AlexSCorey, line 5: What does this file do?  The name of it 'Debug' is not very descriptive.  Do we need it?

jamestalton: It is the Debug tab in the UI which only shows up in a development build

frontend/common/crud/useGet.tsx
jamestalton: Our code is clean enough that we do not run into an issue when we need dedep of network calls.
In the case of tests, where a test gets an item, edits an item, goes to details, which get the item,
All in under 2000ms, the edited item is not gotten by SWR because of the dedup, so we turn it off.

PR #320, opened 2023-03-21
skip randomly failing e2e test

jamestalton:
A test was added that broke tests. API is returning a 403, when it looks like it should not be.
Disabling test as this blocks other PRs.
Test will be reenabled once it works consistently.

### DIFF
--- a/cypress/e2e/awx/views/jobs.cy.ts
+++ b/cypress/e2e/awx/views/jobs.cy.ts
@@ -98,7 +98,8 @@ describe('jobs', () => {
     });
   });
 
-  it('deletes a job from the jobs list toolbar', () => {
+  // this test is failing randomly with a 403. need to fix before reenabling
+  it.skip('deletes a job from the jobs list toolbar', () => {
     cy.get('.pf-c-select__toggle').click();
     cy.clickButton('ID');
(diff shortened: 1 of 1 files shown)


vidyanambiar (approved):

cypress/e2e/awx/views/jobs.cy.ts
```diff
@@ -98,7 +98,8 @@ describe('jobs', () => {
     });
   });
 
-  it('deletes a job from the jobs list toolbar', () => {
+  // this test is failing randomly with a 403. need to fix before reenabling
```
AlexSCorey, line 101: I wonder if it is trying to delete a job that has already been deleted, or one that hasn't started?

AlexSCorey (changes requested):
Also, please add a bit of context in the description above. Some questions that might be helpful:
1. What have we done to track down the issue?
2. What is our plan going forward to fix it?  I don't think leaving a comment in the code will suffice.  We should document this in official channels so that those that need to know our workload can see what needs to be done, what's in progress, and what has been done.

AlexSCorey (approved):
Looks good.

vidyanambiar:
@AlexSCorey I am working on this. The issue appears to be that the job is somehow in running state when the delete is triggered (even though I was checking for the job status to change to "successful". I'm investigating a fix

PR #321, opened 2023-03-22
Adds last login and authentication type to user details

mabashian:
These two fields were missing from the user details.  Last login displays the time since the user last logged in.  Authentication type should be displayed when the user is either an LDAP user or a social auth user.  Here's what it looks like in action:

Screenshot 2023-03-21 at 8 59 29 PM

I coded the behavior of the Authentication type field to match the existing AWX behavior.  We have outstanding enhancement requests to show something more specific than `SOCIAL` since we have the auth provider name in the user.auth array but we should talk through the UX there before implementing.


jamestalton (approved):
lgtm

mabashian:
Note that there is one small difference between the old and new UI here.  In the old UI we didn't show the auth type for local users.  For these local users I decided to show a label that says `Local`.  This is shown only when `user.ldap_dn` is empty and `user.auth` is an empty array.

Screenshot 2023-03-22 at 9 25 46 AM

PR #329, opened 2023-03-22
Adds rbac parameters to user add, edit and delete buttons in user list

mabashian:
Screenshot 2023-03-22 at 1 59 42 PM
Screenshot 2023-03-22 at 1 59 30 PM
Screenshot 2023-03-22 at 1 59 14 PM

The actions around adding/remove from teams and orgs is _much_ more complicated from an RBAC perspective so I left that alone for now.

### DIFF
--- a/frontend/awx/access/users/UserPage/UserTeams.tsx
+++ b/frontend/awx/access/users/UserPage/UserTeams.tsx
@@ -33,7 +33,7 @@ export function UserTeams(props: { user: User }) {
   const selectTeamsAddUsers = useSelectTeamsAddUsers(view.selectItemsAndRefresh);
   const removeTeamsFromUsers = useRemoveTeamsFromUsers(view.unselectItemsAndRefresh);
   const { data } = useOptions<OptionsResponse<ActionsResponse>>('/api/v2/users/');
-  const canCreateUser = Boolean(data && data.actions && data.actions['POST']);
+  const canAddUserToTeam = Boolean(data && data.actions && data.actions['POST']);
 
   const toolbarActions = useMemo<IPageAction<Team>[]>(
     () => [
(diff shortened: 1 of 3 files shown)


frontend/awx/access/users/UserPage/UserTeams.tsx
```diff
@@ -33,7 +33,7 @@ export function UserTeams(props: { user: User }) {
   const selectTeamsAddUsers = useSelectTeamsAddUsers(view.selectItemsAndRefresh);
   const removeTeamsFromUsers = useRemoveTeamsFromUsers(view.unselectItemsAndRefresh);
   const { data } = useOptions<OptionsResponse<ActionsResponse>>('/api/v2/users/');
-  const canAddTeam = Boolean(data && data.actions && data.actions['POST']);
+  const canCreateUser = Boolean(data && data.actions && data.actions['POST']);
```
jamestalton, line 36: should this `canAddUserToTeam`

frontend/awx/access/users/hooks/useDeleteUsers.tsx
```diff
@@ -18,15 +18,34 @@ export function useDeleteUsers(onComplete: (users: User[]) => void) {
...
+  };
   const bulkAction = useBulkConfirmation<User>();
   const deleteUsers = (users: User[]) => {
+    const undeletableUsers = users.filter(cannotDeleteUser);
+
     bulkAction({
       title: t('Permanently delete users', { count: users.length }),
       confirmText: t('Yes, I confirm that I want to delete these {{count}} users.', {
-        count: users.length,
+        count: undeletableUsers.length,
```
jamestalton, line 33: would this be deletableUsers.count?

frontend/awx/access/users/UserPage/UserTeams.tsx
mabashian: Sure - we use the presence of the POST attribute on an OPTIONS response to `/api/v2/users` to determine whether or not a user can add a team but also whether or not a user can add other users to a team.

PR #330, opened 2023-03-22
Cancel job from icon-only row action 

vidyanambiar:
Includes framework updates:
This PR includes changes to the DropdownAction in the framework to align actions to the right of the row. 

Screenshot 2023-03-22 at 1 06 00 PM


jamestalton (approved):
lgtm

PR #337, opened 2023-03-23
Update projects list columns and the create/edit project pages

lgalis:
- Updated Status column to the import_state. 
- Make type read only as only Git is supported as of now
- Make URL field mandatory for project Create/update.
- Status updated in the details page

Screenshot from 2023-03-23 15-57-46

Screenshot from 2023-03-23 15-48-49

Screenshot from 2023-03-23 15-57-11


jamestalton (approved):
lgtm

PR #339, opened 2023-03-27
error state component

ShaiahWren:
Hi, 

This pr takes the place of https://github.com/ansible/ansible-ui/pull/280, which is now closed. The UI should show no changes.  

Let me know if you have any questions or concerns. 

Thank you!

### DIFF
--- a/framework/PageTable/PageTable.tsx
+++ b/framework/PageTable/PageTable.tsx
@@ -14,7 +14,7 @@ import {
   Stack,
   Title,
 } from '@patternfly/react-core';
-import { ExclamationCircleIcon, PlusCircleIcon, SearchIcon } from '@patternfly/react-icons';
+import { SearchIcon } from '@patternfly/react-icons';
 import {
   CollapseColumn,
   SortByDirection,
@@ -40,8 +40,20 @@ import {
   useRef,
   useState,
 } from 'react';
+<<<<<<< HEAD
 import styled from 'styled-components';
+=======
+<<<<<<< HEAD
+>>>>>>> ea9ec1f3 (error state component)
 import { IPageAction, PageActionSelection } from '../PageActions/PageAction';
+=======
+import { EmptyStateNoData } from '../components/EmptyStateNoData';
+import { EmptyStateError } from '../components/EmptyStateError';
+import { Scrollable } from '../components/Scrollable';
+import { useBreakpoint } from '../components/useBreakPoint';
+import { IPageAction } from '../PageActions/PageAction';
+import { PageActionType } from '../PageActions/PageActionType';
+>>>>>>> 787e35d0 (error state component)
 import { PageActions } from '../PageActions/PageActions';
 import { PageBody } from '../PageBody';
 import { useColumnModal } from '../PageColumnModal';
@@ -281,6 +293,7 @@ export function PageTable<T extends object>(props: PageTableProps<T>) {
 
   if (error) {
     return (
+<<<<<<< HEAD
       <EmptyStateDiv>
         <EmptyState variant={EmptyStateVariant.small} style={{ paddingTop: 48 }}>
           <EmptyStateIcon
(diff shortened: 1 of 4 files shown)


framework/components/EmptyStateError.tsx
```diff
@@ -0,0 +1,28 @@
...
+import {
+  Title,
+  EmptyStateBody,
+  EmptyState,
+  EmptyStateIcon,
+  EmptyStateVariant,
+  Button,
+} from '@patternfly/react-core';
+
+export function EmptyStateError(props: { message?: string }) {
```
ZitaNemeckova, line 11: The default messages needs to be translated.

```
import { useTranslation } from 'react-i18next';
...
const title = t('Something went wrong');
...
```

ZitaNemeckova: Yep :)

framework/PageTable/PageTable.tsx
```diff
@@ -164,43 +165,29 @@ export function PageTable<T extends object>(props: PageTableProps<T>) {
...
-            color="var(--pf-global--danger-color--100)"
-          />
-          <Title headingLevel="h2" size="lg">
-            {/* Unable to connect */}
-            {props.errorStateTitle}
-          </Title>
-          {/* <EmptyStateBody>There was an error retrieving data. Check your connection and reload the page.</EmptyStateBody> */}
-          <EmptyStateBody>{error.message}</EmptyStateBody>
-        </EmptyState>
+        <EmptyStateError message={error.message} />
```
ZitaNemeckova, line 168: @ShaiahWren I think it makes sense to allow `title` to be an optional param as well. WDYT?

ShaiahWren: Sure, I can do that.

ShaiahWren: Ah, so title is already an optional param called 'message' - in `EmptyStateError` the default title is `const title = t('Something went wrong');`.  Perhaps `error.message` is not the correct alternative to the default message. I'm trying to reconcile what is suggested in the mockups with what we already have:
Screenshot 2023-03-29 at 9 12 49 AM

ZitaNemeckova (approved):
LGTM 👍 needs rebase

vidyanambiar (commented):
- I see some styling changes in the Empty state introduced by this PR. The content width is smaller and introduces more wrapped text in the empty state messages. Is this an intended styling change?
- 
image
image

- In the error scenario it might be good to have more vertical spacing between the error message and the prompt "Please try refreshing.." to be consistent with the spacing above the error message:
- 
image

framework/PageTable/PageTable.tsx
```diff
@@ -6,13 +6,12 @@ import {
   EmptyStateBody,
   EmptyStateIcon,
   EmptyStateSecondaryActions,
-  EmptyStateVariant,
```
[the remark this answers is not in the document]
ShaiahWren: HI @vidyanambiar, thank you for your review! My recent changes should address these concerns. Let me know if there are any more questions.

vidyanambiar:
Thanks for addressing the comments, @ShaiahWren. I tested the empty state and it looks good to me. 👍 

I noticed a styling issue in the error state. The "Please refresh.." text shows up in white and is hard to read. 
image

vidyanambiar (approved):
lgtm

framework/PageTable/PageTable.tsx
```diff
@@ -281,50 +283,55 @@ export function PageTable<T extends object>(props: PageTableProps<T>) {
 
   if (error) {
     return (
-      <EmptyStateDiv>
```
ZitaNemeckova, line 284: If `<EmptyStateDiv>` uses are removed everywhere it's definition should be too.

ZitaNemeckova (approved):
LGTM. Great work!

PR #343, opened 2023-03-27
Add basic job output

keithjgrant:
Adds basic job output behavior

* handles expand/collapse for plays/tasks on playbook jobs
* displays error/stack trace for jobs w/ "error" status
* falls back to "flat mode" (no expand/collapse functionality) for job types that don't support it or if children_summary fails to load

Still TODO in future PRs:
* websockets & live events streaming
* page up/down and scroll to top/bottom buttons
* follow mode
* search & filters


vidyanambiar:
Another item for future PRs could be the empty state, similar to "No output found for this job" in the existing awx UI.

PR #344, opened 2023-03-28
Adds AWX inventory details view

mabashian:
Add support for inventory details view in AWX.  Currently missing from the implementation is:

1) The Variables detail which will come in a follow-on PR

Regular inventory:
Screenshot 2023-03-31 at 10 41 48 AM

Smart inventory:
Screenshot 2023-03-31 at 10 42 10 AM

Constructed inventory:
Screenshot 2023-03-31 at 10 42 35 AM

One open question that I had was around handling the replacing the :inventory_type in the route.  I copy/pasted the mapping in several places but wasn't sure if that was a best practice or if there was a way to make that generic and reusable.

### DIFF
--- a/frontend/awx/interfaces/Inventory.ts
+++ b/frontend/awx/interfaces/Inventory.ts
@@ -1,6 +1,78 @@
 import { Inventory as SwaggerInventory } from './generated-from-swagger/api';
+import {
+  SummaryFieldsByUser,
+  SummaryFieldsOrganization,
+  SummeryFieldObjectRole,
+} from './summary-fields/summary-fields';
+import { Label } from './Label';
 
-export interface Inventory extends Omit<SwaggerInventory, 'id' | 'name' | 'summary_fields'> {
+export interface CommonInventory
+  extends Omit<
+    SwaggerInventory,
+    'id' | 'name' | 'type' | 'kind' | 'summary_fields' | 'organization'
+  > {
   id: number;
   name: string;
+  type: 'inventory';
+  summary_fields: {
+    organization: SummaryFieldsOrganization;
+    created_by: SummaryFieldsByUser;
+    modified_by: SummaryFieldsByUser;
+    object_roles: {
+      admin_role: SummeryFieldObjectRole;
+      update_role: SummeryFieldObjectRole;
+      adhoc_role: SummeryFieldObjectRole;
+      use_role: SummeryFieldObjectRole;
+      read_role: SummeryFieldObjectRole;
+    };
+    user_capabilities: {
+      edit: boolean;
+      delete: boolean;
+      copy: boolean;
+      adhoc: boolean;
+    };
+  };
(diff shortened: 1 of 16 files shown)


vidyanambiar (approved):
Looks good 👍  A couple comments below:

- Not seeing Credentials and Labels in this implementation. Could those be added in a future PR as well?
- The awx UI has help texts for certain fields. This may be a framework level discussion since it applies to various details pages and not only templates.

frontend/awx/interfaces/Inventory.ts
```diff
@@ -1,6 +1,72 @@
-import { Inventory as SwaggerInventory } from './generated-from-swagger/api';
+import { SummaryFieldsByUser, SummaryFieldsOrganization } from './summary-fields/summary-fields';
+import { Label } from './Label';
 
-export interface Inventory extends Omit<SwaggerInventory, 'id' | 'name' | 'summary_fields'> {
```
jamestalton, line 3: This removes the usage of the interface that is generated from swagger.
Whats the thoughts behind this?

frontend/awx/resources/inventories/InventoryPage/InventoryDetails.tsx
```diff
@@ -0,0 +1,176 @@
...
+export function InventoryDetails(props: { inventory: Inventory }) {
+  const { t } = useTranslation();
+  const { inventory } = props;
+  const history = useNavigate();
+  const params = useParams<{ id: string }>();
+  const [inputInventories, setInputInventories] = useState<Inventory[]>([]);
+  const [inputInventoriesError, setInputInventoriesError] = useState<string>();
+  const instanceGroups = useInstanceGroups(params.id || '0');
+
+  useEffect(() => {
```
jamestalton, line 37: instead of this effect a call to useGet should work better here

jamestalton (changes requested):

frontend/awx/interfaces/Inventory.ts
```diff
@@ -1,6 +1,78 @@
...
+    labels: {
+      count: number;
+      results: Label[];
+    };
+  };
+}
+
+export interface RegularInventory extends CommonInventory {
+  kind: '';
+  verbosity?: never;
```
mabashian, line 70: @jamestalton I wrote a hook that converts the value that the api gives us for verbosity into the string that we want to show in the UI: frontend/awx/common/useVerbosityString.tsx.  In the case of inventories, verbosity is only available for constructed inventories and I couldn't figure out how to call that hook without the linter getting mad that verbosity may not be defined.  To get around that I went down the `never` route but I'm not sure how accepted this route is.  You can reproduce the scenario I'm talking about by removing this line and then going and looking at my call to useVerbosityString in InventoryDetails.

jamestalton (approved):
lgtm

PR #351, opened 2023-03-29
Fix job launch redirect url

keithjgrant:
* Fixes the job launch to redirect to the correct job output URL
* Adds helper util for generating job urls based on job type
* Adds a banner to the job output page if a job is running, indicating live events are not yet supported

Screenshot 2023-03-29 at 4 03 14 PM


vidyanambiar (changes requested):
Tested with playbook run job, workflow job, management job , inventory sync, project update and source control update. Worked in all cases except for workflow job.
image

keithjgrant:
> Worked in all cases except for workflow job.

Updated to add a placeholder component for the workflow job page.

vidyanambiar (approved):
lgtm!

PR #358, opened 2023-03-31
WIP: Expand job row to reveal additional details

vidyanambiar:


frontend/awx/common/CredentialLabel.tsx
```diff
@@ -19,14 +21,23 @@ function CredentialLabel(props: { credential: Credential }) {
   } else {
     type = toTitleCase(credential.kind || '');
   }
+  const vault_id = useMemo(() => {
```
keithjgrant, line 24: useMemo probably isn't necessary around this

frontend/awx/views/jobs/jobUtils.tsx
```diff
@@ -19,7 +19,19 @@ export function getJobsAPIUrl(type: string) {
...
+  | 'pending'
+  | 'waiting'
+  | 'running'
+  | 'successful'
+  | 'error'
+  | 'canceled'
+  | undefined;
+
+export function isJobRunning(status: JobStatus) {
+  if (!status) return false;
```
keithjgrant, line 34: In the past, if status missing, we've cast it to `'waiting'`

see: https://github.com/ansible/awx/blob/devel/awx/ui/src/screens/Job/JobOutput/JobOutput.js#L117

keithjgrant (approved):
Looks good! Just a couple comments above

vidyanambiar:
@keithjgrant Thanks for the review! I've addressed your comments.

PR #362, opened 2023-03-31
[AAP-9981] Updated decision environment pages

lgalis:
Updated decision environments pages according to the latest mocks
Moved the Decision environment under  the Resources left hand menu section

Note- the endpoints used are not yet implemented. Screenshots here use mocked data ).

Screenshot from 2023-03-31 16-10-22
Screenshot from 2023-03-31 16-09-44
Screenshot from 2023-03-31 16-09-07


jamestalton (approved):
lgtm

PR #364, opened 2023-04-03
Formatting job 'started' and 'finished' times

vidyanambiar:
Following UXD recommendation, the format for the "started" and "finished" times is updated to the MM/DD/YYYY, hh:mm:ss AM/PM format. 

Note: In the future we may provide a setting for users to select the format for dates.

Screenshot 2023-04-03 at 11 46 43 AM

### DIFF
--- a/frontend/awx/views/jobs/hooks/useJobsColumns.tsx
+++ b/frontend/awx/views/jobs/hooks/useJobsColumns.tsx
@@ -1,6 +1,6 @@
 import { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
-import { ITableColumn, SinceCell, TextCell } from '../../../../../framework';
+import { ITableColumn, DateTimeCell, TextCell } from '../../../../../framework';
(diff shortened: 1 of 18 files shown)


frontend/awx/views/jobs/hooks/useJobsColumns.tsx
```diff
@@ -62,14 +63,14 @@ export function useJobsColumns(options?: { disableSort?: boolean; disableLinks?:
       },
       {
         header: t('Started'),
-        cell: (job: UnifiedJob) => job.started && <SinceCell value={job.started} />,
+        cell: (job: UnifiedJob) => job.started && <p>{formatDateString(job.started)}</p>,
```
jamestalton, line 66: Could we enhance the SinceCell?
- Rename to DateTimeCell
- Add a format prop: "since" | "date-time"

That way any other places that need the same formatting can use that component.

jamestalton (approved):
lgtm

PR #366, opened 2023-04-03
Update the source name and timestamp field names for an Audit Event

lgalis:
Update the source name and timestamp field names for an Audit Event: 
name-> source_name, 
timestamp->received_at


ddonahue007 (approved):
lgtm

PR #372, opened 2023-04-04
Fix for broken Instance group and Default execution environment fields in Organization form

vidyanambiar:
Also addresses an issue with the selected Execution Environment not being correctly saved in the Template form.

image

image


marshmalien (approved):
Tested locally. LGTM 👍

PR #376, opened 2023-04-06
Enhanced table columns

jamestalton:
- JSDoc for table column options
- Adds options for displaying columns in the expanded row

### DIFF
--- a/framework/PageDetails/PageDetails.tsx
+++ b/framework/PageDetails/PageDetails.tsx
@@ -2,11 +2,15 @@ import { DescriptionList, PageSection } from '@patternfly/react-core';
 import { ReactNode } from 'react';
 import { useSettings } from '../Settings';
 
-export function PageDetails(props: { children?: ReactNode; disablePadding?: boolean }) {
+export function PageDetails(props: {
+  children?: ReactNode;
+  disablePadding?: boolean;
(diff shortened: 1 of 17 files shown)


vidyanambiar (changes requested):
Expanded row: 
- In some cases the description appears above the Created and Modified fields (Organizations) and other cases it appears below (Projects, Templates..). Could we update it to always show at the top for consistency?
image

- Perhaps a "Description" title will be good to distinguish it from other fields. 
image

jamestalton:
The description, when coming from a column should always show up above the details. This will give it prominence under the item and allow the description to fill the whole horizontal space.

If a developer adds "expandedRow" content specifically to a table (i.e. a chart), it will show under the details that are generated from the columns.

In the case of the project screenshot above, the description is being added by the "expandedRow" on the table and not as a column with description.

We will need a follow up PR to address the use of "expandedRow" to show descriptions.

vidyanambiar:
I think a "Description" title would be helpful in such cases to distinguish this field from the other fields. In the screenshot above the description looks like an extension of the "Created" field which could be confusing to users.

jamestalton:
Updated the rest of the tables such as projects to use the column description.

We can add a description title to it, but since it now shows above the other details, i'd like to try it with out the title first.

vidyanambiar (approved):

PR #377, opened 2023-04-06
Add 202 response handling for usePostRequest hook.

kialam:
Fixes case where the BulkActionDialog shows an error upon successful completion of a POST request that returns a `202`.

Before:
Screenshot 2023-04-06 at 9 03 04 AM

After:
Screenshot 2023-04-06 at 9 02 23 AM


jamestalton (commented):
lgtm

PR #382, opened 2023-04-07
[EDA] Rulebook Activations CRUD Deletion Test

akus062381:
Adds a test to delete a rulebook activation from the details page.
Makes organizational changes to several spec files.

Editing a rulebook activation cannot be added because this functionality does not exist.
Testing a user's ability to enable/disable a rulebook activation will be added next week after that functionality is working.


lgalis (approved):
LGTM

PR #383, opened 2023-04-07
Update the decision environment selection in the Rulebook activation pages

lgalis:
Update the decision environment selection in the Rulebook activation pages

Screenshot from 2023-04-07 17-34-19
Screenshot from 2023-04-07 17-41-10


ddonahue007 (approved):
lgtm

PR #391, opened 2023-04-11
Refactor StatusCell and add useStatusColumn hook

kialam:
Based on suggestions and meetings with UXD and the feedback given, this PR:

- refactors our `StatusCell` component to use the `Textcell` component instead of PF's `Alert` component
- adds a new hook for the status column, `useStatusColumn` that takes props for tooltips and redirect handling
- updates the Projects List status column to use the new hook and redirects the user to the current or last project sync job.
- removes the Github and Unknown icons from the `Scm` component.

Screenshot 2023-04-11 at 12 05 34 PM

### DIFF
--- a/frontend/common/columns.tsx
+++ b/frontend/common/columns.tsx
@@ -189,6 +189,7 @@ export function useOrganizationNameColumn(options?: {
 }
 export function useStatusColumn(options?: {
   tooltip?: string;
+  tooltipAlt?: string;
   disableLinks?: boolean;
   disableSort?: boolean;
 }) {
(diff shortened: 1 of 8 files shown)


frontend/common/columns.tsx
```diff
@@ -185,6 +187,48 @@ export function useOrganizationNameColumn(options?: {
...
+            status={item.status}
+            to={RouteObj.JobOutput.replace(':job_type', item.type ? item.type : '').replace(
+              ':id',
+              (
+                item.summary_fields?.current_job?.id ??
+                item.summary_fields?.last_job?.id ??
+                ''
+              ).toString()
+            )}
+            disableLinks={options?.disableLinks}
```
marshmalien, line 222: This enhancement looks great and resolves the vertical row alignment issues we had with status and project type table cells. 

I have two findings:
1) We should disable the link to the project's sync job when it is missing `summary_fields.current_job` and `summary_fields.last_job`. Currently a project without a job update shows a page not found. 
2) It would be nice if we could make the useStatusColumn hook more generic so that we could reuse it in Inventories. If we include arguments for the link and status, then I think this could also be plugged into the inventories list. I don't think this is necessary for this PR to merge, just a suggestion.

marshmalien (changes requested):

kialam:
@marshmalien thanks for the review. I've gone ahead and disabled linking out if there is no `current_job` or `last_job` fields. I also put the `useProjectStatusColumn` hook into the Projects directory instead of the shared `common` directory. As it stands the code is specific to projects and If we want to revisit the hook at a later date to make it more generic we can definitely do that.

marshmalien (approved):
Very nice! 

Screenshot 2023-04-18 at 1 02 10 PM

PR #395, opened 2023-04-12
Page action switch support

jamestalton:
PageActions now have 5 simpler types:
- Button
- Link
- Switch
- Dropdown
- Seperator
  
PageActions now have a selection prop indicating what type of selections the action works on:
- Multiple
- Single
- None

PageActions are now "pinned" outside of the dropdown using the isPinned prop. This used to be inferred by ButtonVariant.Primary, but now that prop is only for buttons and this makes it clearer in code.
`isPinned: true`

## Enabling and Disabling AWX Instances

https://user-images.githubusercontent.com/6277895/231814974-4733a36e-d384-4041-b654-cc0139e63888.mov

## Collapsed Menu support on small screens

https://user-images.githubusercontent.com/6277895/231815062-9619e040-e3fc-4d62-969c-958b04507576.mov

## Option to show label in row

https://user-images.githubusercontent.com/6277895/231817012-2cd6a4e7-80e1-474c-9cc2-9dcd953e648c.mov


keithjgrant (approved):
Feels a lot cleaner 👍🏻

keithjgrant:
fwiw I filed an issue with PF for that focus ring: https://github.com/patternfly/patternfly/issues/5481

PR #396, opened 2023-04-12
Add/Edit inventory + Framework/E2E updates

vidyanambiar:
- Form to create and edit an inventory
- "Edit Inventory" as a primary button in the inventory details page
- Framework update so that table toolbar buttons of type `PageDropdownAction` can be correctly disabled
- New `PageFormLabelSelect` common component for use in the InventoryForm, TemplateForm and other forms that may require label select (updated `JobTemplateInputs` to use this component)
- Util file `getAddedAndRemoved.ts` moved from controller directory to awx directory

### DIFF
--- a/cypress/e2e/awx/resources/inventories.cy.ts
+++ b/cypress/e2e/awx/resources/inventories.cy.ts
@@ -1,7 +1,9 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 import { randomString } from '../../../../framework/utils/random-string';
 import { Inventory } from '../../../../frontend/awx/interfaces/Inventory';
+import { Label } from '../../../../frontend/awx/interfaces/Label';
 import { Organization } from '../../../../frontend/awx/interfaces/Organization';
+import { InstanceGroup } from '../../../../frontend/awx/interfaces/generated-from-swagger/api';
 
 describe('inventories', () => {
   let organization: Organization;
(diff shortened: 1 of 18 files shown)


keithjgrant (approved):

cypress/e2e/awx/resources/inventories.cy.ts
```diff
@@ -28,12 +31,65 @@ describe('inventories', () => {
     cy.hasTitle(/^Inventories$/);
   });
 
-  it('renders the inventory details page', () => {
+  it('can render the inventories list page', () => {
```
marshmalien, line 34: This looks like a duplicate the test above.

marshmalien (approved):
Pulled down the code and tested the following:
* Ability to create and edit a regular inventory through the form 
* Apply and remove values from each inventory form input
* Form displays warning when user doesn't fill out required fields
* Form displays warning when user attempts to use a resource they don't have permission to "use" (ex. instance groups) 
* Successful save sends user to inventory details
* Create button sends correct payload to API
* Cancel button takes us back to the inventories list

frontend/awx/common/PageFormLabelSelect.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const [labelOptions, setLabels] = useState<Label[]>();
+
+  useEffect(() => {
+    async function fetchLabels() {
+      const allLabels = [];
+      try {
+        const labels = await requestGet<ItemsResponse<Label>>(
+          `/api/v2/labels/?order_by=name&page=1&page_size=200`
+        );
+        allLabels.push(...labels.results);
```
kialam, line 32: Suggestion: add a check here to make sure duplicate label names aren't being added.

cypress/support/commands.ts
```diff
@@ -9,6 +9,15 @@ import { Organization } from '../../frontend/awx/interfaces/Organization';
 import { Project } from '../../frontend/awx/interfaces/Project';
 import { Team } from '../../frontend/awx/interfaces/Team';
 import { User } from '../../frontend/awx/interfaces/User';
+<<<<<<< HEAD
```
kialam, line 12: Looks like some random merge editor code snuck in here.

frontend/awx/common/PageFormLabelSelect.tsx
vidyanambiar: @kialam Good catch! This requires enhancements to the underlying framework component (PageFormCreatableSelect). I also noticed that the existing implementation (also used in the Template form) does not include the option to create a new label. I've created a new Jira issue for both enhancements and would prefer to address them separately.

cypress/support/commands.ts
vidyanambiar: Thanks! Addressed 👍

kialam (approved):

PR #397, opened 2023-04-12
Routed tabs (with URLs)

keithjgrant:
Adds a RoutedTabs component. The API is very similar to PageTabs, but it uses a sub-router for URL-based routing between tabs.

AWX User page has been updated to use RoutedTabs as a proof-of-concept.


jamestalton (approved):
lgtm

PR #399, opened 2023-04-13
[EDA] Login/Logout tests

akus062381:
Adding tests for login/logout in the EDA UI.


cypress/support/commands.ts
```diff
@@ -26,6 +26,7 @@ declare global {
         password: string,
         serverType: string
       ): Chainable<void>;
+      uiLogout(): Chainable<EdaUser | undefined>;
```
vishsanghishetty, line 29: same rename the command to edaLogout

PR #400, opened 2023-04-13
Create eda user and set permissions to enable deploying to Openshift

rooftopcellist:
### Summary

Currently, the eda-ui image build target results in an image that cannot be used in Openshift deployments because of how Openshift injects a non-root uid.

This PR makes the necessary changes so that the container is run as a non-root use, and that user (`eda`) has access to the files needed to bring up the UI.


jamestalton (approved):
lgtm

PR #420, opened 2023-04-18
Update routing for template, job, credential pages

keithjgrant:
Updates Template page, Job page, and Credential page to use RoutedTabs. Also updates these routes for more consistent urls (e.g. `/credentials/:id/details` rather than `/credentials/details/:id`)


vidyanambiar (approved):
looks great!

vidyanambiar:
There seem to be a couple of test failures due to tabs now being represented as links instead of buttons..

cypress/support/component.tsx
```diff
@@ -58,24 +72,28 @@ Cypress.Commands.add('mount', (component, options, rerenderKey) => {
...
-              <Route path="/:id" element={component} />
+              <Route path={`${route?.path || '/:id'}`} element={component} />
             </Routes>
           </Page>
         </ActiveUserProvider>
       </PageFramework>
     </MemoryRouter>,
-    options,
-    rerenderKey
+    {},
```
kialam, line 86: This is hard to grok/parse. Can you add a comment to clarify why the mount method needs these extra parameters? Somewhat related; can these params be optional?

kialam (changes requested):

cypress/support/component.tsx
```diff
@@ -58,24 +72,28 @@ Cypress.Commands.add('mount', (component, options, rerenderKey) => {
...
+    undefined
```
kialam, line 87: Same concern as above

cypress/support/component.tsx
keithjgrant: good call. These are the default values there, so they should be okay to omit entirely

kialam (approved):
👍

PR #425, opened 2023-04-19
Projects expanded list add missing fields

kialam:
Adds the remaining missing fields for the Projects list:
- Last used
- Default Environment (this detail does not show up if there is no default environment returned by the API)
Screenshot 2023-04-19 at 10 33 29 AM

Adds error messaging next to Project name if project is missing an execution environment

Screenshot 2023-04-19 at 10 18 57 AM


jamestalton (approved):
lgtm

PR #429, opened 2023-04-19
Navigate to correct UI (user details) from username dropdown in the masthead

vidyanambiar:
When the user clicks the "User details" menu item in the username dropdown in the masthead, they should be brought to their user's details page. (Currently it is brings the user to the list of users)

image

Updated for both AWX and EDA. cc: @lgalis


lgalis (approved):
LGTM

jamestalton (approved):
lgtm

PR #430, opened 2023-04-19
Refactor custom commands

vishsanghishetty:
This PR is for [ticket](https://issues.redhat.com/browse/AAP-11189)

## Acceptance Criteria
- [x]  `commands` file has all the commands declarations in the global namespace and common commands used by both EDA and AWX
- [x] `auth` file holds all the authentication-related command definitions of both AWX and EDA
- [x] `eda-commands` file holds all the EDA-related command definitions
- [x] `awx-commands` file holds all the AWX-related command definitions
- [x]  `rest-commands` file holds all the REST API-related command definitions

some of the commands which are common need to be adjusted and moved to the `commands` file.


jamestalton (approved):
lgtm

PR #438, opened 2023-04-20
Conditionally render banner.

kialam:
Conditionally render the banner on the AWX dashboard page based on system setting.


jamestalton (approved):
lgtm

PR #442, opened 2023-04-21
[AAP-9966] Updated extra var id paramter. Add Variable code block to the Activation Details page

lgalis:
Updated extra var id paramter. Add Variable code block to the Activation Details page

Screenshot from 2023-04-21 15-06-02


ddonahue007 (approved):
lgtm

PR #448, opened 2023-04-24
Fix the credential id field in the decision environment pages

lgalis:
Fix the credential id field in the decision environment pages

Screenshot from 2023-04-24 10-35-36
Screenshot from 2023-04-24 10-35-14
Screenshot from 2023-04-24 10-35-00


jamestalton (approved):
lgtm

PR #457, opened 2023-04-25
Add paginated output to the activation instance details page

lgalis:
Alternative to #460

Screenshot from 2023-04-25 14-20-00

PR #460, opened 2023-04-25
Get page size for the activation instance output and read all records

lgalis:
**Alternative for #457** 
Get page size for the activation instance output and read all records

Screenshot from 2023-04-25 15-36-56


jamestalton (approved):
lgtm

PR #462, opened 2023-04-25
[AAP-11555] Remove status filtering from the activation history page

lgalis:
Remove status filtering from the activation history page


jamestalton (approved):
lgtm

PR #468, opened 2023-04-26
[AAP-11556] Update activation id in activation instance records. 

lgalis:
Update activation id in activation instance records. ( The field in the activation-instance endpoint changed to activation_id)
Updated Instance names to <instance> - Activation name
Update page size for the rulebooks query on the Create activation form

Screenshot from 2023-04-26 13-56-49

Screenshot from 2023-04-26 13-56-58


jamestalton (approved):
lgtm

PR #470, opened 2023-04-26
Move page table toolbar filter logic into its own file

jamestalton:
This is in preparation for support more filter types in the toolbar.


vishsanghishetty (approved):
LGTM 🚀

PR #472, opened 2023-04-27
Fix masthead link to the user page for EDA

lgalis:
Fixed link to the user details page from the masthead dropdown for EDA
Screenshot from 2023-04-26 21-20-10


jamestalton (approved):
lgtm

PR #474, opened 2023-04-27
Page Table Toolbar Sorting Support

jamestalton:
Screenshot 2023-04-27 at 10 13 45 AM

Adds a sort component to the table toolbar.

This is needed
- for users to be able to sort on columns that are in the expandable row such as `created`.
- for the list and card views as there is no other way to sort on those views.

Changing the column header sort, updates the toolbar sort. Updating the toolbar sort, updates the column header sort.

In addition, there are enhancements to make the table toolbar properly wrap items when there are too many items on the toolbar. This is going to be important as we add support for more filters which are needed by analytics and others.

### DIFF
--- a/framework/PageTable/PageToolbarSort.tsx
+++ b/framework/PageTable/PageToolbarSort.tsx
@@ -107,6 +107,7 @@ function ToolbarSortSelect(props: {
   const onSelect = useCallback(
     (e: unknown, value: string | SelectOptionObject) => {
       setSort?.(value.toString());
+      setOpen(false);
     },
     [setSort]
   );
--- a/framework/PageTable/PageToolbarSort.tsx
+++ b/framework/PageTable/PageToolbarSort.tsx
@@ -10,7 +10,7 @@ import {
   ToolbarToggleGroup,
 } from '@patternfly/react-core';
 import {
-  SortAlphaDownIcon,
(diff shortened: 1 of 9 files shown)


framework/PageTable/PageToolbarSort.tsx
```diff
@@ -0,0 +1,165 @@
...
+      if (column.sort) {
+        if (column.defaultSort) {
+          // Assumes the defauilt sort is a text column
+          sortOptions.push({ label: column.header, value: column.sort, type: 'text' });
+          continue;
+        }
+
+        switch (column.type) {
+          case 'datetime':
+            sortOptions.push({ label: column.header, value: column.sort });
```
ZitaNemeckova, line 148: Why not let this options go to `default`?

vidyanambiar (commented):
Thanks for adding this capability @jamestalton 🎉 I pulled the changes and tested them locally and they work great!

Noticed issues in some cases:
In the Users list, toolbar sort by "Created" causes an error (400). Looks like it's not supported by the users API:
image

jamestalton:
Yeah. The user table columns need to have sorting disabled. I had them disabled at one point for users.

vidyanambiar (approved):

PR #475, opened 2023-04-27
Make user roles tab match existing UI

vidyanambiar:
#### UXD recommendation:
In the Users section, Admin users do not require the banner `"System administrators have unrestricted access to all resources"` that shows up in the User details tab for a system admin. The UX should mirror the flow in the existing UI where it is part of the empty state message for the admin and only in the Roles tab. 

image


jamestalton (approved):
lgtm

PR #476, opened 2023-04-27
Update app sidebar open state based on window size.

jamestalton:
This makes the experience much nicer when changing the size of the window. 

If the window gets to the size where the sidebar is a floating drawer, it auto closes.
If the window goes to a size where the side bar would anchor to the left, it opens.


vidyanambiar (approved):
👍

PR #477, opened 2023-04-27
AWX Dashboard Jobs Chart support showing canceled and error jobs.

jamestalton:
Currently the job chart does not show canceled and error jobs because the AWX server does not return that data.
The recent jobs card shows these and it feels to the user like there is a disconnect.

I have the changes needed for AWX and will see about getting them in. If that goes in, we can merge this and have the right support.


vishsanghishetty (approved):
LGTM 🚀

PR #479, opened 2023-04-28
Delete/Remove Option should always be last and in red text

vidyanambiar:
#### UXD recommendation:

Delete/Remove should always be the last option (if applicable) in a kebab menu and should be in red text (pf-color-red-100). The text should read "Delete [item]" / "Remove [item]"

The delete option should also be separated by a divider line.

image


jamestalton (approved):
lgtm

PR #480, opened 2023-04-28
Page toolbar sort default direction support

jamestalton:
Example sort on created should default to most recent first


vidyanambiar (approved):
lgtm!

PR #483, opened 2023-04-28
Aap 7561 variables form field

AlexSCorey:
This adds language toggling to the code editor, and adds the editor to the template field.  It also adds syntax highlighting for json and yaml


vidyanambiar (commented):
Awesome work 🎉  I tested by creating a job template with some extra variables and it works great. 

Some minor comments:
1. Unlike the existing awx UI, the code editor appears collapsed by default. Is that an intended change from the existing experience?
2. It's nice to see syntax errors being highlighted. Is there a setting that would allow us to view individual error messages on separate lines? They are currently concatenated on the same line. 
image

keithjgrant:
Nice work on this Alex.

I'm still looking through the code, but several things (all pretty minor) came to mind while testing... I'm happy to leave these and iterate in future PRs if we want.

Screenshot 2023-05-01 at 10 35 16 AM

1. I don't like the validation throwing errors at me while I'm typing. I think it needs to wait until blur to validate
2. The error icon overlaps the scrollbar (and the highlighted row outline, if an adjacent row is focused) and feels a bit messy
3. It's making me scroll even when I've only got a few lines of content. Ideally it would buffer with enough space for couple extra blank lines at the bottom
4. It feels crowded in the multi-column layout. I think it should be moved to the end of the form and made full-width to match the mockups
5. The left corners of the YAML button should have a border radius to match the right corners of the JSON button

jamestalton:
`It feels crowded in the multi-column layout. I think it should be moved to the end of the form and made full-width to match the mockups`
You can put a `<PageFormSection>` around the code editor and it will make it left aligned and full with with the right flags.

keithjgrant (approved):

jamestalton (changes requested):

jamestalton (approved):
lgtm

PR #489, opened 2023-05-02
Adds copyright to the about modal

mabashian:
Replaces placeholder text with dynamic copyright.  See https://github.com/ansible/awx/blob/devel/awx/ui/src/components/About/About.js#L35-L43 for what we do in AWX


vidyanambiar (approved):
lgtm

PR #494, opened 2023-05-02
"Back to" list button

keithjgrant:
Adds "Back to Templates", etc buttons to these pages:

* Credential
* User
* Template
* Job

This uses a `<PageBackTab>` component that works as a child of `<RoutedTabs` to read any persisted filters from `sessionStorage` — this is set by adding `usePersistentFilters` on the list page. The list page and the individual item page need to share a common persistentFilterKey (a string) in order make them sync together.

Note that this routing requires tapping into react-router 6, so I couldn't get it working with PageTabs, only the RoutedTabs component. We can continue to add the back button to more pages as we convert them from PageTabs to RoutedTabs.


jamestalton (approved):
lgtm

vidyanambiar (approved):
I don't see the Back button for workflow templates. Is that expected?

Other than that this looks great 👍

keithjgrant:
> I don't see the Back button for workflow templates. Is that expected?

Yeah, that page hasn't been converted to RoutedTabs yet. I'll be sure to get it in the next round

PR #509, opened 2023-05-05
Update edit forms to correctly display errors from the API

vidyanambiar:
Updates "edit resource" forms (rules, groups, projects, credentials, roles & decision environments) to use the `usePatchRequest` hook so that error messages received from the API call can be correctly displayed on the form.

image


vishsanghishetty (approved):
🚀

PR #510, opened 2023-05-05
[EDA] [QE] Add User Tabs Test to Delete Token; Add Custom Commands for Tokens

akus062381:
This PR:

Relates to https://issues.redhat.com/browse/AAP-11021

-Adds a custom command to create a user token in EDA
-Adds a custom command to delete a user token in EDA
-Adds a test to verify a user's ability to delete a token in EDA

### DIFF
--- a/cypress/support/auth.ts
+++ b/cypress/support/auth.ts
@@ -89,7 +89,7 @@ Cypress.Commands.add('edaLogin', () => {
         Cypress.env('EDA_PASSWORD') as string,
         'EDA server'
       );
-      cy.hasTitle(/^Welcome to EDA Server$/);
+      cy.hasTitle('Welcome to');
     },
     {
       cacheAcrossSpecs: true,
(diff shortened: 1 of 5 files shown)


cypress/support/auth.ts
```diff
@@ -89,7 +89,7 @@ Cypress.Commands.add('edaLogin', () => {
         Cypress.env('EDA_PASSWORD') as string,
         'EDA server'
       );
-      cy.hasTitle(/^Welcome to EDA Server$/);
```
vishsanghishetty, line 92: curious as to why `EDA Server` has been removed here and `AWX` in the `awxLogin` command

akus062381: Because in the standalone version of EDA, it says "Welcome To Ansible Automation Platform", not "Welcome to EDA Server"

vishsanghishetty (approved):
🚀

vishsanghishetty:
@akus062381  looks like lint check tests are failing. must be done before pushing always.

- run `npm run prettier` and then `npm run prettier: fix`

- for TS errors `npm run tsc`

- for altogether `npm run checks` (covers all 3)

PR #511, opened 2023-05-07
[AAP-11949] Add permissions to the Role details page

lgalis:
Add permissions to the Role details page

Screenshot from 2023-05-08 14-50-17


vidyanambiar (commented):
Could we map the resource types to a more descriptive/readable format for eg. `Activation instance` instead of `activation_instance`? That way we can also make sure they get translated.

lgalis:
@vidyanambiar - please review

PR #514, opened 2023-05-08
[AAP-11559] [AAP-11921] Separate create and edit functions

lgalis:
Separate create and edit functions
Token field mandatory in the credential create/edit pages.


jamestalton (approved):
lgtm

PR #518, opened 2023-05-08
Adds top level schedule list

AlexSCorey:
---
name: Pull Request
description: This adds the top level schedules list in the side bar for awx

It also adds the routes for Schedules in Routes.tsx, and it adds logic for computing the route for edit and details view of the schedule.

Things to note:

1. Top level schedules list does not have add functionality.  Schedules must be attached to a resource, thus the add functionality should be inside the schedules tab of the resource.  
2. #1 required that we create logic to determine the routes for edit and details view for this list.


frontend/awx/views/schedules/hooks/getScheduleResourceUrl.tsx
```diff
@@ -0,0 +1,39 @@
+import { RouteObj } from '../../../../Routes';
+import { Schedule } from '../../../interfaces/Schedule';
+
+export function getScheduleResourceUrl(schedule: Schedule): string {
+  const getResourceTypeSubString: (type: string) => string = (type) => {
+    let urlSubString = '';
+    switch (type) {
+      case 'job':
```
keithjgrant, line 8: you can just return the string inside each of these `case` statements since you don't do anything else with the string below

frontend/awx/AwxRouter.tsx
```diff
@@ -59,8 +61,9 @@ export function AwxRouter() {
 
         <Route path={RouteObjWithoutPrefix.Jobs} element={<Jobs />} />
         <Route path={RouteObjWithoutPrefix.JobPage} element={<JobPage />} />
+        <Route path={RouteObjWithoutPrefix.CreateSchedule} element={<CreateSchedule />} />
```
vidyanambiar, line 64: Instead of a `"coming soon"` UI could you have this route to the `PageNotImplemented` component. Similar to https://github.com/ansible/ansible-ui/blob/main/frontend/awx/AwxRouter.tsx#L74

vidyanambiar: It would be good to add an `Edit Schedule` route as well (also pointing to `PageNotImplemented`) for the list row action button to route to.

frontend/awx/views/schedules/hooks/useSchedulesColumns.tsx
```diff
@@ -0,0 +1,38 @@
...
+        project_update: t('Source Control Update'),
+        system_job: t('Management Job'),
+        workflow_job: t('Workflow Job'),
+      };
+      return jobTypeLabels[schedule.summary_fields.unified_job_template.unified_job_type];
+    },
+    [t]
+  );
+
+  const typeColumn = useTypeColumn<Schedule>({ ...options, makeReadable });
```
vidyanambiar, line 30: `Resource type` is a sortable field in the awx implementation.

frontend/awx/views/schedules/hooks/useSchedulesColumns.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const nameColumn = useNameColumn({ ...options, onClick: nameClick });
+  const descriptionColumn = useDescriptionColumn();
+  const tableColumns = useMemo<ITableColumn<Schedule>[]>(
+    () => [nameColumn, descriptionColumn, typeColumn],
+    [typeColumn, descriptionColumn, nameColumn]
+  );
+  return tableColumns;
+}
```
vidyanambiar, line 38: The awx implementation has 2 additional sortable columns:
`Related Rrsource` and `Next run`. Could we add those as well?

frontend/awx/views/schedules/hooks/useSchedulesFilter.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export function useSchedulesFilter() {
+  const nameToolbarFilter = useNameToolbarFilter();
+  const descriptionToolbarFilter = useDescriptionToolbarFilter();
+
+  const toolbarFilters = useMemo<IToolbarFilter[]>(
+    () => [nameToolbarFilter, descriptionToolbarFilter],
+    [nameToolbarFilter, descriptionToolbarFilter]
+  );
+  return toolbarFilters;
```
vidyanambiar, line 16: the awx implementation includes filters for `Created By` and `Modified By`. It would be good to have those too.

vidyanambiar (changes requested):
Works well. I added a couple comments related to feature parity with the awx schedules list.

vidyanambiar (commented):
It looks like the `Name` and `Related Resource` columns are duplicated. For eg. the related resource for `Cleanup Activity Schedule` should be `Cleanup Activity Stream`. Other than that, the additional sortable columns and filters look good to me.

image

vidyanambiar (approved):
lgtm!

PR #522, opened 2023-05-08
fix automation server toolbar action plus enhance cards

jamestalton:


vishsanghishetty (approved):
Automation servers page looking very nice 🚀

PR #533, opened 2023-05-09
[AAP-11987][AAP-12032] Remove users and rule audit filters. Roles not links in User select roles

lgalis:
Remove users and rule audit filters
The roles should not be links in the User Select roles modal:
Screenshot from 2023-05-10 13-22-11
Screenshot from 2023-05-10 13-21-59


AlexSCorey (approved):
Looks good to me.  I pulled down the code and everything works as expected.  I connected with Laura offline and we decided we will bring this feature in line with conventions at a later date.

PR #536, opened 2023-05-10
Page nav side bar context   to know when sidebar is open

jamestalton:
Creates a context for the navigation side bar so that code can know if the sidebar is open.
Hides the page subnavigation if the sidebar is open.


vidyanambiar (approved):
Works great ✨

PR #542, opened 2023-05-10
Adds query param support for language and pseduolocalization specification

mabashian:
Corresponds to https://github.com/ansible/awx/pull/13661 in AWX.  This allows a user to test out strings that have been marked for translation by adding `?pseudolocalization=true` to the query params.  It also allows the user to force a specific language by adding `?lang=[language code]`.


sshedmake:
👏 just that, tested this locally. Worked as expected.

jamestalton (approved):
lgtm

PR #547, opened 2023-05-11
Lines up supported locales with AWX

mabashian:
List of supported locales:

en es fr ja ko nl zh zu

This also moves the pseudolocalization from `fr` to `zu`


jamestalton (approved):
lgtm

PR #554, opened 2023-05-11
Fix job relaunch redirect and handle 202 responses

kialam:
Currently when relaunching a job the UI redirects to the old job output and not the job output that has been relaunched/is underway. This PR fixes the redirect.

job_redirect

Concurrently, I also discovered that the API on occasion will return a `202` with a response body (example POST request to `resource/id/update` endpoint). This PR handles the two cases where:
1. `202` with no body -> return `null`
2. `202` with response body -> return `response.json()`


keithjgrant (approved):

vidyanambiar (approved):
lgtm!

jamestalton (approved):
lgtm

PR #555, opened 2023-05-11
Add an AWX config provider and point to the correct doc links

vidyanambiar:
- Adds an AWX configuration provider. Components can get access to the config settings via the `useAwxConfig` hook.
- Adds a `getDocsBaseUrl` utility to get the base URL for docs based on the AWX version

### DIFF
--- a/frontend/awx/administration/execution-environments/ExecutionEnvironments.tsx
+++ b/frontend/awx/administration/execution-environments/ExecutionEnvironments.tsx
@@ -32,10 +32,13 @@ import { ExecutionEnvironment } from '../../interfaces/ExecutionEnvironment';
 import { useAwxView } from '../../useAwxView';
 
 import { useDeleteExecutionEnvironments } from './hooks/useDeleteExecutionEnvironments';
+import getDocsBaseUrl from '../../common/util/getDocsBaseUrl';
+import { useAwxConfig } from '../../common/useAwxConfig';
 
 export function ExecutionEnvironments() {
   const { t } = useTranslation();
   const navigate = useNavigate();
+  const config = useAwxConfig();
   const toolbarFilters = useExecutionEnvironmentsFilters();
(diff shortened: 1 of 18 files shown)


jamestalton (approved):
lgtm

kialam (changes requested):
Nice work on this! I pulled this down and tested locally. I'm seeing both the help text for the project details source control URL and project base path are not rendering properly:

Screenshot 2023-05-11 at 2 58 36 PM
Screenshot 2023-05-11 at 2 58 43 PM

Are you seeing this as well?

frontend/awx/administration/execution-environments/ExecutionEnvironments.tsx
```diff
@@ -98,7 +101,7 @@ export function ExecutionEnvironments() {
     <PageLayout>
       <PageHeader
         title={t('Execution Environments')}
-        titleDocLink="https://docs.ansible.com/automation-controller/latest/html/userguide/execution_environments.html"
+        titleDocLink={`${getDocsBaseUrl(config)}/html/userguide/execution_environments.html`}
```
kialam, line 104: The page header popover will not render without passing the `titleHelp` prop, so we won't see the link out to the docs. We don't have to fix this as part of this PR, however.

vidyanambiar:
> I'm seeing both the help text for the project details source control URL and project base path are not rendering properly:
> Are you seeing this as well?

Good catch! I am seeing this as well. However, I'm also seeing it on the `main` branch. Since it's not introduced by this PR  I can look into it separately.

frontend/awx/administration/execution-environments/ExecutionEnvironments.tsx
vidyanambiar: Thanks! Updated the titleHelp so that the popover appears and we can see the link to the docs.

kialam (approved):

PR #556, opened 2023-05-11
Add routed tabs to remaining awx resource pages

keithjgrant:
* Updates the following pages to RoutedTabs and adds persistent filters/back button: organizations, teams, inventories, projects
* Updates remaining awx routes from `.../edit/:id` to `.../:id/edit`


kialam (approved):
Pulled this down and tested locally. Very nice!

PR #559, opened 2023-05-15
Cleanup toolbar filters in preperation for advanced filter support

jamestalton:
Cleanup toolbar filters to have a common base interface.

Biggest change here is adding a comparison type to 
```
export interface IToolbarTextFilter extends ToolbarFilterCommon {
  /** Filter for filtering by user text input. */
  type: 'string';

  /** The comparison to use when filtering. */
  comparison: 'contains' | 'startsWith' | 'endsWith' | 'equals';
}
```

Which will need to be expanded on later when advanced filters need to use those.


vidyanambiar (approved):
lgtm

ZitaNemeckova (approved):

PR #561, opened 2023-05-16
[AAP-11983] Redirect to the login screen on 403 error on /users/me

lgalis:
EDA endpoints return 403 for unauthenticated users. This PR will redirect to the login screen if  /users/me results in 403.

This PR is for the proposed blocker  https://issues.redhat.com/browse/AAP-11983


jamestalton (changes requested):
We should not redirect on 403.
The server API should return a 401.

PR #573, opened 2023-05-19
Fixes devision env delete button label.  Makes token field in user -> Create Controller token a required field

AlexSCorey:
This addresses 2 separate eda issues

1. Fixes the delete button label on the decision environment list
2. Makes the token field in Users -> Create Controller Token form a required field.
Screen Shot 2023-05-19 at 9 11 23 AM
Screen Shot 2023-05-19 at 8 58 36 AM


vidyanambiar (approved):
lgtm

PR #575, opened 2023-05-19
Update rule audit to read activation_instance data

keithjgrant:
Update the interface for rule audits to use the activation_instance data field that's being added to the API

* depends on https://github.com/ansible/aap-eda/pull/260 *


jamestalton (approved):
lgtm

vidyanambiar (approved):
Looks good! 👍 
Just to confirm - this is still being added by the API and not available yet, correct? I don't see `activation_instance` in the response but I do see the following:
```
activation_instance_id: 603
activation_name: "My Rulebook Activation"
```

lgalis (approved):
LGTM

PR #578, opened 2023-05-23
[EDA] Fix broken user details and user tokens views for non-admins and non-auditors

marshmalien:
https://issues.redhat.com/browse/AAP-11573

* Fix bug where users without admin or auditor roles could not create or view their tokens
* Fix bug where users without admin or auditor roles could not view their details via the masthead dropdown
* Add `/users/me` and `/users/me/tokens` EDA routes

**When user does not have Admin role:**
* Hide user details edit and delete actions

**When user does not have Admin or Auditor roles:** 
* Hide Access sidebar group
* Hide users breadcrumb route
* GET user details from `api/eda/v1/users/me` instead of `api/eda/v1/users/:id`

**When user is viewing self** 
* Show Controller Tokens tab 

permissions

**Note**
Ideally I would've moved user, roles, and permissions into a context provider that wrapped the EventDriven page. Adding that new functionality would've remove duplicate calls to `EdaUserInfo()`, but it was not necessary for this bug fix and it bloated the PR.

### DIFF
--- a/frontend/eda/EventDrivenSidebar.tsx
+++ b/frontend/eda/EventDrivenSidebar.tsx
@@ -5,11 +5,18 @@ import { usePageNavBarClick } from '../../framework/PageNav/PageNavSidebar';
 import { RouteObj } from '../Routes';
 import { CommonSidebar } from '../common/CommonSidebar';
 import { isRouteActive } from '../common/Masthead';
+import { EdaUserInfo } from '../common/Masthead';
 
 export function EventDrivenSidebar() {
   const { t } = useTranslation();
   const location = useLocation();
   const onClick = usePageNavBarClick();
+  const activeUser = EdaUserInfo();
+  const canViewAccess =
+    activeUser &&
+    (activeUser.is_superuser ||
+      activeUser.roles.some((role) => role.name === 'Admin' || role.name === 'Auditor'));
+
   return (
     <CommonSidebar>
       <NavItem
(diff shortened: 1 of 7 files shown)


vishsanghishetty (approved):
lgtm 🚀

frontend/eda/EventDrivenSidebar.tsx
```diff
@@ -5,11 +5,16 @@ import { usePageNavBarClick } from '../../framework/PageNav/PageNavSidebar';
...
 import { CommonSidebar } from '../common/CommonSidebar';
 import { isRouteActive } from '../common/Masthead';
+import { EdaUserInfo } from '../common/Masthead';
 
 export function EventDrivenSidebar() {
   const { t } = useTranslation();
   const location = useLocation();
   const onClick = usePageNavBarClick();
+  const activeUser = EdaUserInfo();
+  const canViewAccess =
```
lgalis, line 15: The canViewAccess should also return true is the user is a superuser: activeUser?.is_superuser. The superuser does not have any roles returned.

lgalis (changes requested):
The User

lgalis (approved):
LGTM

frontend/eda/UserAccess/Users/EdaUserDetails.tsx
```diff
@@ -101,23 +117,37 @@ export function EdaUserDetails({ initialTabIndex = 0 }) {
     [deleteUsers, navigate, t]
   );
 
+  const activeUser = EdaUserInfo();
+  if (!activeUser) return <LoadingPage breadcrumbs tabs />;
+  const isViewingSelf = Number(user.id) === Number(activeUser.id);
+  const canEditUser = activeUser?.roles.some((role) => role.name === 'Admin');
```
lgalis, line 123: There should be a check for activeUser?.is_superuser here

frontend/eda/UserAccess/Users/EdaUserDetails.tsx
```diff
@@ -101,23 +117,37 @@ export function EdaUserDetails({ initialTabIndex = 0 }) {
...
+  const canViewUsers = activeUser?.roles.some(
```
lgalis, line 124: ...and here

lgalis (changes requested):
Please rebase as well so the e2e tests will pass and we can merge this PR.
Thank you!

PR #579, opened 2023-05-23
Adds support for importing all the available locales

mabashian:
I missed this when I added https://github.com/ansible/ansible-ui/pull/547.  Without this change we aren't able to load locales outside of `en` and `fr`


rooftopcellist (approved):

vidyanambiar (approved):
lgtm

PR #580, opened 2023-05-23
Move toolbar filters types to it's own files

ZitaNemeckova:
Before adding more possible types of filters a small clean-up.

Before/After:
Same


jamestalton (approved):
lgtm

PR #582, opened 2023-05-23
Clean up translations.

kialam:
The original cause of the bug was due to an `i18nKey` prop being provided to a block of text that was wrapped in the `<Trans>` component without its corresponding entry in translations.json. Upon further reading of the docs, it is stated that the `<Trans>` component is [rarely needed](https://react.i18next.com/latest/trans-component#important-note), and that most strings can be marked for translation using the provided `t` macro.

This PR cleans up some translations and help text blocks. Previously, the `<Trans>` component was used to wrap long help text blocks with formatting. I have used the `t` macro in areas where possible, and in other areas where strings include HTML text formatting, I have wrapped using the provided `<Trans>` component.

### DIFF
--- a/frontend/awx/dashboard/AwxDashboard.tsx
+++ b/frontend/awx/dashboard/AwxDashboard.tsx
@@ -40,12 +40,13 @@ export function AwxDashboard() {
     <PageLayout>
       {config?.ui_next && (
         <Banner variant="info">
-          <Trans>
-            <p>
-              <InfoCircleIcon /> You are currently viewing a tech preview of the new {product} user
-              interface. To return to the original interface, click <a href="/">here</a>.
-            </p>
-          </Trans>
+          <p>
+            <InfoCircleIcon />{' '}
+            <Trans>
+              You are currently viewing a tech preview of the new {{ product }} user interface. To
+              return to the original interface, click <a href="/">here</a>.
+            </Trans>
+          </p>
         </Banner>
       )}
       <PageHeader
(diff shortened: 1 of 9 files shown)


rooftopcellist (approved):

frontend/awx/dashboard/AwxDashboard.tsx
```diff
@@ -40,12 +40,13 @@ export function AwxDashboard() {
...
-          <Trans>
-            <p>
-              <InfoCircleIcon /> You are currently viewing a tech preview of the new {product} user
-              interface. To return to the original interface, click <a href="/">here</a>.
-            </p>
-          </Trans>
+          <p>
+            <InfoCircleIcon />{' '}
+            <Trans>
+              You are currently viewing a tech preview of the new {{ product }} user interface. To
```
vidyanambiar, line 46: Do we need a translation key here for the content within this Trans component to show up in the translation file?

kialam: We do not; the `i18nKey` prop is optional. I ran `npm run i18n` and it produced the following strings in the `translation.json` file:

```
"You are currently viewing a tech preview of the new {{product}} user interface. To return to the original interface, click <3>here</3>.": "You are currently viewing a tech preview of the new {{product}} user interface. To return to the original interface, click <3>here</3>.",
```

vidyanambiar: Great! Thanks for checking 👍

vidyanambiar (approved):
lgtm

PR #584, opened 2023-05-25
use styled components for divs

ShaiahWren:
Continue making styled div components to take the place of inline styling. 

The UI should show no changes.

### DIFF
--- a/framework/PageMasthead.tsx
+++ b/framework/PageMasthead.tsx
@@ -11,6 +11,15 @@ import {
 import { BarsIcon } from '@patternfly/react-icons';
 import { Children, ReactNode } from 'react';
 import { usePageNavSideBar } from './PageNav/PageNavSidebar';
+import styled from 'styled-components';
+
+const ToolbarSpacingDiv = styled.div`
+  flex-grow: 1;
+`;
+
+const StyledDiv = styled.div`
+  padding: 0px;
+`;
 
 export function PageMasthead(props: { title: string; brand?: string; children?: ReactNode }) {
   return (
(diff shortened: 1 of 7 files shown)


frontend/common/Masthead.tsx
```diff
@@ -180,13 +188,13 @@ export function AnsibleMasthead(props: { hideLogin?: boolean }) {
                                     <AppBarDropdown icon={<CogIcon />}>
                                         <DropdownGroup label="Theme">
                                             <DropdownItem
-                                                icon={theme === ThemeE.Light ? <SunIcon /> : <div style={{ width: 24 }} />}
+                                                icon={theme === ThemeE.Light ? <SunIcon /> : <SunIconGroupDiv />}
```
ZitaNemeckova, line 191: There's no `SunIconGroupDiv` or `MoonIconGroupDiv `. And you can use one definition for both `<div style={{ width: 24 }} />` replacements.

framework/PageMasthead.tsx
```diff
@@ -11,6 +11,15 @@ import {
 import { BarsIcon } from '@patternfly/react-icons';
 import { Children, ReactNode } from 'react';
 import { usePageNavSideBar } from './PageNav/PageNavSidebar';
+import styled from 'styled-components';
+
+const ToolbarSpacingDiv = styled.div`
+  flex-grow: 1;
+`;
+
+const ToolbarDiv = styled.div`
```
ZitaNemeckova, line 20: You can do 
```js
const StyledToolbar = styled(Toolbar)`
  padding: 0px;
`;
```

frontend/common/Masthead.tsx
ShaiahWren: Now there is an `IconGroupDiv` for both components, and it is defined in a styled component that has been commented out.

ZitaNemeckova (approved):
LGTM :+1:

jamestalton (approved):
lgtm

PR #588, opened 2023-05-30
Update command for creating AWX job template

vidyanambiar:
This test fails often in the CI checks: `jobs > deletes a job from the jobs list toolbar`

image

1. One issue was that we had switched to using the wrong support command for creating the job template resource (we were creating a template specific to EDA using an EDA specific playbook in an AWX specific test). We needed to switch this back to  creating a template using the hello_world playbook for AWX.
2. Regardless of the playbook being used, the launch job appears to terminate in `Error` with a one-line "Finished" job output. Does anyone have more information on why this would happen? I recreated this issue with a manually created template "Test Template" on our E2E server.
Since the failing test (delete job) does not require the job to be in success state (just that it should not be running), I have tweaked the test to proceed with deletion as long as the job is not running.

### DIFF
--- a/cypress/e2e/awx/views/jobs.cy.ts
+++ b/cypress/e2e/awx/views/jobs.cy.ts
@@ -5,19 +5,16 @@ import {
   JobTemplate,
   UnifiedJobList,
 } from '../../../../frontend/awx/interfaces/generated-from-swagger/api';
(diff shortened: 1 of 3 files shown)


AlexSCorey (approved):
Looks good to me.

PR #597, opened 2023-06-05
Delete jobs test: wait for job to come out of pending state

vidyanambiar:
PR builds randomly fail on the following tests:
Screenshot 2023-06-05 at 12 57 44 PM

1. deletes a job from the jobs list row
```
Timed out retrying after 30000ms: expected '<a.pf-m-icon.pf-m-aria-disabled.pf-c-dropdown__menu-item>' not to have attribute 'aria-disabled' with the value 'true', but the value was 'true'
```
2. deletes a job from the jobs list toolbar
```
409: Conflict
Resource is being used by running jobs.
```

These are caused by the Job being stuck in "Pending" state longer than 30s. This PR updates the test to wait for 2 min until the job is not "new", "pending", "waiting" or "running" to make sure that the delete operation can work on it.
I also removed the check for the presence of the "Cancel job" row action since the job may be in running state very briefly and we may or may not see this action at the time of assertion.

### DIFF
--- a/cypress/e2e/awx/views/jobs.cy.ts
+++ b/cypress/e2e/awx/views/jobs.cy.ts
@@ -66,9 +66,6 @@ describe('jobs', () => {
         cy.get('#relaunch-job').should('exist');
         cy.get('.pf-c-dropdown__toggle').click();
         cy.contains('.pf-c-dropdown__menu-item', /^Delete job$/).should('exist');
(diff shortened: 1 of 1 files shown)


vishsanghishetty (approved):
lgtm

vidyanambiar:
This fix (increasing timeout to 2min) may not work...jobs are stuck in Pending state for long periods (at least on the dev instance). Need to investigate a better solution

PR #601, opened 2023-06-06
Adds edit functionality to the job template form

AlexSCorey:
Adds Edit functionality to the job template form.  Also update the AWX Label select component to use new fetcher function


frontend/awx/interfaces/JobTemplateForm.ts
```diff
@@ -1,61 +1,60 @@
-import { ExecutionEnvironment } from './ExecutionEnvironment';
 import { InstanceGroup } from './InstanceGroup';
-import { Inventory } from './Inventory';
-import { Label } from './Label';
-import { Project } from './Project';
-import { Credential } from './Credential';
+import { JobTemplate } from './JobTemplate';
 
-export interface JobTemplateForm {
+export interface JobTemplateSummaryFields
```
keithjgrant, line 4: I think you can just do `type JobTemplateSummaryFields = Omit<...>` rather than using `extends`. Same for `Pick` with JobTemplateRelatedFields  below

keithjgrant (approved):

PR #606, opened 2023-06-07
Hub: parametrize API base path

himdel:
Hub API path is not always `/api/automation-hub`, we'll need to work with `/api` (community), `/api/galaxy` (standalone prod, some standalone dev), `/api/automation-hub` (insights prod, insights dev, most standalone dev envs).

TODO also add a task to run hub with localhost:5001 ootb


himdel:
or would it make sense to move these up to `useHubView` / `usePulpView` , so that we'd be passing just `tasks/` instead of `/api/automation-hub/pulp/api/v3/tasks/` ?

AlexSCorey (approved):
This seems pretty straight forward and ready to merge.  I'm assuming that tests will be added in the future.

PR #612, opened 2023-06-07
bump number of parallel test runners up for component tests

jamestalton:
We have reached that milestone where component tests are taking more than 10 minutes!
Adding a parallel builder to cut the test time down.

Screenshot 2023-06-07 at 3 16 23 PM


akus062381 (approved):

vishsanghishetty (approved):
🚀

PR #618, opened 2023-06-07
Remove link to project from project select

keithjgrant:
Removes the link from table of projects in the project select modal. This prevents the user from accidentally navigating away from the form when trying to select a project.

Screenshot 2023-06-07 at 2 44 09 PM


jamestalton (approved):
lgtm

PR #620, opened 2023-06-09
Enhance PR GitHub Action to only run tests for what has changed

jamestalton:
- Only runs E2E and CCT (Component Check Tests) when a change affecting them happens.
- New cypress dashboards have been setup in the paid cypress account - devs should now be able to access the test results. 

Screenshot 2023-06-09 at 2 13 09 PM


vishsanghishetty (approved):
🚀

PR #624, opened 2023-06-09
Add Clipboard copy to Code Block page details. Set it for the ExtraVars and Event payload

lgalis:
Add Clipboard copy to Code Block page details. Set it for the ExtraVars and Event payload
Screenshot from 2023-06-12 09-46-33


jamestalton (approved):
lgtm

PR #627, opened 2023-06-12
E2E tests: Address issue with filter selection failing randomly

vidyanambiar:
Fix for the following error that appears randomly on the "selectToolbarFilterType" Cypress command in `cypress/support/awx-commands.ts`:
image


vishsanghishetty (approved):
lgtm 🚀

PR #630, opened 2023-06-13
VITE: Bundler switch from Webpack to Vite

jamestalton:
Switch bundler from webpack to vite
- Significantly faster for development, production builds, and testing.


vishsanghishetty (approved):
🚀

.gitignore
```diff
@@ -2,8 +2,10 @@
 /build
 /certs
 /coverage
+/dist
 /node_modules
 /publish
 .DS_Store
 /storybook-static
 .vscode
+*.local
```
vidyanambiar, line 11: The files in the `locales/` folder appear as untracked. Can we add that to .gitignore?
image

vite.config.ts
```diff
@@ -0,0 +1,77 @@
...
+import { defineConfig } from 'vite';
+// import { VitePWA } from 'vite-plugin-pwa';
+import svgr from 'vite-plugin-svgr';
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  plugins: [
+    svgr({ exportAsDefault: true }),
+    react(),
+    // VitePWA({
```
vidyanambiar, line 11: Why is this plugin commented out?

vidyanambiar (approved):
Works great and so much faster than before 💯 🚀

.gitignore
jamestalton: the locales had to move into the public folder for vite to automatically pick them up. you should not need a locales folder at the root.

vite.config.ts
jamestalton: unfortunately something about that PWA plugin is causing the e2e tests to crash and not run.  It should not impact anything to have it disabled. I'll work on getting it renabled in a followon pr.

ZitaNemeckova:
@jamestalton I'd advise against this change as it introduces a new tool that is not as well known across teams as Webpack. Also I'm not sure it's not gonna introduce more problems with using the framework on c.r.c Maybe it's worth more attention from other teams to assess the risks?

cc @h-kataria

h-kataria:
@jamestalton has any other team tried to switch to Vite yet from webpack, want to know if they ran into any issues or had to make a significant amount of changes to support the new tool.

himdel:
My concern here would be the insights platform support. [frontend-components-config](https://github.com/RedHatInsights/frontend-components/tree/master/packages/config) is based on webpack, and won't work without it.

jamestalton:
Holding off on this PR until we can discuss it more as a platform team.

PR #632, opened 2023-06-13
Page carousel component

vidyanambiar:
Adds a `PageCarousel` to the shared UI framework. The component displays cards within it and switches smoothly between pages of cards. The carousel is responsive and displays a certain number of cards within it based on the screen size.

https://github.com/ansible/ansible-ui/assets/43621546/a4635326-5d39-4384-8bfd-e050df4becfb

### DIFF
--- a/framework/PageCarousel/PageCarouselNav.tsx
+++ b/framework/PageCarousel/PageCarouselNav.tsx
@@ -1,18 +1,28 @@
 import { Button, Bullseye, ButtonVariant } from '@patternfly/react-core';
 import { AngleLeftIcon, AngleRightIcon } from '@patternfly/react-icons';
-import { CSSProperties, useEffect } from 'react';
+import { useEffect } from 'react';
 import { useMemo } from 'react';
+import styled, { css } from 'styled-components';
 
-const pageDotStyle: CSSProperties = {
-  height: 10,
-  width: 10,
-  borderRadius: '50%',
-  backgroundColor: 'var(--pf-global--primary-color--100)',
-  padding: 0,
-  marginLeft: 10,
-  marginRight: 10,
-};
-
+const PageNavDotButton = styled(Button)`
+  && {
+    background-color: blue;
+    height: 10px;
+    width: 10px;
+    border-radius: 50%;
+    padding: 0;
+    margin-left: 10px;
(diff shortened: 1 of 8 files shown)


jamestalton (approved):
lgtm

framework/PageCarousel/PageCarouselNav.tsx
```diff
@@ -0,0 +1,82 @@
+import { Button, Bullseye, ButtonVariant } from '@patternfly/react-core';
+import { AngleLeftIcon, AngleRightIcon } from '@patternfly/react-icons';
+import { CSSProperties, useEffect } from 'react';
+import { useMemo } from 'react';
+
+const pageDotStyle: CSSProperties = {
```
keithjgrant, line 6: This could be converted to styled components to match our pattern elsewhere

keithjgrant (approved):

framework/PageCarousel/PageCarouselNav.tsx
vidyanambiar: @keithjgrant I had tried this but couldn't get the dots to render. And the custom styles are not getting applied to the PF Button. The other places that I used it I was not modifying a PF component. Could you look this code over please?

```
const PageNavDotButton = styled(Button)`
  height: 10;
  width: 10;
  border-radius: 50%;
  padding: 0;
  margin-left: 10;
  margin-right: 10;
  ${(props: { isCurrentPage: boolean }) =>
    props.isCurrentPage
      ? css`
          background-color: var(--pf-global--primary-color--100);
        `
      : css`
          background-color: var(--pf-global--disabled-color--200);
        `}
`;

...

  const pageDots = useMemo(() => {
    const dots = [];
    for (let i = 0; i < totalPages; i++) {
      dots.push(
        <PageNavDotButton
          key={i}
          variant={ButtonVariant.plain}
          onClick={() => {
            setPage(i);
          }}
          isCurrentPage={i === currentPage}
        ></PageNavDotButton>
      );
    }
    return dots;
  }, [currentPage, setPage, totalPages]);

```

image

vidyanambiar: @keithjgrant Thanks for helping me resolve this! I've updated the PR

PR #633, opened 2023-06-14
fix failing job template form e2e tests

jamestalton:


vidyanambiar (approved):
Thanks for getting the tests passing again! 🎉 

As discussed offline this E2E fix unblocks our PR builds but the commented out commands and tests will need to be revisited.

PR #637, opened 2023-06-14
Speed up job template test by only waiting for project sync after all resources are created

jamestalton:


cypress/e2e/awx/resources/jobTemplates.cy.ts
```diff
@@ -23,40 +23,42 @@ describe('Job templates form', () => {
...
-      // });
-
-      cy.createAWXCredential({
-        kind: 'machine',
-        organization: organization.id,
-        credential_type: 1,
-      }).then((cred) => {
-        machineCredential = cred;
-      });
+        // cy.createAWXCredential({
```
AlexSCorey, line 37: Please remove the commented code

cypress/support/awx-commands.ts
```diff
@@ -417,15 +417,18 @@ Cypress.Commands.add('deleteAwxUser', (user: User) => {
...
+  ) => {
     cy.awxRequestPost<Partial<Project>, Project>('/api/v2/projects/', {
       name: 'E2E Project ' + randomString(4),
       scm_type: 'git',
       scm_url: 'https://github.com/ansible/ansible-ui',
       ...project,
     }).then((project) => {
-      if (!skipSync) {
-        waitForProjectToFinishSyncing(project.id);
+      if (!options?.skipSync) {
```
AlexSCorey, line 430: I thing this set up is going to bite us.  This is an anti-pattern in controller as a project cannot be used anywhere until it is synced.  At a minimum we should have skipSync default to false.

jamestalton: it does default to false

cypress/e2e/awx/resources/jobTemplates.cy.ts
jamestalton: the commented code needs to be reenabled after it is made to work

PR #650, opened 2023-06-21
Add clipboard copy for the Output field in the activation History details page

lgalis:
Add clipboard copy for the Output field in the activation History details page
Screenshot from 2023-06-23 14-31-57

### DIFF
--- a/cypress/e2e/eda/Rulebook-Activations/rulebook-activations-crud.cy.ts
+++ b/cypress/e2e/eda/Rulebook-Activations/rulebook-activations-crud.cy.ts
@@ -103,6 +103,7 @@ describe('EDA rulebook activations- Edit, Delete', () => {
   let edaProject: EdaProject;
   let edaDecisionEnvironment: EdaDecisionEnvironment;
   let edaRBA: EdaRulebookActivation;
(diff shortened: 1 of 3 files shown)


cypress/e2e/eda/Rulebook-Activations/rulebook-activations-crud.cy.ts
```diff
@@ -136,13 +136,15 @@ describe('EDA rulebook activations- Edit, Delete', () => {
   it('can enable and disable a Rulebook Activation', () => {
     cy.navigateTo(/^Rulebook Activations$/);
     cy.filterTableByText(edaRBA.name);
+    cy.wait(10000);
```
vidyanambiar, line 139: 10s seems like a long wait here. What's the reason for the hardcoded wait? Can we avoid it?

lgalis: @vidyanambiar - this delay was here up until 2 days ago. This test was failing now,  so I just reversed the changes made 2 days ago that removed this delay to see if this is the reason for the failure. I agree the delay is long - was planning to look at some other options.

PR #635, opened 2023-06-14
Adds create functionality for schedule container form

AlexSCorey:
---
description: Adds functionality to create the schedule container...not re-occuranges yet

PR #651, opened 2023-06-22
Update page form select components for JT, WFJT, and Projects

AlexSCorey:
We have adopted a slightly different, yet new, convention for form lookups.  This work updates the lookup components for Job Templates, WorkflowJobTemplates, and for Projects, as each of those lookups will be in the schedule form that is up for PR [here](https://github.com/ansible/ansible-ui/pull/635)

### DIFF
--- a/frontend/awx/resources/templates/JobTemplateInputs.tsx
+++ b/frontend/awx/resources/templates/JobTemplateInputs.tsx
@@ -25,7 +25,6 @@ function JobTemplateInputs(props: { jobtemplate?: JobTemplateForm }) {
   const { t } = useTranslation();
   const { setValue } = useFormContext<JobTemplateForm>();
   const [playbookOptions, setPlaybookOptions] = useState<string[]>();
-  const project = useWatch<JobTemplateForm>({ name: 'project' });
   const projectPath = useWatch({
     name: 'summary_fields.project',
   }) as Project;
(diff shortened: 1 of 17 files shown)


frontend/awx/resources/templates/JobTemplateInputs.tsx
```diff
@@ -95,9 +94,9 @@ function JobTemplateInputs(props: { jobtemplate?: JobTemplateForm }) {
         isRequired={!isInventoryPrompted}
       />
       <PageFormProjectSelect<JobTemplateForm>
-        name="summary_fields.project.name"
+        name="summary_fields.project"
```
vidyanambiar, line 97: I'm seeing an issue in the usage of the PageFormProjectSelect component in the `frontend/awx/access/roles/AddRolesForm.tsx` form.

This particular error `cannot read properties of undefined (reading summary_fields)` is not seen with the old select component. 

image

vidyanambiar: I'm seeing a similar error with the JT and WFJT select components `Cannot read properties of undefined (reading 'summary_fields')`. On the `main` branch too there is a related error (`Cannot read properties of undefined (reading 'id')`) for the old JT select component.
image

image

vidyanambiar (changes requested):
The Projects select component works well in the Template form 👍 . 

I see some issues in the updated select components as used in the `AddRolesForm`.

vidyanambiar (approved):
Looks great!

PR #660, opened 2023-06-28
webpack.config: disable istanbul in development

himdel:
This is still enabled in cypress.config for tests,

but in dev mode, istanbul instrumentation breaks chrome console debugging, disabling.

### DIFF
--- a/webpack.config.cjs
+++ b/webpack.config.cjs
@@ -41,7 +41,6 @@ module.exports = function (env, argv) {
           test: /\.(ts|tsx|js|jsx)$/,
           exclude: /node_modules/,
           use: [
-            isDevelopment && 'coverage-istanbul-loader',
             {
               loader: 'babel-loader',
               options: {


himdel:
@MilanPospisil can you check this fixes the breakpoint problem for you as well?

MilanPospisil:
> @MilanPospisil can you check this fixes the breakpoint problem for you as well?

It really seems to work! Good Job Martin! Without debugger, it was quite hell to get into framework and discover what Im setting up wrong :)

webpack.config.cjs
```diff
@@ -41,7 +41,6 @@ module.exports = function (env, argv) {
           test: /\.(ts|tsx|js|jsx)$/,
           exclude: /node_modules/,
           use: [
-            isDevelopment && 'coverage-istanbul-loader',
```
vidyanambiar, line 44: Can we change this to `!isDevelopment && 'coverage-istanbul-loader'` instead of removing it entirely? I think removing this loader completely would prevent us from tracking coverage for E2E tests.
The cypress config file only uses this loader for component tests and not E2E tests.

vidyanambiar (changes requested):

webpack.config.cjs
vidyanambiar: The line you linked above is under `component: ` so it only applies to component tests. We should be able to collect coverage from E2E tests as well.

himdel: Updated with `isProduction` and a clarifying comment in case we're wrong :)

vidyanambiar (approved):
I'm good with this change. If you prefer to revert to your original commit (removing the istanbul loader altogether), that's fine  too. As discussed offline we can revisit this when we switch to Vite.

PR #664, opened 2023-06-29
[AAP-11992] Add iconAlign property to the TextCell component. External link icon to the right for the Action field

lgalis:
Add iconAlign property to the TextCell component. External link icon to the right for the Action field.

Screenshot from 2023-06-30 13-46-58


vishsanghishetty (approved):
lgtm

vidyanambiar (commented):
Looks good, thanks for adding this! 👍 

I see a slight difference when comparing to the mockups attached to the issue. The icon in the mockup appears to be gray and there's a little more padding between the icon and the text.

lgalis:
@vidyanambiar - I kept the icon as is when on the left, as I thin kthat should be consistent. We can discuss with UX if i the icon needs to be different or with a different spacing.

PR #667, opened 2023-06-30
[AAP-12147] Add clipboard copy to the git hash field in the rulebook activation details page

lgalis:
Add clipboard copy to the git hash field in the rulebook activation details page.
Screenshot from 2023-06-30 16-44-22


vishsanghishetty (approved):
lgtm 🚀

tiyiprh:
Hi @lgalis, is this using the same component as the UI library is? I made a similar comment on this issue: https://issues.redhat.com/browse/AAP-11826
But it looks like in the UI library, the field does not wrap causing the copy icon to be on the next line. Please advise. Thanks!

PR #670, opened 2023-07-03
[AAP-]12160] Add expanded row to the users list, containing the user's roles

lgalis:
Add expanded row to the users list, containing the user's roles

Screenshot from 2023-07-03 13-36-55


vishsanghishetty (approved):
lgtm 🚀

PR #672, opened 2023-07-03
[AAP-12212] Add create option to the credentials select dropdown

lgalis:
Add create option to the credentials select dropdown

Screenshot from 2023-07-03 16-11-25
Screenshot from 2023-07-03 16-11-00
Screenshot from 2023-07-03 16-09-15
Screenshot from 2023-07-03 16-09-08


vishsanghishetty (approved):
lgtm 🚀

PR #677, opened 2023-07-05
[AAP-12156] Switch Credentials and Decision Environments in the left hand Navigation

lgalis:
Switch Credentials and Decision Environments in the left hand Navigation.

Screenshot from 2023-07-05 14-37-02


vishsanghishetty (approved):
lgtm

PR #678, opened 2023-07-05
Framework component to reorder items (with selections)

vidyanambiar:
Framework component to reorder items in a table by dragging rows to the desired position. It also supports selection of items from the list.

This component is being used in the Manage View functionality on the hub dashboard (currently using mock data since we don't have the API endpoint for it yet).

For now, this can be tested by viewing the console log after applying changes in the Manage View modal:

image

It can also be utilized in the future for editing the order of template survey questions:

image


jamestalton (approved):
lgtm

PR #679, opened 2023-07-06
Pre-PatternFly 5 component cleanup

jamestalton:
Initial investigations into PatternFly 5 found that the PageFormTextInput caused the most issues because of inheriting props from FormGroupTextInputProps. PF5 renamed and removed some of those props. We still want PageFormTextInput to handle them, so made the props more explicit in PageFormTextInput which should make the upgrade to PF5 easier.

https://staging.patternfly.org/get-started/upgrade/


lgalis (approved):
LGTM

PR #686, opened 2023-07-07
[AAP-13688] Update field order in rulebook activation create/details pages

lgalis:
Update field order in rulebook activation create/details pages
Screenshot from 2023-07-07 16-37-35
Screenshot from 2023-07-07 16-37-28


vishsanghishetty (approved):
lgtm

PR #692, opened 2023-07-10
Remove unused components and rename page form select

jamestalton:
It removes some unused components. That will help make PF5 migration easier.
It also renames the PageFormSelectOption to PageFormSelect - because it was confusing.
A PageFormSelectOption was PageFormSelect that had options object array.
It sounded like PageFormSelectOption was an option you would pass to a PageFormSelect.


vishsanghishetty:
LGTM 🚀

vishsanghishetty (approved):

PR #694, opened 2023-07-11
Update getRequest to use a useCallback

jamestalton:


MilanPospisil:
Yes, this may be the solution. Because linter wanted to use inline function inside useCallback, but it was not possible outside of the useGetRequest. LGTM :+1:

MilanPospisil (approved):
LGTM :+1:

PR #695, opened 2023-07-11
Update the Automation Calculator Toolbar PR with typescript typing fixes

jamestalton:
Changes the Toolbar `setFilters` from
```
setFilters: Dispatch<SetStateAction<Record<string, string[]>>>
```
to 
```
setFilters: SetFilterState
```
where `SetFilterState`
```
/** Function to set the filters using an update function that takes in the current filters and returns the new filter state.
 * This follows the pattern of react useState setter functions so that it can be used with the useState hook. */
export type SetFilterState = (updateFilters: (currentFilters: FilterState) => FilterState) => void;
```
so that the toolbar filters can be used even when not using a `useState` hook to manage filter state.

### DIFF
--- a/framework/PageTable/PageToolbar/PageToolbarFilter.tsx
+++ b/framework/PageTable/PageToolbar/PageToolbarFilter.tsx
@@ -10,27 +10,32 @@ import {
   ToolbarToggleGroup,
 } from '@patternfly/react-core';
 import { FilterIcon } from '@patternfly/react-icons';
-import { Dispatch, SetStateAction, useState } from 'react';
+import { useState } from 'react';
 import { FormGroupSelect } from '../../PageForm/Inputs/FormGroupSelect';
 import { useBreakpoint } from '../../components/useBreakPoint';
-import { ToolbarTextFilter, IToolbarTextFilter } from './PageToolbarFilterTypes/ToolbarTextFilter';
-import {
-  ToolbarSelectFilter,
-  IToolbarSelectFilter,
-} from './PageToolbarFilterTypes/ToolbarSelectFilter';
-import { IToolbarDateFilter } from './PageToolbarFilterTypes/ToolbarDateFilter';
 import { useFrameworkTranslations } from '../../useFrameworkTranslations';
+import { IToolbarDateFilter } from './PageToolbarFilterTypes/ToolbarDateFilter';
+import {
+  IToolbarSelectFilter,
+  ToolbarSelectFilter,
+} from './PageToolbarFilterTypes/ToolbarSelectFilter';
+import { IToolbarTextFilter, ToolbarTextFilter } from './PageToolbarFilterTypes/ToolbarTextFilter';
 
 export type IToolbarFilter = IToolbarTextFilter | IToolbarSelectFilter | IToolbarDateFilter;
 
-export type IFilterState = Record<string, string[] | undefined>;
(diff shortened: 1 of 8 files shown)


framework/PageTable/PageToolbar/PageToolbarFilter.tsx
```diff
@@ -99,16 +105,26 @@ function ToolbarContent(props: PageToolbarFiltersProps) {
...
-                let values = filters?.[selectedFilter];
-                if (!values) values = [];
-                if (!values.includes(value)) values.push(value);
-                setFilters?.({ ...filters, [selectedFilter]: values });
+                const selectedFilterCurrentValue = filters?.[selectedFilter];
+                const newFilterValue = selectedFilterCurrentValue
+                  ? [...selectedFilterCurrentValue]
+                  : [];
+                if (!newFilterValue.includes(value)) newFilterValue.push(value);
+                setFilters?.((currentFilters) => ({
```
ZitaNemeckova, line 113: It seems this is never called. Not sure why.

framework/PageTable/PageToolbar/PageToolbarFilter.tsx
```diff
@@ -99,16 +105,26 @@ function ToolbarContent(props: PageToolbarFiltersProps) {
...
-                let values = filters?.[selectedFilter];
-                if (!values) values = [];
-                values = values.filter((v) => v !== value);
-                setFilters?.({ ...filters, [selectedFilter]: values });
+                const selectedFilterCurrentValue = filters?.[selectedFilter];
+                let newFilterValue = selectedFilterCurrentValue
+                  ? [...selectedFilterCurrentValue]
+                  : [];
+                newFilterValue = newFilterValue.filter((v) => v !== value);
+                setFilters?.((currentFilters) => ({
```
ZitaNemeckova, line 124: Neither is this.

framework/PageTable/PageToolbar/PageToolbarFilter.tsx
ZitaNemeckova: Now it's reloading the chart all the time.

frontend/awx/analytics/Reports/AutomationCalculator.tsx
```diff
@@ -87,38 +87,41 @@ export default function AutomationCalculator(props: { schema: ChartSchemaElement
...
-      'total_count',
-      'total_org_count',
-      'total_cluster_count',
-      'successful_hosts_total',
-      'successful_elapsed_total',
-    ],
-    group_by: 'template',
-    group_by_time: false,
-  };
+  const defaultParams: ParamsType = useMemo(
```
ZitaNemeckova, line 90: Why is `useMemo` used here? This is static with no expensive function.

jamestalton: defaultParams is used inside
```
  const getParams = useCallback(() => {
```

If defaultParams is not using useMemo then we get an error
```
The 'defaultParams' object makes the dependencies of useCallback Hook (at line 138) change on every render. Move it inside the useCallback callback. Alternatively, wrap the initialization of 'defaultParams' in its own useMemo() Hook.eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)
```

PR #710, opened 2023-07-14
Toolbar - SingleSelectFilter, MultiSelectFilter, DateRangeFilter, pinned filters

jamestalton:
- Filters for selection now have two types
  - type: "select"
  - type: "multiselect"
- Added a DateRangeFilter
- Fixed bug where selections could not be removed using the chips
- Fixed bug where only one pinned filter was being displayed at at time
- Fixed the filter select dropdown to show the descriptions on options
- Enhanced the wrapping of the toolbar on small screens to wrap more consistently especially with pinned filters
- Enhanced the FormGroupSelect so that it takes in an optional selectedIcon
  - This is needed to match PF guidelines where selecting a filter shows the filter icon, but pinned filters do not have an icon
  
  Example of multiple pinned single select filters:
  
Screenshot 2023-07-14 at 11 07 28 AM

Example of wrapping:

Screenshot 2023-07-14 at 11 07 57 AM

Example of filters wrapping:

Screenshot 2023-07-14 at 11 08 28 AM

There will need to be a follow on PR to enhance the wrapping even more. PF guidelines indicate when wrapping that the filters should show up below the other components. That will take more than CSS, it will take javascript with a resize observer and special reordering of the toolbar items when the toolbar wraps.

https://github.com/ansible/ansible-ui/assets/6277895/cb894af3-762e-4caa-b296-1bee6b0e8765


ZitaNemeckova:
```
frontend/awx/views/jobs/hooks/useJobsFilters.tsx
51:        type: 'select',
```
This used to allow select multiple values. Is this on purpose?

ZitaNemeckova:
>> Fixed bug where only one pinned filter was being displayed at at time

This is misleading. Multiple filters could be pinned. They would show in the dropdown.

Before:
Screenshot 2023-07-17 at 13 47 30
Screenshot 2023-07-17 at 13 47 42

After:
Screenshot 2023-07-17 at 13 48 20

jamestalton:
Everything is working good except for zindex of the opened menu items on the toolbar.
Working through that.

skochay330 (approved):

PR #711, opened 2023-07-14
Reflect updated order in selected items

vidyanambiar:
The selected items from the `ReorderItems` component do not reflect the last updated order. This PR fixes that.

This can be tested with the Manage View modal that shows up in the Hub dashboard. The console log reflects the values of `updatedOrder` and `selectedCategories`. `selectedCategories` should match the order of items in `updatedOrder`.


kialam (changes requested):
This works for dragging to reorder items - awesome!
We will want this for the `onSelectItem` (and probably `onSelectAllItems`) handlers as well. Currently if you drag to reorder, then select items, the updated order does not get reflected.
drag-reorder-select

vidyanambiar:
@kialam Good catch! I updated `onSelectItem` so selecting after dragging to reorder should work correctly now. I did not need to update `onSelectAllItems` since it uses `listItems` to build the selected items list and `listItems` contains the latest updated order of items.

kialam (approved):
🥇 Excellent work!

PR #713, opened 2023-07-19
add support to the useView to support extre querystring values withou…

jamestalton:
Removes the `defaultFilters` from the view as it should not be part of that state, and instead uses the value that is passed into the useView hook which is used for the default state values of the hook.

Updated the useView params to be clearer.

Add new values:
- ignoreQueryStringKeys
- filterQueryStringKeys

Future PR will update the useHubView and usePulpView to pass in the `filterQueryStringKeys` by getting those from the `toolbarFilters?: IToolbarFilter[]`

Added jsdoc to the useView.


MilanPospisil:
I have several questions:
1) ignoreQueryStringKeys will be also added into useHubView and usePulpView in future PR?
2) What is the purpose of filterQueryStringKeys? Is it for filters that can not be deleted by clearAllFilters?
3) Why is the useView deleting all parameters except those who are ignored or default? We know what query strings are the view uses, why does we not delete only filter query strings for example when clearAllFilters is clicked? And the rest query strings for paging, ordering and stuffs can be only modified. If we would not delete all query string, we would not need any ignoreQueryStringKeys at all.
4) This is only my personal opinion and I dont think we necessary need it in our app and it will take some time to reimplement it - I only wander whats your opinion about this: What about query strings with prefixes for each component (or some objects of parameters for component)? I have seen it used and it easily solves the problem with different components like grids that are using query string as well. But, I understand that those query strings are not that human readable as those.

jamestalton:
Added better comments to address #1 and #2

```
  /**
   * An array of query string keys to ignore when updating the query string for filters.
   * This is the opposite of filterQueryStringKeys and only one should be used.
   *
   * Example:
   *   if ignoreQueryStringKeys=['abc'] then clearing filters would not clear 'abc' from the query string
   *   this would allow pages to have querystirng values that are not cleared by filters.
   */
  ignoreQueryStringKeys?: string[];

  /**
   * An array of query string keys that represent the filter state in the query string.
   * This is the opposite of ignoreQueryStringKeys and only one should be used.
   *
   * This is useful when the filters are known up front.
   * In most of the framework, the filter keys are known up front and can be passed in,
   * utilizing this option.
   */
  filterQueryStringKeys?: string[];
```

AWX,EDA and HUB would most likely use `filterQueryStringKeys` in a future PR. Not sure if `ignoreQueryStringKeys` will be used, but it is there if we need it.

jamestalton:
> 3. Why is the useView deleting all parameters except those who are ignored or default? We know what query strings are the view uses, why does we not delete only filter query strings for example when clearAllFilters is clicked? And the rest query strings for paging, ordering and stuffs can be only modified. If we would not delete all query string, we would not need any ignoreQueryStringKeys at all.

The view was clearing the filters that were not knows as the filters passed in could specify the keys they wanted to use in the query string and at that point in development, it was easier to clear filters by removing all query string value other than the default ones.

The right fix is this PR plus a follow on PR to only clear the filter keys that are actually in the filters provided.

PR #727, opened 2023-07-25
Job output follow mode & scroll buttons

keithjgrant:
Adds follow mode & pagination buttons to the job output page

* Clicking Unfollow exits follow mode
* Clicking page up or scroll to top buttons exits follow mode
* Scrolling manually does *not* exit follow mode -- I decided I cannot rely on `onScroll` events to exit follow mode, because follow mode itself fires off so many of them and I could not reliably distinguish between those and user-driven scroll events
* Clicking the scroll to bottom button enables follow mode. This enables the UI to scroll to bottom repeatedly so it stays at the bottom while the last events load and render.
* If the job is not running and follow mode is enabled, it will wait until the output has been locked to the bottom consistently for 600ms (checking every 200ms) -- allowing for the final events to render & resize properly -- then it will automatically disable follow mode.

### DIFF
--- a/frontend/awx/views/jobs/JobOutput/JobOutput.cy.tsx
+++ b/frontend/awx/views/jobs/JobOutput/JobOutput.cy.tsx
@@ -27,13 +27,13 @@ describe('JobOutput.cy.tsx', () => {
   });
 
   it('renders job output', () => {
-    cy.mount(<JobOutput job={job as unknown as Job} />);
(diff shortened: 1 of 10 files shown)


jamestalton (approved):
lgtm

akus062381 (changes requested):
@keithjgrant I am testing this now. Upon completion of a job, scrolling up within the output window doesn't work. I can show you if you would like.

vidyanambiar (approved):

akus062381 (approved):

PR #742, opened 2023-08-01
AWX:  Job view use columns for the expended content

jamestalton:
This converts the job view expanded content to use the columns.
This allows for users to use the column management to tweak the placement of the columns in the table, list, and cards.


vidyanambiar (approved):
lgtm!

PR #752, opened 2023-08-04
Customize Awx Dashboard

kialam:
This PR adds changes to allow users to customize their dashboard view.

dashboard-customize


vidyanambiar (approved):
Looks great ✨ 
 
In some combinations of manage view settings, a card may seem a little out of place - for instance inventories in this example shows up in a column of its own. It would be good to get UXD's recommendation for this. This seems to be caused by the width of the underlying dashboard cards for the donut charts where the size `sm` is smaller than the other dashboard cards. If we check how many donut charts are appearing consecutively then we could adjust the width accordingly.
image
image

jamestalton (approved):
lgtm

PR #755, opened 2023-08-04
PageSingleSelect and PageMultiSelect add open search and support dyna…

jamestalton:
Adds search when options length > 10.
Modifies the styling so that the dropdown will make it's width fit the components in the dropdown.

Screenshot 2023-08-04 at 12 46 25 PM


nixocio (approved):
LGTM,

We could include a negative test case that verify that no search input is rendered if the options is smaller than 10.

Also, for cases with more than 10 options should we select the option for the user if the user type the whole option value, and press return.

image

I pressed return on the case above and we did not select the option. 

@jamestalton minor comments.

PR #760, opened 2023-08-07
[AAP-11836] Add hasCheckIcon to the Page Action Switch

lgalis:
Add hasCheckIcon to the Page Action Switch

Screenshot from 2023-08-07 18-02-39

### DIFF
--- a/framework/PageActions/PageActionSwitch.tsx
+++ b/framework/PageActions/PageActionSwitch.tsx
@@ -42,6 +42,7 @@ export function PageActionSwitch<T extends object>(props: {
             label={label}
             labelOff={labelOff}
             isChecked={isChecked}
+            hasCheckIcon={true}
             isDisabled={isDisabled ? true : false}
             onChange={() => {
               switch (action.selection) {


jamestalton:
I'm checking with UXD if we might just want to turn in on for all switches. The extra flag might make it so that devs forget that flag and then the switches across the UIs would look different.

tiyiprh (approved):
LGTM!

framework/PageActions/PageActionSwitch.tsx
```diff
@@ -42,6 +42,7 @@ export function PageActionSwitch<T extends object>(props: {
             label={label}
             labelOff={labelOff}
             isChecked={isChecked}
+            hasCheckIcon={true}
```
vidyanambiar, line 45: Is this needed? I already see the check icon on action of type PageActionType.Switch

lgalis: @vidyanambiar - the flag is necessary for the top of the page action swicth - as seen in the screenshot included in the PR's description.

jamestalton (approved):
lgtm

PR #761, opened 2023-08-08
[AAP-14383] Remove toolbar from the controller token page. Switch to list view.

lgalis:
Remove toolbar from the controller token page. Switch to list view.

Screenshot from 2023-08-07 21-00-47


tiyiprh (approved):
LGTM!

PR #769, opened 2023-08-09
PageToolbar update ToolbarSingleSelect filter and ToolbarMultiSelect filter

jamestalton:
- Moved the PageToolbar directory up the root of the framework
- Renames PageTableToolbar to PageToolbar
- Breaks apart the IPageToolbarSingleSelectFilter and IPageToolbarMultiSelectFilter into separate files
- Updates IPageToolbarSingleSelectFilter to use the new PageSingleSelect component
- Updates IPageToolbarMultiSelectFilter to use the new PageMultiSelect component
- Moved the components that analytics use for toolbar out of the framework as they are not hooked up to the way the framework toolbar is handling filter state.


framework/PageToolbar/PageToolbarFilter.tsx
```diff
@@ -281,35 +272,24 @@ function ToolbarFilterComponent(props: {
 
     case ToolbarFilterType.SingleSelect:
       return (
-        <ToolbarSelectFilter
+        <ToolbarSingleSelectFilter
           id={props.id ?? filter.key}
-          values={values}
-          addFilter={addFilter}
-          removeFilter={removeFilter}
+          placeholder={filter.placeholder ?? 'TODO'}
```
nixocio, line 277: Could we possibly provide some additional context in this situation? It would be great if the TODO could include enough information so that another person could comfortably carry out the task if needed, or even reference a related issue.

framework/PageToolbar/PageToolbarFilter.tsx
```diff
@@ -281,35 +272,24 @@ function ToolbarFilterComponent(props: {
...
 
     case ToolbarFilterType.MultiSelect:
       return (
-        <ToolbarSelectFilter
+        <ToolbarMultiSelectFilter
           id={props.id ?? filter.key}
-          values={values}
-          addFilter={addFilter}
-          removeFilter={removeFilter}
+          placeholder={filter.placeholder ?? 'TODO'}
```
nixocio, line 289: Same comment as above for TODO.

nixocio (approved):
Good job @jamestalton I left minor comments.

framework/PageInputs/PageMultiSelect.cy.tsx
```diff
@@ -92,32 +92,6 @@ describe('PageMultiSelect', () => {
     cy.get('#test').should('have.text', placeholderText);
   });
 
-  it('should support string options', () => {
```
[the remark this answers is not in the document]
jamestalton: Those values were a problem and were hiding a bug underneath.
You could set the value={organization} but then options = [1,2] which does not overlap.
Because of the casting in the code, it would ignore that the types did not match.

framework/PageInputs/PageSingleSelect.cy.tsx
```diff
@@ -110,7 +86,6 @@ describe('PageSingleSelect', () => {
   });
 
   it('should show footer', () => {
-    const options = ['abc', 'def'];
     cy.mount(
       <PageSingleSelectTest placeholder={placeholderText} options={options} footer="Footer" />
```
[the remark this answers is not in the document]
jamestalton: it's using the global ones for the tests.
fine for this test as this is just showing testing the footer shows.

framework/PageToolbar/PageToolbarFilter.tsx
jamestalton: fixed thanks!

framework/PageToolbar/PageToolbarFilter.tsx
jamestalton: fixed thanks!

PR #777, opened 2023-08-12
feat: add new column for activation id to the list and details page

kurokobo:
## Pull Request Summary

This PR adds `Activation ID` column to `Rulebook Activation` page and details page for each activations:
  
image
image
image

## This Pull request provides

- Enhancement

## Additional information

Currently, the only way to check the Activation ID is to look at the number in the URL in the Web UI or use the API.
The Activation ID is important information when looking for Activation Job Pods or Service resources created by the EDA Server, or when calling some process with the API, and it would be nice if it were displayed in an easy-to-see location in the Web UI.

However, since this information is not necessary for all users, it is displayed in a less conspicuous position in this PR, especially on the list page.

I know this is not necessarily a change that will be acceptable to all, so feel free to reject this if you don't like this change 😃


kurokobo:
@jamestalton @lgalis 
Hi, thanks for moving my PR forward!
Do you have any ideas as to the cause of the failure of the CI?: https://github.com/ansible/ansible-ui/actions/runs/5841715214?pr=777

I am not very familiar with Cypress, but I find it hard to believe that the failure was caused by the changes I made to my PR.

Anyway, I have rebased my branch. Thanks in advance.

jamestalton:
The Cypress tests are failing because of GitHub security settings that are keeping the test jobs from access repository secrets. We will review, verify and then we can merge it.

Thanks for the PR!

tiyiprh (approved):
LGTM

PR #778, opened 2023-08-14
Milan collection list

MilanPospisil:
Collection list, delete collection from system, delete collection from repository, deprecate collection.

Collection detail not working - it will only show itself and no tab is working.

TODO in list -
Error handling  not implemented yet, we have to discuss it in meeting.


jamestalton (approved):
lgtm

PR #779, opened 2023-08-14
useManagedItems fixes for Dashboards and Table Columns

jamestalton:
Changes the way item state save being saved and loaded from local storage to fix issues with ordering and selection.


vidyanambiar (approved):
works great! 🚀

kialam (approved):
Pulled this down locally and was able to test show/hide and reorder working when navigating way from Dashboard and then back. 👍

PR #781, opened 2023-08-14
Updates to dashboard chart PR

jamestalton:
- Generate legend data from data groups to keep it in sync for developers easier
- Support both stacked and line charts: defaults to stacked, useLines to use line
- Removed sizing by percentage `width={size.height * 0.1}`. This will not scale well with different sizes.

Screenshot 2023-08-14 at 2 12 31 PM


ZitaNemeckova:
Done as a part of addressing review.

PR #786, opened 2023-08-15
Add http client to hub to handle the task system

nixocio:
Certain HTTP actions from the Pulp API returns a 202, requiring us to parse the task system to verify the task status.

Abstract the HTTP client to handle tasks as part of its functionality. This removes the need to check if a specific endpoint will return a 202 or not.

The `requestCommon` function from the framework doesn't provide the status code in its response.

Unlike modifying the existing `requestCommon` from the framework, which would necessitate changes in multiple files, we've introduced a dedicated HTTP client to address HUB-specific requirements.

See: https://pulp.plan.io/issues/3166


frontend/hub/api/request.ts
```diff
@@ -0,0 +1,134 @@
...
+  body?: BodyInit;
+  signal?: AbortSignal;
+}
+
+interface Response<T> {
+  statusCode: number;
+  response: T;
+}
+
+export async function request<T extends object | TaskResponse>(
```
jamestalton, line 18: These seem like generic function names that are in a hub/api file.
Maybe they should be hub specific names like hubRequest.
How do there differ from the common functions like postRequest that is the in common folder?

jamestalton (changes requested):

frontend/hub/api/request.ts
nixocio: I will rename them to reflect that it is specific to the HUB project. The `request` function for the HUB deals with the 202 as a response, and either return the task response or the generic one passe to the function.

frontend/hub/api/utils.tsx
```diff
@@ -0,0 +1,222 @@
...
+ * @throws {Error} Throws an error if the task href is invalid or if the task did not finish within the specified retries.
+ * @throws {Error} Throws an error with task error description if the task fails with an error.
+ * @throws {Error} Throws an error if the task is not found (status code 404).
+ */
+export async function waitForTask(
+  taskHref: string | null,
+  signal?: AbortSignal,
+  minDelay = 5,
+  maxDelay = 1000,
+  retries = 20
```
himdel, line 181: 5ms is too low for minDelay, and so is 1s for maxDelay

also, this will wait 20s max, this needs to keep trying for at least several minutes, without overloading the server by too many requests

himdel:
What's the difference between

`postHubRequest`, `patchHubRequest`, `putHubRequest`, `deleteHubRequest` and `getHubRequest` (`api/request.ts`)

and

`hubAPIDelete`, `hubAPIPatch`, `hubAPIPut` (`api/utils.ts`)?

---

Also, what's the story of error handling, is there any convention? Do these consistelny return a rejected promise with ... a string error? An error object from the http response? Mix of these?

frontend/eda/Resources/credentials/hooks/useDeleteCredentials.tsx
```diff
@@ -2,7 +2,7 @@ import { useCallback, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { compareStrings, useBulkConfirmation } from '../../../../../framework';
 import { requestDelete } from '../../../../common/crud/Data';
-import { idKeyFn } from '../../../../hub/api';
+import { idKeyFn } from '../../../../hub/api/utils';
```
vidyanambiar, line 5: It would be good to move `idKeyFn` from [frontend/hub/api/utils](https://github.com/ansible/ansible-ui/pull/786/files#diff-7a0a3bfbfe4f1fbef77ff3b622048ac41d5c689af723e357d26380e79b82d605R94) to `frontend/common/utils` if we're looking to access this util function in EDA files such as this one.

frontend/hub/api/request.ts
```diff
@@ -0,0 +1,137 @@
...
+  body?: BodyInit;
+  signal?: AbortSignal;
+}
+
+interface Response<T> {
+  statusCode: number;
+  response: T;
+}
+
+async function hubRequestCommon<T extends object | TaskResponse>(
```
vidyanambiar, line 18: I'd recommend adding JSDoc to this function to describe why it was needed (the 202 specific use case).

frontend/hub/api/utils.tsx
nixocio: @himdel Thanks for the review, I will update as suggested.

nixocio:
> What's the difference between
> 
> `postHubRequest`, `patchHubRequest`, `putHubRequest`, `deleteHubRequest` and `getHubRequest` (`api/request.ts`)
> 

Those are initial abstractions around `hubRequestCommon` and `hubRequestCommon` returns the status code of HTTP and its response. They are necessary since the `requestCommon` from the framework does not return the status code. If later on we have new HTTP clients returning the status code and the response those related to hub can be removed.
 
> and
> 
> `hubAPIDelete`, `hubAPIPatch`, `hubAPIPut` (`api/utils.ts`)?

`hubAPIDelete`, `hubAPIPatch`, `hubAPIPut`  abstract the parse of the 202, querying the task system if a 202 response is returned.

> 
> Also, what's the story of error handling, is there any convention? Do these consistently return a rejected promise with ... a string error? An error object from the http response? Mix of these?

jamestalton (approved):
lgtm

PR #789, opened 2023-08-16
Approvals reject

MilanPospisil:
Adds rejection button into the approvals page.

The reject works for single collection or for multiple collections as bulk operation.

The rejection will move collection from the selected repository into rejected repository. Thus if rejecting the same collection from multiple repositories, the collection will appear only once in rejected.

Not implemented yet - error handling and RBAC. We are currently discussing in the team how to do it in framework, because our API returns task, rather then task completition error.

image


jamestalton (approved):
lgtm

PR #791, opened 2023-08-16
Remove duplicate Launch interface definition

marshmalien:
##### SUMMARY
* **Remove Launch interface**: The Launch interface was removed because it was a close duplicate to the LaunchConfiguration interface. There were fewer instances of Launch between the two interfaces, so it was cut.
* **Add WorkflowJobLaunch interface**: This interface will represent launch configurations for workflow job templates. 
* **Add credential_passwords to LaunchConfiguration interface**: Add credential password properties and use template literal to define keys that follow `vault_password.${unknown_key}` pattern.
```
{
    'credential_passwords': {
        'vault_password.X': 'first-vault-password'
        'vault_password.Y': 'second-vault-password'
    }
}
```

* **TypeScript cleanup**: Swapping the Launch and LaunchConfiguration interfaces introduced TypeScript errors. Cleaned up ts errors in launchHandlers and scheduleHelpers. 

##### ISSUE TYPE
* Refactor


AlexSCorey (approved):
This looks good to me

PR #794, opened 2023-08-17
Milan async toolbar filters

MilanPospisil:
Async dropdown with lazy loading of items has been already implemented into framework.

This PR uses it also in filters. Example was created in Approvals list, where you can now filter also by repository.

image


jamestalton (approved):
lgtm

PR #798, opened 2023-08-21
Add browse to async single filter

MilanPospisil:


framework/PageToolbar/PageToolbarFilter.tsx
```diff
@@ -295,6 +296,18 @@ function ToolbarFilterComponent(props: {
           queryErrorText={filter.queryErrorText}
           queryPlaceholder={filter.queryPlaceholder}
           isRequired={filter.isRequired}
+          footer={
```
jamestalton, line 299: This looks like it would add a browse button even if the filter did not have an `openBrowse` function hooked up.

framework/PageToolbar/PageToolbarFilters/ToolbarAsyncSingleSelectFilter.tsx
```diff
@@ -18,7 +20,10 @@ export interface IToolbarAsyncSingleSelectFilter extends ToolbarFilterCommon {
   queryErrorText?: PageAsyncQueryErrorTextType;
 
   // useHook for modal here
-  openBrowse?: (onSelect: (value: string) => void, defaultSelection?: string) => void;
+  openBrowse?: openBrowseType;
+
+  // default selected item
+  defaultSelection?: string;
```
jamestalton, line 26: When would the defaultSelection be specified as part of the filter itself?
It seems like the `openBrowse` should be taking in the current selection, which is loaded from the querystring values for the filter.

framework/PageToolbar/PageToolbarFilter.tsx
```diff
@@ -295,6 +296,18 @@ function ToolbarFilterComponent(props: {
...
+            <Button
+              variant="link"
+              onClick={() => {
+                filter?.openBrowse?.((selection) => {
+                  addFilter(selection);
+                }, filter.defaultSelection);
```
jamestalton, line 305: It seems like the `openBrowse` should be taking in the current selection, which is loaded from the querystring values for the filter, instead of a filter.defaultSelection.

jamestalton (changes requested):

framework/PageToolbar/PageToolbarFilter.tsx
MilanPospisil: Aahh, yes, my bad, thanks for catching that.

framework/PageToolbar/PageToolbarFilter.tsx
MilanPospisil: Yes, filter default selection is now obsolete.

PR #805, opened 2023-08-22
Hide host metrics visibility based on proper licensing

AlexSCorey:
Hides Host Metrics view for those that do not have the proper licensing


jamestalton (approved):
lgtm

PR #809, opened 2023-08-23
Fix EDA Container Image to support OpenShift

jamestalton:


aknochow (approved):
Tested in openshift 4.11 successfully

PR #812, opened 2023-08-23
bump dependencies

keithjgrant:
Bumps to semver to patch 6.3.1


vidyanambiar (approved):
lgtm

jamestalton (changes requested):
Looking at the package-lock.json, ansible-ui is already using
```
    "node_modules/semver": {
      "version": "6.3.1",
    },
```
I do not thing  the changes change anything as that already resolved to that version. i.e. "semver": "^6.3.0" resolves to "6.3.1" since that is the one the package-lock installs and matches the range.

Running `npm audit` does not show any issues with semver except for an old old version included by browserfy-fs -> levelup -> "semver": "~2.3.1".

PR #815, opened 2023-08-24
Add cypress command to verify that required envs variables are set

nixocio:
Add cypress command to verify that required envs variables are set

image

### DIFF
--- a/.github/workflows/on-pull-request-run-tests.yml
+++ b/.github/workflows/on-pull-request-run-tests.yml
@@ -3,6 +3,9 @@ name: On Pull Request - Run Tests
 on:
   pull_request:
 
+env:
+  IS_GITHUB_ACTION: true
+
 concurrency:
   group: ${{ github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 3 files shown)


.github/workflows/on-pull-request-run-tests.yml
```diff
@@ -327,7 +330,7 @@ jobs:
   awx-e2e:
     name: AWX E2E
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 20
```
jamestalton, line 333: we do not want to bump the timeout up

.github/workflows/on-pull-request-run-tests.yml
```diff
@@ -3,6 +3,9 @@ name: On Pull Request - Run Tests
 on:
   pull_request:
 
+env:
+  IS_GITHUB_ACTION: true
```
jamestalton, line 7: why would we only check for these variables in github actions?

jamestalton (changes requested):

.github/workflows/on-pull-request-run-tests.yml
nixocio: Idea was to skip that check in GithubActions.

.github/workflows/on-pull-request-run-tests.yml
nixocio: It was just a test to verify that the changes were not the ones to cause test failures, I will update.

jamestalton (approved):
lgtm

PR #823, opened 2023-08-25
[AAH-2496] Add initial work for Remote details page

nixocio:
Add initial work for Remote details page

image


nixocio:
After it was approved, I just rebased this one. Merging.

PR #824, opened 2023-08-25
AWX Dashboard update to use Count Bar designed by UXD

jamestalton:
Screenshot 2023-08-25 at 11 45 22 AM


kialam (changes requested):
We need relevant e2e tests:
1. Number within donut chart equal to total count
2. Clicking on number within donut chart takes you to list view

`useManagedAwxDashboard` will need to be updated to reflect the new card view

framework/PageDashboard/PageDashboardCountBar.tsx
```diff
@@ -1,49 +1,157 @@
...
-                  <Link to={item.to}>
-                    <ArrowRightIcon />
-                  </Link>
-                </FlexItem>
-              </Flex>
-            </FlexItem>
-          ))}
-        </Flex>
+                </Link>
+                {total && 'counts' in item && item.counts ? (
```
kialam, line 47: The last two arguments of this conditional is confusing to parse -  could you explain what this is checking?

kialam:
I see in the screenshots that projects shows a sync failure, however hosts and inventories do not have the same legend to indicate to users that they have (zero) failures. We should consistently show legend keys so that there is no confusion as to what these charts are supposed to represent.

kialam (approved):
I had some questions and concerns, however the code functions and works correctly with the manageView dashboard.

jamestalton:
> I see in the screenshots that projects shows a sync failure, however hosts and inventories do not have the same legend to indicate to users that they have (zero) failures. We should consistently show legend keys so that there is no confusion as to what these charts are supposed to represent.

That is by design. We are only showing the counts of what is there. It makes for a more pleasant dashboard when everything on it is showing green. If we showed 0 failures but still showed the red square in the legend it breaks that up and draws the eye to it when it should not.

framework/PageDashboard/PageDashboardCountBar.tsx
jamestalton: comments added to explain

PR #827, opened 2023-08-28
Add missing required to Organization in create User form

ZitaNemeckova:
Before:
Screenshot 2023-08-28 at 13 03 50

After:
Screenshot 2023-08-28 at 13 04 30


jamestalton:
I believe users can be created without an organization.
I think the UserForm code should handle that better.

ZitaNemeckova:
Yeah it's possible to create an user without org via API. Closing.

PR #830, opened 2023-08-28
[AAP-13112] Change Project Git Hash field on Activation Details page

Dostonbek1:
## Problem

When the activation is created without a project, the git hash doesn't appear in the activation detail view. 

## Solution

There is a [PR](https://github.com/ansible/eda-server/pull/375) up for the eda-server side that creates a new `git_hash` field for Activations, which depends on the rulebook field to populate rather than project field.

Depends on https://github.com/ansible/eda-server/pull/375

Solves: [AAP-13112](https://issues.redhat.com/browse/AAP-13112)


lgalis (approved):

lgalis:
Closed in favor of https://github.com/ansible/ansible-ui/pull/959

PR #833, opened 2023-08-29
Top Content in Page Table

MilanPospisil:
This PR adds option to add custom content between PageToolbar and Table.

After review, example from Collections will be deleted. So eslint error will dissapear too.


frontend/hub/collections/Collections.tsx
```diff
@@ -57,6 +57,13 @@ export function Collections() {
         {...view}
         defaultTableView="list"
         defaultSubtitle={t('Collection')}
+        topContent={
```
MilanPospisil, line 60: This will be removed after review.

jamestalton (approved):
lgtm

PR #838, opened 2023-08-30
Isolate AWX, HUB, and EDA webpack entry points

jamestalton:
Changes the webpack to have a unique entry point for AWX, HUB, and EDA,
instead of a single index.tsx that builds just off the AAP_MODE environment variable.

This is in preparation for the downstream platform work.
This will also be a good pattern as we move analytics out of AWX and also if LS integrates with the repo.

Two follow on PRs are planned:

1. Move the projects up one directory. (i.e. instead of /frontend/eda it would just be /eda)
2. Abstract the navigation and routing for each project so that downstream the platform can use those.


frontend/awx/Awx.tsx
```diff
@@ -1,10 +1,68 @@
...
 import { AwxSidebar } from './AwxSidebar';
-import { WebSocketProvider } from './common/useAwxWebSocket';
 import { AwxConfigProvider } from './common/useAwxConfig';
+import { WebSocketProvider } from './common/useAwxWebSocket';
+
+const container = document.createElement('div');
+container.style.position = 'fixed';
+container.style.width = '100%';
+container.style.height = '100%';
+container.style.overflow = 'hidden';
```
keithjgrant, line 30: Why is this all necessary? Fixed positioning seems odd here.

I would suggest only min-height: 100% (or 100dvh)

jamestalton: This is so that the main container is exactly the size of the window and has no overflow.
That is needed so that subsections of the page can handle the overflow without the whole pages getting a scrollbar.

keithjgrant (approved):

PR #842, opened 2023-08-30
adds / to project and orgs get url

vishsanghishetty:
Calls were getting redirected with 301 errors when fetching the Projects and Organizations details page.
During troubleshooting, James found that there is a `/` missing at the end of useGet custom hook URL


jamestalton (approved):
lgtm

PR #844, opened 2023-08-31
Async filter builder

MilanPospisil:
Builder for the async selector. This allows much less code to build standardized modal with page table selection. User can only pass useFilters, useColumns and view parameters and the builder will construct asyn selector function, like previous useRepositorySelector.

TODO (next PRs):
- also construct useRepoQueryOptions for async dropdown list, if it will be possible to construct this from view parameters - the data returned will be the same.

- add comments to async filters and this builder.


jamestalton (approved):
lgtm

PR #847, opened 2023-08-31
[AAP-15552] Modify PageFormDataEditor to allow upload, download of files

nixocio:
Modify PageFormDataEditor to allow upload, download of files

Add function to render row of actions, update unit tests.

Add react-dropzone as dependency.

image


jamestalton (approved):
lgtm

PR #850, opened 2023-08-31
Add useClipboard hook

nixocio:
Add useClipboard hook to ease interact with clipboard API, and deal with errors.

### DIFF
--- a/framework/hooks/useClipboard.tsx
+++ b/framework/hooks/useClipboard.tsx
@@ -1,0 +1,57 @@
+import { useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { usePageAlertToaster } from '../PageAlertToaster';
+import { AlertProps } from '@patternfly/react-core';
+
+type UseClipboardResult = {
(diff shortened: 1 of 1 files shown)


framework/hooks/useClipboard.tsx
```diff
@@ -0,0 +1,40 @@
...
+    }
+    try {
+      await navigator.clipboard.writeText(text);
+      setCopySuccess(true);
+    } catch (err) {
+      setErrorMessage(t('Failed to copy to clipboard.'));
+    }
+  };
+
+  return { copySuccess, errorMessage, writeToClipboard };
```
nixocio, line 39: @jamestalton we could also return a boolean error to allow to customize message, or not show an alert toaster for instance.

```suggestion
  return { copySuccess, errorMessage, error, writeToClipboard };
```

jamestalton: would we want this to automatically use the toaster and show an error if something failed?

nixocio: We can do that, I will update accordingly.

PR #851, opened 2023-09-01
Collection Detail Version Picker

MilanPospisil:
Adds collection version picker to collection detail.

image

### DIFF
--- a/frontend/hub/collections/CollectionDetails.tsx
+++ b/frontend/hub/collections/CollectionDetails.tsx
@@ -65,6 +65,9 @@ export function CollectionDetails() {
   const { t } = useTranslation();
   const [searchParams, setSearchParams] = useSearchParams();
   const itemActions = useCollectionActions(() => void refresh());
+  const [collection, setCollection] = useState<CollectionVersionSearch | undefined>(undefined);
+
+  let collectionError: JSX.Element | undefined = undefined;
 
   function setVersionParams(version: string) {
     setSearchParams((params) => {
@@ -72,31 +75,27 @@ export function CollectionDetails() {
       return params;
     });
   }
+  // load all collections versions belong to the repository
+  const collectionsResult = useGet<HubItemsResponse<CollectionVersionSearch>>(
+    hubAPI`/v3/plugin/ansible/search/collection-versions/?name=${
+      searchParams.get('name') || ''
+    }&namespace=${searchParams.get('namespace') || ''}&repository_name=${
+      searchParams.get('repository') || ''
+    }&order_by=-pulp_created`
+  );
 
-  function loadVersions() {
-    // load all collections versions belong to the repository
-    const { data, error, isLoading } = useGet<HubItemsResponse<CollectionVersionSearch>>(
-      hubAPI`/v3/plugin/ansible/search/collection-versions/?name=${
-        searchParams.get('name') || ''
-      }&namespace=${searchParams.get('namespace') || ''}&repository_name=${
-        searchParams.get('repository') || ''
-      }&order_by=-pulp_created`
+  if (
+    collectionsResult.error ||
+    (collectionsResult.isLoading == false &&
+      collectionsResult.data &&
+      collectionsResult.data.data.length == 0)
+  ) {
+    collectionError = (
+      <AwxError error={collectionsResult.error || { name: 'not found', message: t('Not Found') }} />
     );
-
-    let result: { error: JSX.Element | null; data: CollectionVersionSearch[] } = {
-      error: null,
-      data: data ? data.data : [],
-    };
-
-    if (error || (isLoading == false && data && data.data.length == 0)) {
-      result.error = <AwxError error={error || { name: 'not found', message: t('Not Found') }} />;
-    }
-
-    return result;
   }
 
-  const versionsResponse = loadVersions();
-  const collections = versionsResponse.data;
+  const collections = collectionsResult.data ? collectionsResult.data.data : [];
 
   // load collection by search params
   const version = searchParams.get('version');
@@ -111,8 +110,6 @@ export function CollectionDetails() {
     versionFilter = '&version=' + version;
   }
 
-  let collection = null;
-
   const { data, refresh, error, isLoading } = useGet<HubItemsResponse<CollectionVersionSearch>>(
     hubAPI`/v3/plugin/ansible/search/collection-versions/?name=${
       searchParams.get('name') || ''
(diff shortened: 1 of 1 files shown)


frontend/hub/collections/CollectionDetails.tsx
```diff
@@ -55,20 +56,42 @@ import { HubItemsResponse } from '../useHubView';
...
-  const { data, refresh } = useGet<HubItemsResponse<CollectionVersionSearch>>(
-    hubAPI`/v3/plugin/ansible/search/collection-versions/?name=${params.name || ''}&namespace=${
-      params.namespace || ''
-    }&repository=${params.repository || ''} }`
+  const [searchParams] = useSearchParams();
+  const [collection, setCollection] = useState<CollectionVersionSearch | undefined>(undefined);
+  const itemActions = useCollectionActions(() => void refresh());
+
+  // load all collections versions belong to the repository
+  const { data, refresh, error, isLoading } = useGet<HubItemsResponse<CollectionVersionSearch>>(
```
nixocio, line 71: 👍

nixocio (approved):

vidyanambiar (changes requested):
We'll need a space between "updated" and the hours:

Also in the case below (collections i created using galaxykit), all the version numbers in the picker are the same and the 2nd, 3rd, and and 4th entries look identical. Is that normal? 



Screenshot 2023-09-01 at 9 10 18 AM

himdel:
Don't forget about the version picker *modal* as well, when there is more than 10 version, the picker should be showing the first 10 and a button to open the modal with pagination.

MilanPospisil:
> 

Ok, than I will finish my changes to Framework first, this will simplify those modals.

frontend/hub/collections/CollectionDetails.tsx
```diff
@@ -55,20 +56,42 @@ import { HubItemsResponse } from '../useHubView';
...
+    hubAPI`/v3/plugin/ansible/search/collection-versions/?name=${
+      searchParams.get('name') || ''
+    }&namespace=${searchParams.get('namespace') || ''}&repository_name=${
+      searchParams.get('repository') || ''
```
himdel, line 75:
```suggestion
       collection?.repository.name || ''
```

maybe?

MilanPospisil: Collection is selected collection. That would not work...

frontend/hub/collections/CollectionDetails.tsx
```diff
@@ -84,7 +107,59 @@ export function CollectionDetails() {
             selectedItem={collection}
           />
         }
+        description={t('Repository: ') + collection?.repository.name}
```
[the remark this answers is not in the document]
MilanPospisil: Do you think repository can be null? Hmm... I should have generate all inner object of API interfaces with nullables, so typescript will always catch those. We cant rely that the API returns what it should return :-)

MilanPospisil:
> We'll need a space between "updated" and the hours:
> 
> Also in the case below (collections i created using galaxykit), all the version numbers in the picker are the same and the 2nd, 3rd, and and 4th entries look identical. Is that normal?
> 
> Screenshot 2023-09-01 at 9 10 18 AM

I cant replicate the problem. Do you have repository in URL? The query does have repository filter and I have multiple versions in multiple repos and I dont see them in one dropdown.

MilanPospisil:
I apologize for mayor change. I forgot about version in URL - sometimes we need to redirect to certain version, rather than select it. Now the selector dropdown changes URL and version is loaded using URL parameters. If URL param version is missing, it will use is_highest instead of version. It can load version directly without any selection if version is in URL.

frontend/hub/collections/CollectionDetails.tsx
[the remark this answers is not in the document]
himdel: Wasn't `null` the behavior for orphaned collection versions? But maybe those get filtered out now, I'm not sure.

frontend/hub/collections/CollectionDetails.tsx
himdel: Why not? The repo of the current collection sounds like the right thing to filter by when searching for other versions of the same collection in the same repo..

(but maybe I'm misunderstanding `searchParams`? I thought it was a query string param, but maybe it's a route param?)

frontend/hub/collections/CollectionDetails.tsx
```diff
@@ -55,20 +56,92 @@ import { HubItemsResponse } from '../useHubView';
...
+  } else {
+    versionFilter = '&version=' + version;
+  }
+
+  const { data, refresh, error, isLoading } = useGet<HubItemsResponse<CollectionVersionSearch>>(
+    hubAPI`/v3/plugin/ansible/search/collection-versions/?name=${
+      searchParams.get('name') || ''
+    }&namespace=${searchParams.get('namespace') || ''}&repository_name=${
+      searchParams.get('repository') || ''
+    }&order_by=-pulp_created` +
```
himdel, line 118: (and here)

vidyanambiar (approved):
lgtm!

frontend/hub/collections/CollectionDetails.tsx
MilanPospisil: Never mind, I added version to search params, so now collection is loaded based on search params, not based on selection (selection changes search params version and collection is loaded then).

frontend/hub/collections/CollectionDetails.tsx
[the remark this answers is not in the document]
MilanPospisil: You are maybe right. Im just saying that we should first change the definition in repository if this is true, so repository is nullable, that will enforce any check automaticaly.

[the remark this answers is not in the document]
himdel: Yeah, I was mostly just pointing out that `foo?.bar.baz` doesn't make sense and should be either `foo.bar.baz` or `foo?.bar?.baz` :)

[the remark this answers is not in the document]
MilanPospisil: Yeah, thats why we should do API results definitions with nullable ?, so typescript will enforce it.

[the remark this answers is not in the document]
MilanPospisil: Lets merge this and I will do it in seperate PR.

frontend/hub/collections/CollectionDetails.tsx
MilanPospisil: I hope this will always return 1 item anyway :)

PR #852, opened 2023-09-01
PageTable padding, PageTabs isBox styling, PageChart layout enhancements

jamestalton:
https://issues.redhat.com/browse/AAP-15732

Recently there was an addition to the table component to support a `topContent` property. This puts that content at the top of the table between the table toolbar and the table itself. The use case is for Analytics to be able to use this to place a chart at the top of a table. Here is an example of what it would look like if we used that for the AWX jobs table. NOTE: The chart showing in jobs is not in this PR, it is just an example.

Screenshot 2023-09-01 at 9 58 51 AM

UXD discussions revolved around the padding that shows up around the tables. The concensus seems to be that we want to be consistent and that since it looks better without the padding when there is a chart, that it makes sense to turn the padding off consistently. The only view that will still have the padding is the cards view.

One of the big considerations here is the details pages. UXD recommends that we remove the padding and that to make the tabs a little clearer we switch to the PatternFly `isBox` style of tabs.

Screenshot 2023-09-01 at 10 02 19 AM

In addition to those above changes, this PR also enhances the stackedArea/line chart component to handle better responsive layout.

To do this we use html components to render things like the axis labels and chart legend. Discussed this approach with the PatternFly team and they think that it solves the responsive layout issues and are going to look into that same kind of changes on the PatternFly side.

Screenshot 2023-09-01 at 9 20 59 AM

- Layout is much more responsive and spacing for the labels and legend is much cleaner.
- Axis labels are now html components.
- Legend is now an html component.
    - Legend now supports counts and links
    - Legend is the same component that is used in the CountCard at the top of those screenshots.
- Chart also auto calculates it's domain range and determines an approximate size for the yAxis tick labels and adjusts the spacing between the yAxis label and the ticks so that they do not overlap.
    - Example: There was one server that had hundreds of jobs and the y-axis tick label was "200" which almost touched the y-axis label. This solves that issue.
    
NOTE: This does change the chart so that it renders to the edges. This is good design in that it allows any place that is including the chart to handle it's own padding. i.e. a chart in a card should use the CardBody for padding, where a chart in a Page should use the PageSection for padding.

NOTE: The chart changes does include a negative 16px margin that overflows its render area so that any labels or chart elements that render are not cut off at the borders.

### DIFF
--- a/framework/PageDashboard/PageChartLegend.tsx
+++ b/framework/PageDashboard/PageChartLegend.tsx
@@ -1,4 +1,5 @@
 import { Link } from 'react-router-dom';
+import styled from 'styled-components';
 
 export function PageChartLegend(props: {
   id: string;
@@ -7,51 +8,25 @@ export function PageChartLegend(props: {
 }) {
   if (props.horizontal) {
     return (
-      <div
-        style={{
-          display: 'flex',
-          flexWrap: 'wrap',
-          flexDirection: 'row',
-          alignItems: 'center',
-          justifyContent: 'center',
-          gap: 20,
-        }}
-      >
+      <PageChartLegendHorizontalStyle>
         {props.legend.map((item, index) => {
           if (item.count === 0) return <></>;
           return (
-            <div
-              key={index}
-              style={{
-                display: 'flex',
-                flexDirection: 'row',
-                alignItems: 'center',
-                justifyContent: 'center',
-                gap: 6,
-              }}
-            >
+            <PageChartLegendHorizontalItemStyle key={index}>
               <PageChartLegendColor
                 id={`${props.id}-${item.label.toLowerCase().replace(/ /g, '-')}-count`}
                 color={item.color}
               />
               <PageChartLegendCount count={item.count} />
               <PageChartLegendLabel label={item.label} link={item.link} />
-            </div>
+            </PageChartLegendHorizontalItemStyle>
           );
         })}
-      </div>
+      </PageChartLegendHorizontalStyle>
     );
   }
   return (
-    <div
-      style={{
-        display: 'grid',
-        gridTemplateColumns: 'auto auto auto',
-        rowGap: 6,
-        columnGap: 6,
-        alignItems: 'center',
-      }}
-    >
+    <PageChartLegendVerticalStyle>
       {props.legend.map((item, index) => {
         if (item.count === 0) return <></>;
         return (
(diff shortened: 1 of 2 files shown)


frontend/common/RoutedTabs.tsx
```diff
@@ -153,9 +147,17 @@ export function RoutedTabs(props: {
...
+            className="border-left border-bottom"
+            style={{
+              backgroundColor: 'var(--pf-global--BackgroundColor--200)',
+              alignSelf: 'normal',
+            }}
+          />
           {props.postComponents && (
             <>
-              <Divider orientation={{ default: 'vertical' }} />
+              {/* <Divider orientation={{ default: 'vertical' }} /> */}
```
vidyanambiar, line 160: Can we get rid of the commented line?

frontend/awx/views/jobs/Jobs.tsx
```diff
@@ -110,6 +95,13 @@ export default function Jobs() {
         emptyStateDescription={t('Please run a job to populate this list.')}
         {...view}
         defaultSubtitle={t('Job')}
+        // topContent={
```
vidyanambiar, line 98: Let's get rid of this if it isn't needed.

frontend/common/RoutedTabs.tsx
```diff
@@ -102,31 +101,26 @@ export function RoutedTabs(props: {
...
   return (
     <>
-      <PageSection type={PageSectionTypes.tabs} className="border-bottom">
+      <PageSection type={PageSectionTypes.tabs}>
         <Flex spaceItems={{ default: 'spaceItemsNone' }}>
           {props.preComponents && (
             <>
               <FlexItem>{props.preComponents}</FlexItem>
-              <Divider orientation={{ default: 'vertical' }} component="div" />
+              {/* <Divider orientation={{ default: 'vertical' }} component="div" /> */}
```
vidyanambiar, line 109: can we get rid of this?

framework/PageTable/PageTable.tsx
```diff
@@ -310,16 +309,31 @@ export function PageTable<T extends object>(props: PageTableProps<T>) {
...
-      {props.topContent && <PageSection>{props.topContent}</PageSection>}
+      {topContent}
       <PageTableView {...props} tableColumns={managedColumns} />
     </>
   );
 
-  if (props.scrollOutsideTable) {
-    tableContent = <Scrollable>{tableContent}</Scrollable>;
-  }
+  // if (props.scrollOutsideTable) {
```
vidyanambiar, line 334: Let's remove these lines if they're not needed.

framework/PageDashboard/PageChartLegend.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+export function PageChartLegend(props: {
+  id: string;
+  legend: { label: string; count?: number; color: string; link?: string }[];
+  horizontal?: boolean;
+}) {
+  if (props.horizontal) {
+    return (
+      <div
+        style={{
```
vidyanambiar, line 11: It would be good to utilize `styled-components` instead of the inline styling.

vidyanambiar (commented):
This looks great! 👍 

I added a couple minor comments.

frontend/awx/views/jobs/Jobs.tsx
jamestalton: At the moment it is example code that UXD would like to see get enabled long term.

framework/PageDashboard/PageChartLegend.tsx
jamestalton: Yes. I personally find inline styles to be easier to follow than having to follow a link to a style that is elsewhere in the code. Performance impacts in this case are small.

vidyanambiar (approved):

PR #858, opened 2023-09-05
[AAH-2492] Update usePulpView to allow defaultSort, and defaultSortDirection

nixocio:
Update how remotes are being sorted when fetched.

Also, make the URL to fetch the remotes dedicated to ansible/collections/

### DIFF
--- a/framework/useView.tsx
+++ b/framework/useView.tsx
@@ -37,7 +37,7 @@ export interface IView {
  * - filters
  *
  * It also manages the query string parameters for the table.
- * It will load the dafulat values from the query string if they exist.
+ * It will load the default values from the query string if they exist.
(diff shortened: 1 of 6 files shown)


himdel:
Why not make `defaultSort` a `usePulpView` param? That would ensure the list view can pick the default sort, instead of relying on a potentially shared list of headers? (How do you ensure there's only one entry with defaultSort when creating the list dynamically?)

nixocio:
@himdel defaultSort is being added here to usePulpView.

https://github.com/ansible/ansible-ui/blob/93056179832c076de6004f814272b584bca81fe8/frontend/hub/usePulpView.tsx#L37

I will add the same to useHubView.

PR #816, opened 2023-08-24
Shaiahwren approve collections

ShaiahWren:
This pr add functionality to the approval page for approving and rejecting collections. 

The approval "state" on the UI in columns updates about 5 seconds after modal close. Any suggestions for remedying this are welcome:)

PR #860, opened 2023-09-05
Shaiahwren sign collections (#816)

ShaiahWren:
This pr continues with the collection signing and approval feature on ansible-ui.


nixocio:
hey @ShaiahWren there is an error when running `npm run tsc`.

```
frontend/hub/approvals/hooks/useApprovalActions.tsx:30:9 - error TS2322: Type '(collection: CollectionVersionSearch) => boolean' is not assignable to type 'string | ((item: CollectionVersionSearch) => string | undefined) | undefined'.
  Type '(collection: CollectionVersionSearch) => boolean' is not assignable to type '(item: CollectionVersionSearch) => string | undefined'.
    Type 'boolean' is not assignable to type 'string'.

30         isDisabled: (collection) =>
           ~~~~~~~~~~
```

frontend/hub/api/request.ts
```diff
@@ -57,6 +59,41 @@ async function hubRequestCommon<T extends object | TaskResponse>(
...
+  signal?: AbortSignal
+): Promise<unknown> {
+  const body = new FormData();
+  body.append('file', file);
+  {
+    repository && body.append('repository', repository);
+  }
+  {
+    signed_collection && body.append('signed_collection', signed_collection);
+  }
```
himdel, line 76: this looks like the same code as in crud/Data .. should we really be adding this in both places?
(Looks pretty hub specific)

frontend/common/crud/Data.tsx
```diff
@@ -35,10 +35,18 @@ export async function postRequest<ResponseBody, RequestBody = unknown>(
...
   signal?: AbortSignal
 ): Promise<unknown> {
   const body = new FormData();
   body.append('file', file);
+  {
+    repository && body.append('repository', repository);
+  }
+  {
+    signed_collection && body.append('signed_collection', signed_collection);
+  }
```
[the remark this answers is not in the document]
himdel: (but also https://github.com/ansible/ansible-ui/pull/860/files#r1345924653)

frontend/hub/approvals/hooks/useApprovalsColumns.tsx
```diff
@@ -41,14 +40,18 @@ export function useApprovalsColumns(_options?: { disableSort?: boolean; disableL
...
+                    ? t`Needs signature and review`
+                    : t`Needs review`
+                }
                 color={PFColorE.Warning}
               />
             );
           }
 
-          if (approval.repository?.pulp_labels?.pipeline == 'approved') {
-            if (approval.is_signed && display_signatures) {
```
himdel, line 51: probably not meant to remove this?

if display_signmatures is false, this needs to show only Approved

frontend/hub/api/request.ts
ShaiahWren: you're right, it was in crud/Data, used in uploadCollection.tsx I believe... but since it's used nowhere else, should we move it permanently down to requests? I moved it there initially because it was hub-specific

himdel: Any code that appends repository and signed collections is not something the other projects can use.
It's not really something hub could use for other file uploads either.

So.. this is not really a `hubPostRequestFile` or a `postRequestFile`, this is more like `hubUploadCollectionSignature`.

But then again, if we could use actual functions to call apis, instead of all the would-be generic code, things would be much simpler everywhere ... so...  *shrug*, as long as you make sure other hub file uploads still work, :+1:

frontend/common/crud/Data.tsx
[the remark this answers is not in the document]
ShaiahWren: So... why don't I remove this from crud/Data and just keep request.ts... sound good? :)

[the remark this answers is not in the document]
himdel: Sounds good :) it makes more sense in the hub version :+1:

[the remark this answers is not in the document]
ShaiahWren: FYI (possibly unrelated?) the collection upload is quite glitchy and there may need to be extra error handling for files its doesn't like. I can try to show you how to reproduce the error if you'd like. :)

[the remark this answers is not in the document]
himdel: Interesting, is it the case of some tarballs from galaxy failing to upload here? And do those fail in the old ui as well?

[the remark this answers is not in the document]
ShaiahWren: Things are looking better now, just rebased.. maybe there was some work done on it. I'll look out for the same errors.

Looks like this took care of your comments, now for the tsc errors :)
Do you think we can merge this  as if (after tsc passes) and then do a separate pr to refactor the approval/signing service stuff?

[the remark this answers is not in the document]
ShaiahWren: Also - do you know why tsc is unhappy in `collectionSignatureUpload` on line 67?

framework/PageActions/PageAction.tsx
```diff
@@ -51,12 +51,12 @@ interface IPageActionNoneCommon extends IPageActionCommon {
 interface IPageActionSingleCommon<T extends object> extends IPageActionCommon {
   selection: PageActionSelection.Single;
   isHidden?: (item: T) => boolean;
-  isDisabled?: string | ((item: T) => string | undefined);
+  isDisabled?: string | ((item: T) => string | undefined | boolean);
```
himdel, line 54: these are causing the first 2 tsc failures... I don't think you want to add boolean here, you need to fix your isDisabled to return string or undefined

I think the convention is that isDisabled returns a string reason for why it's disabled, or undefined when not disabled; not just a boolean true/false

frontend/hub/api/request.ts
```diff
@@ -57,6 +59,41 @@ async function hubRequestCommon<T extends object | TaskResponse>(
   };
 }
 
+export async function hubPostRequestFile(
+  url: string,
+  file: Blob,
+  repository?: string,
+  signed_collection?: string,
+  signal?: AbortSignal
+): Promise<unknown> {
```
himdel, line 68: `Promis<unknown>` .. given the tsc error I'm *guessing* for async functions, you'd just want unknown, not Promise<unknown> .. unless that blows up somewhere else :) 

Or return an actual value and use that for the type? Since we're waiting for a task, maybe returning the TaskResponse would make sense

frontend/hub/collections/CollectionSignatureUpload.tsx
```diff
@@ -0,0 +1,94 @@
...
+    collection = data.data[0];
+  }
+  const collectionID = collection?.collection_version?.pulp_href;
+
+  async function submitData(data: UploadData) {
+    const repoRes = (await getRequest(
+      pulpAPI`/repositories/ansible/ansible/?pulp_label_select=pipeline=staging`
+    )) as PulpItemsResponse<Repository>;
+
+    const stagingRepo = repoRes.results[0].pulp_href;
```
himdel, line 68: sounds like the Repository type is missing a `pulp_href: string`

framework/PageActions/PageAction.tsx
ShaiahWren: Thank you, so helpful... do you mean that it's literally a plain-text description? I'll look at some other examples:)

ShaiahWren: Probably something like this: `            : t`You do not have rights to this operation`,
`

himdel: Yup, something like that, wrapped in a `t(...)` for l10n.

frontend/hub/api/request.ts
```diff
@@ -1,6 +1,8 @@
 import { createRequestError } from '../../common/crud/RequestError';
 import { getCookie } from '../../common/crud/cookie';
 import { TaskResponse } from '../tasks/Task';
+import { parseTaskResponse } from './utils';
+// import { TaskResponse } from '../api-schemas/generated/TaskResponse';
```
himdel, line 5:
```suggestion
```

?

himdel:
Looks like the linter stuff is fixed :+1:, but there are some changes in package-lock.json, are those intentional?

nixocio (approved):

nixocio:
> Looks like the linter stuff is fixed 👍, but there are some changes in package-lock.json, are those intentional?

@ShaiahWren overall looks good. Please, see @himdel comment above.

himdel (approved):
:+1: 

(the package-lock change is whitespace-only now, so it shouldn't break anything if there's a problem removing it completely :))

PR #863, opened 2023-09-06
Add requirements file to add, edit and details page for remotes

nixocio:
Add YAML requirements file to add, edit and details page for remotes.

image

image

See: https://issues.redhat.com/browse/AAH-2496
Also: https://issues.redhat.com/browse/AAH-2493

### DIFF
--- a/frontend/hub/remotes/RemoteDetails.tsx
+++ b/frontend/hub/remotes/RemoteDetails.tsx
@@ -2,6 +2,7 @@ import { Label, LabelGroup } from '@patternfly/react-core';
 import { useTranslation } from 'react-i18next';
 import { useParams } from 'react-router-dom';
 import { PageDetail, PageDetails, PageHeader, PageLayout, useGetPageUrl } from '../../../framework';
+import { PageDetailCodeEditor } from '../../../framework/PageDetails/PageDetailCodeEditor';
 import { LoadingPage } from '../../../framework/components/LoadingPage';
 import { AwxError } from '../../awx/common/AwxError';
 import { useGet } from '../../common/crud/useGet';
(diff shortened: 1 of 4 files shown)


frontend/hub/remotes/RemoteForm.tsx
```diff
@@ -52,14 +61,15 @@ export function CreateRemote() {
         submitText={t('Create remote')}
         onSubmit={onSubmit}
         onCancel={() => navigate(-1)}
-        defaultValue={{ name: '', url: '' }}
+        defaultValue={{ name: '', url: '', requirements_file: yamlRequirementsTemplate }}
```
himdel, line 64: this would prefill the requirements file with the sample, right?

Note that the requirements file is an optional value, mostly you'll want it emtpy, except when syncing from huge remote repositories.

frontend/hub/remotes/RemoteDetails.tsx
```diff
@@ -103,6 +104,13 @@ export function RemoteDetails() {
             )}
           </PageDetail>
         </PageDetails>
+        <PageDetails numberOfColumns="single">
+          <PageDetailCodeEditor
+            label={t('YAML requirements')}
+            value={remote?.requirements_file ?? '---'}
+            showCopyToClipboard={true}
```
himdel, line 111: should this be disabled when requirements_file is falsy? (copy to clipboard or the whole code editor?)

nixocio: @himdel +1

frontend/hub/remotes/RemoteForm.tsx
nixocio: I changed the logic, if the user open the yaml editor the sample code is there as a helper, but it does not get sent to the backend if the user does not change that value.

himdel (approved):

PR #867, opened 2023-09-07
[AAP-15860] Move error alert to the top of PageForm component

nixocio:
Move error alert to the top of PageForm component

Also, update padding on the Alert component to align with the remaining page structure.

image

### DIFF
--- a/framework/PageForm/PageForm.tsx
+++ b/framework/PageForm/PageForm.tsx
@@ -20,12 +20,12 @@ import {
   UseFormReturn,
   useFormState,
 } from 'react-hook-form';
+import { useTranslation } from 'react-i18next';
 import { RequestError } from '../../frontend/common/crud/RequestError';
 import { useBreakpoint } from '../components/useBreakPoint';
 import { PageBody } from '../PageBody';
 import { SettingsContext } from '../Settings';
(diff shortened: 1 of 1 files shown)


AlexSCorey (commented):
If we are going to move this error I think it should be sticky and always visible, even when the user scrolls down to the bottom on a long, single column form.  I would get very confused if I scrolled to the bottom of a form to submit it, but wasn't alerted to the error because it was hidden at the top.

nixocio:
I added this change @AlexSCorey 

image

framework/PageForm/PageForm.tsx
```diff
@@ -120,6 +120,22 @@ export function PageForm<T extends object>(props: {
...
         }}
       >
+        {error && (
+          <Alert
+            variant="danger"
+            title={error.message ?? ''}
+            isInline
+            style={{
+              paddingLeft: isMd && props.onCancel ? 24 : undefined,
+              position: 'sticky',
```
jamestalton, line 130: i'd like to avoid sticky.
I think the alert should be outside the scrollable area of the form

jamestalton (changes requested):

PR #870, opened 2023-09-07
Remote ky http client

nixocio:
Remote ky http client, adjust HTTP clients to use fetch API, and replace HTTPError with RequestError - custom error.

Update parse of errors for HUB http clients


frontend/common/crud/requestCommon.ts
```diff
@@ -5,7 +5,7 @@ export function requestCommon(options: {
   method: string;
   body?: unknown;
   headers?: HeadersInit;
-  signal: AbortSignal;
+  signal?: AbortSignal;
```
jamestalton, line 8: there is never a case where we would not want an abort signal passed in here
making this optional will allow devs to have the bad practice of not supporting an abort signal

jamestalton (changes requested):

frontend/common/crud/requestCommon.ts
nixocio: That was a test leftover. Updated.

MilanPospisil (approved):
LGTM :+1: 
But tests are failing, are they relevant? Does not seems they are.

jamestalton (approved):
lgtm

PR #872, opened 2023-09-07
[AAP-15667] Replace activation 'Deleted' status with a label

lgalis:
- Replace activation 'Deleted' status with a label
- Replace 'Rule name' with "name in the Rule audit details page.
- 

Screenshot from 2023-09-07 17-12-56
Screenshot from 2023-09-08 10-48-39


vishsanghishetty (approved):

tiyiprh (changes requested):
Can the Details page also use the same "Delete" label as on the table?

lgalis:
@tiyiprh - added the label to the details page - screenshot updated

PR #878, opened 2023-09-08
Turn off autocomplete on all forms + inputs

mabashian:
autocomplete has caused us nothing but trouble in the past.  This PR deliberately turns it off on all forms and text inputs.


keithjgrant (approved):
This makes sense to me

vishsanghishetty (approved):

PR #883, opened 2023-09-12
Cypress upgraded to v13.1.0

vidyanambiar:
Version 13 of Cypress contains a new feature called 'Test Replay' which promises easy debugging of CI test failures in Cypress Cloud.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -114,7 +114,7 @@
         "@types/js-yaml": "4.0.5",
         "@typescript-eslint/eslint-plugin": "6.6.0",
         "@typescript-eslint/parser": "6.6.0",
-        "cypress": "12.17.4",
+        "cypress": "13.1.0",
         "cypress-react-selector": "3.0.0",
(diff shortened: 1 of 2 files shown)


vidyanambiar:
CI checks fail on `npm ci` due to an unresolved peer dependency. See thread: https://redhat-internal.slack.com/archives/C027K1897U4/p1694536513212149

keithjgrant:
> Any recommendations?

I'd prefer to wait for the release rather than get in the habit of using --force or --legacy-peer-deps. I don't want to get stuck in a position where we could end up relying on those flags indefinitely if some dependency doesn't update

keithjgrant:
oh, it looks like the released that change with v2.2.5!

vidyanambiar:
@keithjgrant Thanks! I've updated`@4tw/cypress-drag-drop` to 2.2.5 👍

nixocio (approved):
👍

lgalis (approved):

PR #893, opened 2023-09-13
Subscription chart padding fix

skochay330:
Fix for the subscription usage chart padding, following the changes in https://github.com/ansible/ansible-ui/pull/852


jamestalton (approved):
lgtm

PR #897, opened 2023-09-14
Fix autogenerated interfaces, expand some of them and unifiy CollectionVersionSearch

MilanPospisil:
Autogenerate interfaces again - now without undefined types, so we can easily expand existing interfaces in new files using the same interface name - in 'api-schemas/expanded' folder.

Interfaces also have all inner structures defined as nullables ?, that will enforce programmer to always access them with ? in the path, which is safer, because we cant always rely that backend will send us those structures. Without the ? in paths, any null will fail the loading of whole page.

CollectionVersionSearch is now using the autogenerated interface with expansions - some fields are added. Also makes sure there are no duplicate definitions.


himdel:
Summing up discussion...

```
class Foo {
  foo: string;
  bar: unknown;
}

class Bar extends Foo {
  bar: { quux: number };
}
```

works just fine without `Omit` .. but apparently only outside strict mode :)

So .. I still see value in having the classes contain all the fields, including unknown ones .. but not if it makes development harder :).

---

This also forces a lot of `|| ''` in places where they would not be necessary otherwise .. but it seems in line with the rest of the codebase ¯\\\_(ツ)\_/¯.

The one thing I'm a bit worried about, places like https://github.com/ansible/ansible-ui/pull/897/files#diff-ad607758fff05314b5b34b0e8212206795dc38cc26caa8c08b76f52616e75ba3R59:

```js
  await hubAPIPost(
      pulpAPI`/repositories/ansible/ansible/${
        parsePulpIDFromURL(collection.repository?.pulp_href || '') || ''
      }/move_collection_version/`,
 ...
```

Whereas ``pulpAPI`/repositories/ansible/ansible/${parsePulpIDFromURL(collection.repository.pulp_href)}/move_collection_version/` `` would blow up and fall into the usual error handling code when repository is missing, now it will go using wrong URLs. Probably not critical here, but something to think about :). (Perhaps more relevant when a `?filter={foo}` can be empty, since filter= is the same as no filter. )

(But that part looks like it's out of scope here, approving.)

himdel (approved):

PR #902, opened 2023-09-14
disable https in dev

keithjgrant:
this is only really needed for oath testing in dev, so we can leave it commented out


jamestalton (approved):
lgtm

PR #903, opened 2023-09-14
Update logout endpoint to be per project on Masthead

nixocio:
Update logout endpoint to be per project on Masthead

### DIFF
--- a/frontend/common/Masthead.tsx
+++ b/frontend/common/Masthead.tsx
@@ -50,6 +50,7 @@ import { API_PREFIX } from '../eda/constants';
 import { useAnsibleAboutModal } from './AboutModal';
 import { postRequest } from './crud/Data';
 import { useActiveUser } from './useActiveUser';
+import { hubAPI } from '../hub/api/utils';
 
 const MastheadBrandDiv = styled.div`
   display: flex;
@@ -78,6 +79,10 @@ const ToolbarSpan = styled.span`
 
 function isEdaServer(): boolean {
   return process.env.UI_MODE === 'EDA';
+}
+
+function isHubServer(): boolean {
+  return process.env.UI_MODE === 'HUB';
 }
 
 export function AnsibleMasthead(props: { hideLogin?: boolean }) {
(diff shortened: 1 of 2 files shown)


nixocio:
I talked with @jamestalton and down the road, we may abstract the Masthead component to receive those project specific details as props or another solution, right now adding this one as to allow log out using Hub, and to keep traceability when changing this later on.

frontend/common/Masthead.tsx
```diff
@@ -74,6 +75,10 @@ function isEdaServer(): boolean {
   return process.env.UI_MODE === 'EDA';
 }
 
+function isHubServer(): boolean {
+  return process.env.UI_MODE === 'HUB';
```
keithjgrant, line 79: I'd like to move away from accessing the env variables in the components. I'd suggest passing in the api URLs via a prop. Each app can pass this in where the Masthead is rendered in AwxMain, HubMain, etc

keithjgrant (changes requested):

frontend/common/Masthead.tsx
jamestalton: Im working on abstracting the masthead and making it composable.
That way we can have an AwxMasthead that contains a menu that has a logout that is AWX specific. etc...

frontend/hub/api/utils.tsx
```diff
@@ -8,6 +8,11 @@ import {
   putHubRequest,
 } from './request';
 
+export function getBaseAPIPath() {
```
jamestalton, line 11: Why do we need getBaseAPIPath when there is hubAPI?

jamestalton (changes requested):

frontend/common/Masthead.tsx
nixocio: @keithjgrant Thanks for your review, as James mentioned there is a long term goal to refactor the masthead to achieve what you described.

frontend/hub/api/utils.tsx
nixocio: Updated.

jamestalton (approved):
lgtm

keithjgrant (approved):

PR #909, opened 2023-09-15
Update how remotes are fetched after deleted on details page

nixocio:
Update how remotes are fetched after deleted on details page


himdel (approved):
Makes sense :)

We'll probably need similar logic (navigate to list instead of reload), for all deletions from the detail screen, and for all create actions.

jamestalton (approved):
lgtm

PR #914, opened 2023-09-18
[Analytics]: Fix error handling

ZitaNemeckova:
How to easily reproduce:
- write a typo in a request call

Before:
Screenshot 2023-09-18 at 11 52 47

After:
Screenshot 2023-09-18 at 11 51 58


skochay330:
I get the error below when I try to access the reports page. This error also exists on the main branch, so it is not a result of these changes. 

Screenshot 2023-09-18 at 4 08 08 PM

PR #917, opened 2023-09-18
Legend enhancements

skochay330:
This PR enhances the `PageChartLegend` component to allow for more flexibility and accommodate a wider range of use cases. 

- Added prop `allowZero`: if true, will show all legend items at all times, even if item count is zero
- Added prop `showLegendCount`: if false, will remove the count in each legend item

Before:
Screenshot 2023-09-18 at 2 44 21 PM

After: 
Screenshot 2023-09-18 at 2 43 44 PM


jamestalton (approved):
lgtm

PR #918, opened 2023-09-18
Adds variable fields to the inventory form and inventory details view

AlexSCorey:
Adds the variable fields to the inventory form, and to the inventory details view.  There are no forms yet for constructed inventory and smart inventory.

### DIFF
--- a/frontend/awx/resources/inventories/InventoryForm.tsx
+++ b/frontend/awx/resources/inventories/InventoryForm.tsx
@@ -1,7 +1,13 @@
 import { FieldValues } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
 import { useNavigate, useParams } from 'react-router-dom';
-import { PageFormCheckbox, PageHeader, PageLayout, useGetPageUrl } from '../../../../framework';
+import {
+  PageFormCheckbox,
+  PageFormDataEditor,
+  PageHeader,
+  PageLayout,
+  useGetPageUrl,
+} from '../../../../framework';
 import { PageFormGroup } from '../../../../framework/PageForm/Inputs/PageFormGroup';
 import { PageFormTextInput } from '../../../../framework/PageForm/Inputs/PageFormTextInput';
 import { PageFormSubmitHandler } from '../../../../framework/PageForm/PageForm';
@@ -192,6 +198,11 @@ function InventoryInputs() {
           `Optional labels that describe this inventory, such as 'dev' or 'test'. Labels can be used to group and filter inventories and completed jobs.`
         )}
         name="inventory.summary_fields.labels.results"
+      />
+      <PageFormDataEditor
+        toggleLanguages={['yaml', 'json']}
+        name="inventory.variables"
+        label={t('Variables')}
       />
       <PageFormGroup
         label={t('Options')}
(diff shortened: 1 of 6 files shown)


frontend/awx/resources/inventories/InventoryForm.tsx
```diff
@@ -168,20 +185,19 @@ function InventoryInputs() {
   return (
     <>
       <PageFormTextInput
-        name="inventory.name"
+        data-cy="inventory-name"
```
vidyanambiar, line 188: This data-cy is overwritten by the data-cy defined in the [PageFormTextInput](https://github.com/ansible/ansible-ui/blob/main/framework/PageForm/Inputs/PageFormTextInput.tsx#L264) component. This seems to be why the E2E test is failing while looking the element up by "inventory-name".

frontend/awx/resources/inventories/InventoryForm.tsx
```diff
@@ -168,20 +185,19 @@ function InventoryInputs() {
...
+        name="name"
         label={t('Name')}
         placeholder={t('Enter name')}
         isRequired
       />
       <PageFormTextInput
+        data-cy="inventory-description"
```
vidyanambiar, line 195: same comment as above. This will get overwritten by `data-cy="description"`.

vidyanambiar (commented):
Works well! 👍

frontend/awx/resources/inventories/InventoryForm.tsx
```diff
@@ -1,65 +1,71 @@
...
+    });
 
     // Update new inventory with selected instance groups
-    await submitInstanceGroups(newInventory, instanceGroups ?? [], []);
+    if (instanceGroups?.length > 0)
+      await submitInstanceGroups(newInventory, instanceGroups ?? [], []);
 
     // Update new inventory with selected labels
-    await submitLabels(newInventory, inventory.summary_fields.labels.results);
+    if (labels.length > 0) await submitLabels(newInventory, labels);
```
vidyanambiar, line 64: Re: the E2E failure on editing the inventory: I suspect this has introduced an issues in the PATCH request when we attempt to update labels. The request needs to be patched with `summary_fields.labels`.

vidyanambiar (commented):
When I attempt to save incorrect YAML, a validation error comes up as expected. However, the Cancel button on the form appears to be blocked and I can't seem to exit the form.

frontend/awx/resources/inventories/InventoryForm.tsx
```diff
@@ -86,45 +100,36 @@ export function CreateInventory() {
...
 
     // Update inventory with selected labels
-    await submitLabels(updatedInventory, editedInventory.summary_fields.labels.results);
-    navigate(
-      RouteObj.InventoryDetails.replace(':inventory_type', 'inventory').replace(
-        ':id',
-        (updatedInventory.id ?? 0).toString()
-      )
-    );
+    await submitLabels(updatedInventory, inventory?.summary_fields.labels.results || []);
```
vidyanambiar, line 129: _Re: the E2E failure on editing the inventory:_ I suspect this has introduced an issue in the PATCH request when we attempt to update labels. The request needs to be patched with summary_fields.labels. But `updatedInventory` does not record changes to summary_fields since it is not part of `InventoryFields`.

vidyanambiar (approved):
lgtm 👍

PR #919, opened 2023-09-18
fix getPageUrl to use encodeURIComponent

jamestalton:


nixocio (approved):
I got this code locally, and I was able to edit

image

PR #920, opened 2023-09-19
Analytics report graph builder

MilanPospisil:


frontend/awx/analytics/useAnalyticsReportBuilderView.ts
```diff
@@ -0,0 +1,246 @@
...
+      pageItems: data?.meta.legend,
+      error,
+      ...view,
+      ...selection,
+      originalData: data as undefined,
+    };
+  }, [error, selection, view, data]);
+}
+
+export function getAwxError(err: unknown) {
```
ZitaNemeckova, line 199: In the follow-up PR we need to introduced our own error handling because backend is sending back info on what is set wrongly and we need to show specific error message with hint how to fix it.

ZitaNemeckova (approved):
There are some known issues that will be addressed in follow-up PR.

PR #921, opened 2023-09-19
Streamline setup environment variables

jamestalton:


nixocio (approved):
We will need to notify others once this one gets merged to update their variables.

PR #923, opened 2023-09-19
Make page masthead more composable for downstream

jamestalton:


frontend/awx/AwxMasthead.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <ToolbarItem>
+          <PageMastheadDropdown
+            id="help-menu"
+            icon={<QuestionCircleIcon />}
+            items={[
+              {
+                id: 'documentation',
+                icon: <ExternalLinkAltIcon />,
+                label: t('Documentation'),
+                onClick: () =>
```
keithjgrant, line 63: This should support an href and a standard link instead of JS redirect

keithjgrant (changes requested):

keithjgrant (approved):

PR #934, opened 2023-09-21
Turn login autocomplete on in dev

mabashian:
We'll keep autocomplete off by default for prod but for dev we'll turn it back on.  Here was the original PR: https://github.com/ansible/ansible-ui/pull/878


jamestalton (approved):
lgtm

PR #937, opened 2023-09-21
Add page wizard form

marshmalien:
**PageWizard Component**
Custom page wizard component using PatternFly styles 

**Props**
* steps (required): An array of PageWizardStep objects that define the individual steps or pages of the wizard.

* defaultValue (optional): Default values for the wizard data. These values will be passed to each step of the wizard. If not provided, an empty object is used as the default.

* onCancel (optional): A callback function that will be executed when the user cancels or navigates away from the wizard. By default, it will navigate back in the router history stack.

* onSubmit (required): A callback function that will be executed when the user completes the wizard and submits the data. 

```
const steps = [
  {
    id: 'step1',
    title: 'Step 1',
    content: <Step1Content />,
  },
  {
    id: 'step2',
    title: 'Step 2',
    content: <Step2Content />,
  },
];

<PageWizard
  steps={steps}
  defaultValue={{}} 
  onCancel={onCancel}
  onSubmit={onSubmit}
/>
```


jamestalton (approved):
lgtm

PR #943, opened 2023-09-25
[AAP-15916] Update error handling on PageForm

nixocio:
Create genericErrorAdapter as default error handling on PageForm.

Create awxErrorAdapater to deal with AWX error edge cases.

Create hubErrorAdapter to deal with Hub error edge cases.

Refactor RequestError and add unit tests.

Create a wrapper around PageForm default errorAdapter to be awxErrorAdapter.
This one should be used as part of the AWX project.

Create a wrapper around PageForm default errorAdapter to be hubErrorAdapter.
This one should be used as part of the Hub project.

Remove getAwxError in favor or error handling inside PageForm - or its generic version.

Modify CI to run requestError tests.

image


framework/PageForm/PageForm.tsx
```diff
@@ -215,11 +217,17 @@ export function PageFormGrid(props: {
 
 export function PageFormSubmitButton(props: { children: ReactNode; style?: CSSProperties }) {
   const { isSubmitting, errors } = useFormState();
+  const { clearErrors } = useForm();
   const { t } = useTranslation();
   const hasErrors = errors && Object.keys(errors).length > 0;
   return (
     <Tooltip content={t('Please fix errors')} trigger={hasErrors ? undefined : 'manual'}>
       <Button
+        onClick={() => {
```
nixocio, line 226: It was decided to clean Errors in case we have errors that not specific per field. Thus when the form is submitted the second time the error will be displayed again if remains the same.

nixocio:
I found one more edge case on the AWX adapter, we will have to deal with objects like
 ```json
{
	"__all__": [
		"Inventory with this Name and Organization already exists."
	]
}
```

**Review by AlexSCorey — COMMENTED** [review]
I'm not quite done looking at this, but looks great so far.

File: cypress.common.config.ts [hunk 2]
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
+import { defineConfig } from 'cypress';
+import { baseConfig } from './cypress.base.config';
+
+
+baseConfig.component!.specPattern = 'frontend/common/**/*.cy.{js,jsx,ts,tsx}';
```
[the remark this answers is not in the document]
nixocio: Right now we do not have, I just followed the standard from the remaining files.

framework/PageForm/genericErrorAdapter.cy.ts
```diff
@@ -0,0 +1,53 @@
...
+  it('should return generic error when error is an instance of Error', () => {
+    const error = new Error('Test Error');
+    const result = genericErrorAdapter(error);
+    expect(result).to.deep.equal({
+      genericErrors: [{ name: '__nonfield', message: 'Test Error' }],
+      fieldErrors: [],
+    });
+  });
+
+  it('should return empty errors when error is not a known error', () => {
```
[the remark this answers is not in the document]
nixocio: I will test that one a bit more, but an unknown error should be deal as a form level error.

framework/PageForm/PageForm.tsx
```diff
@@ -133,12 +135,12 @@ export function PageForm<T extends object>(props: {
         {error && (
           <Alert
             variant="danger"
-            title={error.message ?? ''}
+            title={t('Error')}
```
himdel, line 138: is there no support for message title and message description now?

nixocio: @himdel I can add a conditional to add this one, but since error now is a string, or list of string messages, I refactored that one.

himdel:
So.. the interface between the backend-specific parts and the shared parts is an object like...

```js
{
  genericErrors: [
    { name: '_nonFieldError', message: 'Test Error' },
  ],
  fieldErrors: [
    { name: 'name', message: 'Invalid name' },
  ],
}
```

The `fieldErrors` part looks good :+1:  (maybe `{ [fieldName]: [messages] }` would be easier to deal with than `[ {fieldName, message} ]` but sounds like a detail).

But .. the `genericErrors` part .. this bit should use the same kind of logic as the generic error handling elsewhere,
I know it's slightly out of scope for this PR, but:

* `name: '_nonFieldError',` - since this is not associated with any fields, it shouldn't pretend to have the field info .. IMO this just comes from the galaxy API hack where we needed to fit generic errors into an existing `{field:string}` format.
* `message` - most generic errors have either a single string, or a title/description pair .. we should support both?

himdel:
Actually, one more bit .. there were some parts of the UI where we needed the error message to be jsx, not just a string.
(Think `Successfully saved <b>Whatever</b>`)

So .. is the errorAdapter the right place to produce that jsx (it's probably the last place that sees the raw data)?
(I'm not seeing the errorAdapter interface and types documented in the README, so maybe I'm missing something? .. but most of the code deals with type string, not ReactNode)

framework/PageForm/PageForm.tsx
marshmalien: This is a great solution to clearing form-level errors.

frontend/awx/adapters/awxErrorAdapter.tsx
```diff
@@ -0,0 +1,40 @@
+import { ErrorOutput, GenericError } from '../../../framework/PageForm/typesErrorAdapter';
+import { isRequestError } from '../../common/crud/RequestError';
+
+export const awxErrorAdapter = (error: unknown): ErrorOutput => {
+  const genericErrors: GenericError[] = [];
+  const fieldErrors: GenericError[] = [];
+
+  if (
```
kialam, line 8: This conditional feels overloaded. Is there a way to use a `Switch` statement that checks for `__all__ ` within `error.json` instead?

frontend/hub/adapters/hubErrorAdapter.tsx
```diff
@@ -0,0 +1,69 @@
...
+  title: string;
+}
+
+export const hubErrorAdapter: ErrorAdapter = (error): ErrorOutput => {
+  // errors can come in several flavors depending on if the API is from
+  // pulp or ansible.
+  const genericErrors: GenericError[] = [];
+  const fieldErrors: GenericError[] = [];
+
+  if (
```
kialam, line 22: Can there be a `ResponseError` interface that checks for these attributes rather than having this evaluated with a set of conditionals?

framework/PageForm/PageForm.tsx
nixocio: I moved the Alert to its own component now, allowing different titles and components, based on the type of error.

frontend/hub/adapters/hubErrorAdapter.tsx
nixocio: Thanks for your review @kialam, I am refactoring those conditions.

himdel (approved):
Looks good :+1:, it *should* handle all the errors I know of :crossed_fingers: :)

(I haven't *thoroughly* checked the awx parts)

jamestalton (approved):
lgtm

nixocio:
I had to rebase this one a couple of times to fix conflicts. Hopefully, last one.

framework/PageForm/genericErrorAdapter.cy.ts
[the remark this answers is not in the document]
nixocio: If an instance of native Error will be treated as form level error.

marshmalien (approved):
Nice work! The changes look good to me!

PR #950, opened 2023-09-25
[AAP-16181] Add timeout to the clipboard copy alerts

lgalis:


nixocio (approved):

framework/hooks/useClipboard.tsx
```diff
@@ -16,14 +17,15 @@ type UseClipboardResult = {
  * - copySuccess: A boolean indicating if the last copy attempt was
  * successful.
  */
-export function useClipboard(): UseClipboardResult {
+export function useClipboard(timeout: number | undefined = undefined): UseClipboardResult {
```
nixocio, line 20:
```suggestion
export function useClipboard(timeout: number  = 2000): UseClipboardResult {
```

@lgalis What about we have a default value?

lgalis: @nixocio - changed for all the copy alerts to time out after the default value ( 8000 )

PR #953, opened 2023-09-26
Remove not used components PageFormAlerts and PageFormButtons

nixocio:
Remove not used components PageFormAlerts and PageFormButtons


marshmalien (approved):
LGTM!

PR #958, opened 2023-09-26
FlexRow and FlexColumn Components

jamestalton:
These are easier to use and understand flex components for aligning items in rows and columns.
The syntax is much cleaner than PatternFlys <Flex> component.
Also by having a Row version and a Column version we tailor the props to be easier to understand.


ZitaNemeckova:
Maybe some prefix to css classes to prevent conflict with PF native CSS classes or console project. WDYT?

keithjgrant:
I have a few concerns here, in addition to Zita's suggestion:

* In general, I’d like to dig into this deeper to better understand the root issue here. If we’re going to replace PF’s flexbox abstraction with our own, I want to first evaluate the use cases where we need and why it a little closer. I'm happy to dig into this myself when I get the time.
* Related to that, you had to add a fullWidth prop, which is kind of a code smell to me. Block level elements are full width by default, so if this is necessary, it means we’re using this on something that’s already the flex item of a parent flexbox (or parent grid). It feels like we’re going against the grain of something PF is doing
* I don’t want to introduce yet another way to do CSS (the `clsx` dependency) until we settle that issue as a team; we’ve mostly landed on styled-components up to this point

jamestalton:
> Maybe some prefix to css classes to prevent conflict with PF native CSS classes or console project. WDYT?

the css classes have a prefix class
> .flex-column.align-center

not sure we need more than that

jamestalton:
@keithjgrant I see some big benefits here, lets meet and discuss.

PR #967, opened 2023-09-28
[WIP] Data cy refactor dropdown multi select

akus062381:
Continuing the work of refactoring the test code to account for data-cy locators, which includes adding locators where they are missing


AlexSCorey (approved):
Looks good to me.

PR #969, opened 2023-09-28
PageRoutedTabs for Jobs, Credentials, Inventories

keithjgrant:
Update routing for Jobs, Credentials, and Inventories.

Also updates `PageRoutedTabs` to allow for `false` in the tabs array, making conditional tabs simpler to define.


AlexSCorey (approved):
This looks good to me.

PR #979, opened 2023-09-29
Update team routes

keithjgrant:
Update Team page routes to use `PageRoutedTabs`


AlexSCorey (approved):
This looks good.  This also feels like a good time to add RBAC to the team tabs if there is any.

PR #982, opened 2023-10-02
Fix 403 error for non-admins in the job template form

marshmalien:
https://issues.redhat.com/browse/AAP-16698

The API returns a 403 when we send `webhook_key` and `webhook_url` to the API as non-admins. This PR removes those two properties from the PATCH request. 

I also removed translation tags from business logic inside the User form.

Reproduce:
1. Create a "Normal" user: _normal_user_
2. Give the user an "Admin" role to a job template: _demo_template_
3. Sign in as _normal_user_
4. Attempt to edit  _demo_template_
5. See error "You do not have permission to perform this action"

Screenshot 2023-10-02 at 1 37 21 PM

### DIFF
--- a/frontend/awx/access/users/UserForm.tsx
+++ b/frontend/awx/access/users/UserForm.tsx
@@ -41,8 +41,8 @@ export function CreateUser() {
       } catch {
         throw new Error(t('Organization not found.'));
       }
-      user.is_superuser = userType === t('System administrator');
-      user.is_system_auditor = userType === t('System auditor');
+      user.is_superuser = userType === 'System administrator';
+      user.is_system_auditor = userType === 'System auditor';
       if (confirmPassword !== user.password) {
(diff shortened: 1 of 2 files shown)


AlexSCorey (approved):
Looks good

frontend/awx/access/users/UserForm.tsx
```diff
@@ -41,8 +41,8 @@ export function CreateUser() {
       } catch {
         throw new Error(t('Organization not found.'));
       }
-      user.is_superuser = userType === t('System administrator');
-      user.is_system_auditor = userType === t('System auditor');
+      user.is_superuser = userType === 'System administrator';
+      user.is_system_auditor = userType === 'System auditor';
```
nixocio, line 45: 💯

vishsanghishetty (approved):

PR #985, opened 2023-10-03
Copy collections to repository

MilanPospisil:


frontend/hub/collections/hooks/useCopyToRepository.tsx
```diff
@@ -0,0 +1,224 @@
...
+            ? [collection.collection_version?.pulp_href]
+            : [],
+          destination_repositories: repoHrefs,
+        };
+
+        if (signingService) {
+          params.signing_service = signingService;
+        }
+
+        const taskResponse = (await postRequest(
```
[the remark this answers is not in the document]
MilanPospisil: There is quite a mess in the methods, there are lots of them. I already included this topic in to the moday meeting so we are all in one page.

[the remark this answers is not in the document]
nixocio: We cleaned up several of HTTP methods, and consolidated them. KY was removed. But it is always a good idea to clarify, and improve if needed.

There is https://github.com/ansible/ansible-ui/tree/main/frontend/hub/api#hub-http-requests

nixocio (commented):
@MilanPospisil Overall good job. Once I selected all inside the modal, should one be able to click and removal all as well? once I bulk selected I could not remove any of them. Is that the expected behavior?

image

MilanPospisil:
> @MilanPospisil Overall good job. Once I selected all inside the modal, should one be able to click and removal all as well? once I bulk selected I could not remove any of them. Is that the expected behavior?
> 
> image

Really? Ah, you mean that click deselect all checkbox, good catch, framework must have another callback for this. Select none in dropdown works, but not this checkbox.

frontend/hub/collections/hooks/useCopyToRepository.tsx
```diff
@@ -0,0 +1,252 @@
...
+      actions={[
+        <Button
+          key="select"
+          variant="primary"
+          id="select"
+          onClick={() => {
+            void (async () => {
+              await copyToRepositories();
+            })();
+          }}
```
himdel, line 182:
```suggestion
          onClick={() => copyToRepositories()}
```

frontend/hub/collections/hooks/useCopyToRepository.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+      if (repos.data?.length > 0) {
+        const reposMapped = repos.data.map((item) => item.repository || ({} as Repository));
+        setFixedRepositories(reposMapped);
+      }
+    }
+
+    void (async () => {
+      await getSelected();
+    });
```
himdel, line 145:
```suggestion
    getSelected();
```

(unless this is a bug and was supposed to await?)

himdel: (possibly the whole thing could be rewrritten to

```js
useEffect(() => {
  const { name, version } = collection!.collection_version;
  request(hubAPI`/v3/plugin/ansible/search/collection-versions?limit=100000&name=${name}&version=${version}`)
    .then(({ data }) => {
      if (data.length) {
        const reposMapped = data.map((item) => item.repository || ({} as Repository));
        setFixedRepositories(reposMapped);
      }
    });
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);
```

frontend/hub/collections/hooks/useCopyToRepository.tsx
```diff
@@ -0,0 +1,252 @@
...
+      setIsLoading(false);
+    }
+  };
+
+  useEffect(() => {
+    async function getSelected() {
+      const repos = await request(
+        hubAPI`/v3/plugin/ansible/search/collection-versions?limit=100000&name=${
+          collection.collection_version?.name || ''
+        } &&version=${collection.collection_version?.version || ''}`
```
himdel, line 134:
```suggestion
        }&version=${collection.collection_version?.version || ''}`
```

frontend/hub/collections/hooks/useCopyToRepository.tsx
```diff
@@ -0,0 +1,252 @@
...
+      if (signingService) {
+        params.signing_service = signingService;
+      }
+
+      let api_op = operation as string;
+      if (operation == 'approve') {
+        api_op = 'move';
+      }
+
+      // TODO test errors
```
himdel, line 113: is the TODO still relvant?

frontend/hub/collections/hooks/useCopyToRepository.tsx
```diff
@@ -0,0 +1,252 @@
...
+      };
+
+      if (signingService) {
+        params.signing_service = signingService;
+      }
+
+      let api_op = operation as string;
+      if (operation == 'approve') {
+        api_op = 'move';
+      }
```
himdel, line 111:
```suggestion
      let api_op = {
        approve: 'move_collection_version',
        copy: 'copy_collection_version',
      };
```

frontend/hub/collections/hooks/useCopyToRepository.tsx
```diff
@@ -0,0 +1,252 @@
...
+      if (!repository) {
+        return;
+      }
+
+      const pulpId = parsePulpIDFromURL(repository?.pulp_href);
+
+      let signingService = '';
+
+      if (autoSign) {
+        const signingServiceName = props.context.settings.GALAXY_COLLECTION_SIGNING_SERVICE;
```
himdel, line 77: `autoSign` is only relevant for approval, not for copy
for copy, the important thing is whether a signing service is configured

```suggestion
      const signingServiceName = props.context.settings.GALAXY_COLLECTION_SIGNING_SERVICE;
      if (((operation === 'approve' && autoSign) || (operation === 'move')) && signingServiceName) {
```

frontend/hub/collections/hooks/useCopyToRepository.tsx
```diff
@@ -0,0 +1,252 @@
...
+        selectItem={(item) => {
+          const newItems = [...selectedRepositories, item];
+          setSelectedRepositories(newItems);
+        }}
+        selectItems={(items) => {
+          const newItems = [...selectedRepositories];
+          for (const item of items) {
+            if (
+              !selectedRepositories.find((item2) => item.name == item2.name) &&
+              !fixedRepositories.find((item2) => item2.name != item.name)
```
himdel, line 230:
```suggestion
              !fixedRepositories.find((item2) => item2.name == item.name)
```

frontend/hub/collections/hooks/useCopyToRepository.tsx
MilanPospisil: This is because of error:  180:13  error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises

frontend/hub/collections/hooks/useCopyToRepository.tsx
MilanPospisil: Well I must say I like await much more than then, it seems readable to me. I use promises directly only when I need to.

frontend/hub/collections/hooks/useCopyToRepository.tsx
MilanPospisil: Yeah, obsolete.

frontend/hub/collections/hooks/useCopyToRepository.tsx
MilanPospisil: Interesting, never used that approach.

frontend/hub/collections/hooks/useCopyToRepository.tsx
MilanPospisil: Ah, thats hidden error that was not even relevant :D Because selected repos are displayed based on both selected and fixed and if selected contains fixed repos (repos that collection is already in), it does not matter, because API will ignore that.

frontend/hub/collections/hooks/useCopyToRepository.tsx
MilanPospisil: Thanks, signing is still mystery for me.

frontend/hub/collections/hooks/useCopyToRepository.tsx
himdel: so... `onClick={() => void copyToRepositories()}` ?

frontend/hub/collections/hooks/useCopyToRepository.tsx
```diff
@@ -0,0 +1,245 @@
...
+        collection={collection}
+        context={context}
+        onClose={onClose}
+        operation={operation}
+      />
+    );
+  };
+}
+
+function CopyToRepositoryModal(props: {
```
vidyanambiar, line 40: This works well for now.
 
I believe we can also utilize the `MultiSelectDialog` framework component here to create the modal for selecting the repositories. We may need to expand it to accept a "disableRow" prop that allows disabling selection on existing repositories.

vidyanambiar (commented):
I copied a collection over to the `validated` and `staging` repositories. But in the list of collections there is only one new entry for that collection with the `validated` repository. There is no new corresponding entry for the `staging` repo. But when the "select repositories" modal comes up for that collection, `staging` is marked as selected. Is that normal?

Minor: In the screenshot in the PR description, the disabled checkboxes are grayed out but when I tested it they behaved correctly (disabled) but  were not grayed out so they looked like they were enabled.
image

frontend/hub/collections/hooks/useCopyToRepository.tsx
MilanPospisil: Yes, as I have wrote this in description, this should be separate PR.

frontend/hub/collections/hooks/useCopyToRepository.tsx
MilanPospisil: Magic :-)

MilanPospisil: But it does not work, the function is than not called, I reset this commit back.

MilanPospisil:
> I copied a collection over to the `validated` and `staging` repositories. But in the list of collections there is only one new entry for that collection with the `validated` repository. There is no new corresponding entry for the `staging` repo. But when the "select repositories" modal comes up for that collection, `staging` is marked as selected. Is that normal?
> 
> Minor: In the screenshot in the PR description, the disabled checkboxes are grayed out but when I tested it they behaved correctly (disabled) but were not grayed out so they looked like they were enabled. image

You dont see collections in staging repo in collection list. It appears in approval list.
The disabled rows will be added in separate PR. I removed this change, because tests failed and I dont want to do all the framework changes in this PR.

frontend/hub/approvals/hooks/useApproveCollections.tsx
```diff
@@ -67,15 +82,32 @@ export function approveCollection(
...
+          )
+        );
+      } else {
+        copyToRepository(collection, 'approve');
       }
-    );
+    } else {
+      await hubAPIPost(
+        pulpAPI`/repositories/ansible/ansible/${
+          parsePulpIDFromURL(collection.repository?.pulp_href || '') || ''
```
himdel, line 103: these seem rather more dangerous with post...

```suggestion
          parsePulpIDFromURL(collection.repository.pulp_href)
```

and make sure to have a condition that checks `collection.repository` first?

frontend/hub/collections/hooks/useCopyToRepository.tsx
```diff
@@ -0,0 +1,243 @@
...
+        operation == 'approve' ? t('Can not approve collection.') : t('Can not copy collection')
+      );
+      setIsLoading(false);
+    }
+  };
+
+  useEffect(() => {
+    async function getSelected() {
+      const repos = await request(
+        hubAPI`/v3/plugin/ansible/search/collection-versions?limit=100000&name=${
```
himdel, line 125: limit=10000 will get replaced with limit=100 api side; you should never rely on having the full list available for anything

(there is probably no use for "Select All", "Select page" should be enough)

MilanPospisil: You may understand it wrong. This is not selecting all repositories, this is loading all repositories the collection is in. You think this may lead to problems with performance? If you are ok with it, we can merge it. But after fix - after rebasing, the checkboxes for selection suddenly dissapeared.

MilanPospisil: The checkboxes started to work after reload suddenly again :-D However, I discovered some problem with API - it does not correctly copies collection into some repo, it seems random, but consistent - for example I cant copy to published 2 (it has approved pipeline), the same for old application. So what about merge this and consult this with API and possible there will be API fix, or UI fix on both new and old API.

himdel: oh, so this is used just to populate the preselected(&disabled) checkboxes?
Feel free to merge, it just won't preselect more than 100 repos, so you may want to add some kind of message like that when there are more.

What's the API problem?

MilanPospisil: We disscussed that in sunday. I probably forgot to create also distribution. Seems not related to this PR.

himdel (approved):

PR #989, opened 2023-10-03
Fix template form to template details navigation bug

marshmalien:
This PR fixes navigation to template details. 

We were missing a trailing '/' when querying organizations by role_level. I hardcoded the role_level query string to fix the 301 error `'/api/v2/organizations/?role_level=notification_admin_role'`

Screenshot 2023-10-03 at 4 36 09 PM

### DIFF
--- a/frontend/awx/resources/templates/TemplatePage/TemplatePage.tsx
+++ b/frontend/awx/resources/templates/TemplatePage/TemplatePage.tsx
@@ -36,9 +36,9 @@ export function TemplatePage() {
     error: isNotifAdminError,
     refresh: refreshNotifAdmin,
     isLoading: isNotifAdminLoading,
-  } = useGet<AwxItemsResponse<Organization>>('/api/v2/organizations', {
-    role_level: 'notification_admin_role',
-  });
+  } = useGet<AwxItemsResponse<Organization>>(
+    '/api/v2/organizations/?role_level=notification_admin_role'
+  );
   const getPageUrl = useGetPageUrl();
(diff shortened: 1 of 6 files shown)


frontend/awx/resources/templates/TemplatePage/TemplatePage.tsx
```diff
@@ -36,9 +36,9 @@ export function TemplatePage() {
     error: isNotifAdminError,
     refresh: refreshNotifAdmin,
     isLoading: isNotifAdminLoading,
-  } = useGet<AwxItemsResponse<Organization>>('/api/v2/organizations', {
-    role_level: 'notification_admin_role',
-  });
+  } = useGet<AwxItemsResponse<Organization>>(
+    '/api/v2/organizations/?role_level=notification_admin_role'
```
keithjgrant, line 40: hmm, this should have worked the way it was. Is there a bug in `useGet`? (or maybe the problem was the missing trailing slash?). You don't necessarily need to change it back... we do it this way in other places, it looks like

keithjgrant (approved):

frontend/awx/resources/templates/TemplatePage/TemplatePage.tsx
marshmalien: @keithjgrant and I chatted more about this 1:1. I checked locally and useGet doesn't handle 301 redirects when there is a trailing missing slash. This is an enhancement we should look into at a later time.

PR #990, opened 2023-10-04
Update routing for credential types, notification templates, management jobs

keithjgrant:


AlexSCorey (approved):
@keithjgrant answered all my questions offline.  RBAC stuff for this side nav tab should wait for other rbac work.

PR #999, opened 2023-10-06
Stop running task

MilanPospisil:
Adds menu to the tasks list and 1 operation that is able to stop running task.

When task cannot be stopped, it can be selected, but the bulk action then starts to complain:

image

### DIFF
--- a/frontend/hub/tasks/Task.tsx
+++ b/frontend/hub/tasks/Task.tsx
@@ -4,7 +4,7 @@ export interface Task {
   pulp_created: string;
   started_at: string;
   finished_at: string;
-  state: 'completed' | 'failed' | 'running';
+  state: 'completed' | 'failed' | 'running' | 'waiting';
   error?: TaskError;
 }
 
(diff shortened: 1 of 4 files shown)


frontend/hub/tasks/Task.tsx
```diff
@@ -4,7 +4,7 @@ export interface Task {
   pulp_created: string;
   started_at: string;
   finished_at: string;
-  state: 'completed' | 'failed';
+  state: 'completed' | 'failed' | 'running';
```
himdel, line 7:
```suggestion
  state: 'completed' | 'failed' | 'running' | 'waiting';
```

waiting tasks can also be stopped

frontend/hub/tasks/hooks/useTasksActions.tsx
```diff
@@ -0,0 +1,69 @@
...
+        keyFn: (item) => item.name,
+        isDanger: true,
+        confirmationColumns,
+        actionColumns,
+        onComplete,
+        actionFn: (task: Task) => stopRunningTask(task),
+        isItemNonActionable: (task: Task) =>
+          task.state == 'running' || task.state == 'waiting'
+            ? ''
+            : 'Can only cancel running or waiting tasks',
```
nixocio, line 58: should be translated?

MilanPospisil: Oh, thanks, it should.

PR #1001, opened 2023-10-06
Collection delete version and handle redirection after delete

MilanPospisil:


frontend/hub/collections/CollectionDetails.tsx
```diff
@@ -52,50 +52,78 @@ import { PageDetail } from '../../../framework/PageDetails/PageDetail';
...
+          // Set a new query parameter or update existing ones
+          newParams.set('redirectIfEmpty', '');
+
+          setSearchParams(newParams);
+        }
+
+        setCollections(res.data);
+      })();
+    } catch (error) {
+      setCollectionError(<AwxError error={{ name: 'not found', message: t('Not Found') }} />);
```
himdel, line 123: this will never happen without `await`, the async function just returns a rejected promise on failure

frontend/hub/api/utils.tsx
```diff
@@ -81,10 +81,16 @@ export function pulpHrefKeyFn(item: { pulp_href: string }) {
...
+  collection_version?: { pulp_href: string; version?: string };
   repository?: { name: string };
 }) {
-  return item.collection_version?.pulp_href + '_' + item.repository?.name;
+  return (
+    item.collection_version?.pulp_href +
+    '_' +
+    item.repository?.name +
+    '_' +
+    item.collection_version?.version
```
himdel, line 92: (nothing wrong with this, but .. a situation where the collection version pulp_href is the same but the version string is different should be impossible in the db)

frontend/hub/collections/hooks/useDeleteCollectionsFromRepository.tsx
```diff
@@ -1,53 +1,122 @@
...
+  version?: boolean
+) {
+  let itemsToDelete: string[] = [];
+
+  if (!version) {
+    // load all associated collection versions
+    const results = await requestGet<HubItemsResponse<CollectionVersionSearch>>(
+      hubAPI`/v3/plugin/ansible/search/collection-versions/?name=${
+        collection.collection_version?.name || ''
+      }&repository_name=${collection.repository?.name || ''}`
```
himdel, line 86: this function probably shouldn't try to call any apis if `collection.repository` is null?

MilanPospisil: I will add the condition for repository, but when repo is null, user should not see the collection in UI anyway.

frontend/hub/collections/hooks/useDeleteCollectionsFromRepository.tsx
```diff
@@ -1,53 +1,127 @@
...
+      hubAPI`/v3/plugin/ansible/search/collection-versions/?name=${
+        collection.collection_version?.name || ''
+      }&repository_name=${collection.repository.name}`
+    );
+    itemsToDelete = results.data?.map((item) => item.collection_version?.pulp_href || '');
+  } else {
+    itemsToDelete.push(collection.collection_version?.pulp_href || '');
+  }
+
+  const res: { task: string } = await postRequest(
```
nixocio, line 98: If we use `hubAPIPost` this one could be simplified.

nixocio (approved):

PR #998, opened 2023-10-05
Revert "Cypress upgraded to v13.1.0"

keithjgrant:
Reverts ansible/ansible-ui#883

PR #1007, opened 2023-10-09
Cypress upgrade to v13.3.0

vidyanambiar:
This PR updates Cypress to version 13 and attempts to update the setting that permits tests to run on PRs when package.json is updated.

PR https://github.com/ansible/ansible-ui/pull/998 reverted the earlier Cypress upgrade to version 13 because a JobOutput component test was failing. The expand/collapse buttons in the job output screen were not being rendered. It did not get caught during the PR review stage because the package.json updates did not trigger component/E2E tests.

### Fixes:

1. @keithjgrant and I found that Cypress adds the `specFile` as a search param in the URL. This caused the JobOutput UI's dependency on `location.search` to show and hide the expand/collapse buttons to be affected. Since JobOutput already has access to the filtered state of the UI, the `location.search` check in this case was redundant so we removed it.
2. The new `specFile` query param affects the table state that we define using the `useView` and `useSearchParams` hooks. These hooks take in the search params and add them to the filters for the table. The unexpected `specPath` param from Cypress was causing failures in an EDA component test for the projects list. As a result we needed to remove the specFile query param from the URL in these hooks. This is a temporary solution until we can get the following bug addressed by Cypress: https://github.com/cypress-io/cypress/issues/28021

### DIFF
--- a/framework/components/useSearchParams.tsx
+++ b/framework/components/useSearchParams.tsx
@@ -7,7 +7,9 @@ export function useSearchParams(): [URLSearchParams, (setSearchParams: URLSearch
   const searchParams = useMemo<URLSearchParams>(() => {
     /** Cypress component tests add a specPath param that must be ignored */
     let search = location.location?.search;
-    if (search && search.includes('?specPath')) {
+    if (search && search.includes('?specPath=')) {
+      search = search.substring(0, search.indexOf('?specPath='));
+    } else if (search && search.includes('&specPath=')) {
       search = search.substring(0, search.indexOf('&specPath='));
     }
     return new URLSearchParams(search ?? '/');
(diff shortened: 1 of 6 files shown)


framework/components/useSearchParams.tsx
```diff
@@ -4,10 +4,15 @@ import { useWindowLocation } from './useWindowLocation';
...
   const pathname = location.location?.pathname || '/';
-  const searchParams = useMemo<URLSearchParams>(
-    () => new URLSearchParams(location.location?.search ?? '/'),
-    [location.location?.search]
-  );
+  const searchParams = useMemo<URLSearchParams>(() => {
+    /** Cypress component tests add a specPath param that must be ignored */
+    let search = location.location?.search;
+    if (search && search.includes('?specPath')) {
+      search = search.substring(0, search.indexOf('&specPath='));
```
keithjgrant, line 11: You search with a `?` in the if statement, but find indexOf with a `&`. Does Cypress always add this at the beginning of the query string, or do we need to check for it anywhere in the string?

vidyanambiar: Thanks for catching that! Updated 👍

vidyanambiar:
The E2E failures do not seem to be related to the Cypress upgrade. They are being separately investigated.

PR #1012, opened 2023-10-10
cleanup old select components

jamestalton:


framework/Settings.tsx
```diff
@@ -148,64 +148,68 @@ export function SettingsDialog(props: { open: boolean; setOpen: (open: boolean)
...
-            placeholderText="Select theme"
-          >
-            <SelectOption value="system">{'System default'}</SelectOption>
-            <SelectOption value="light">{'Light'}</SelectOption>
-            <SelectOption value="dark">{'Dark'}</SelectOption>
-          </FormGroupSelect>
-          <FormGroupSelect
+            placeholder="Select theme"
+            options={[
+              { label: 'System default', value: 'system' },
```
nixocio, line 160: Do we select those labels to be translated?

framework/Settings.tsx
```diff
@@ -148,64 +148,68 @@ export function SettingsDialog(props: { open: boolean; setOpen: (open: boolean)
...
             }
-            placeholderText="Select table layout"
-          >
-            <SelectOption value="comfortable">{'Comfortable'}</SelectOption>
-            <SelectOption value="compact">{'Compact'}</SelectOption>
-          </FormGroupSelect>
-          <FormGroupSelect
+            placeholder="Select table layout"
+            options={[
+              { label: 'Comfortable', value: 'comfortable' },
```
nixocio, line 177: Same comment as above.

nixocio (approved):
@jamestalton it seems like we need to select several labels, and placeholders to be translated.

jamestalton:
Not worrying about translations on the settings dialog as UXD has a new design for it and at that point we will be putting in translations.

PR #1027, opened 2023-10-11
Removes old code, and adds topology dependency

AlexSCorey:
This addes Patternfly topology dependency and removes dead code


marshmalien (approved):
🔥

vishsanghishetty (approved):

PR #1043, opened 2023-10-16
Fix for failing E2E tests on hub

vidyanambiar:
Thanks to @nixocio for bringing these failures to my attention!

These failures were caused by recent routing updates and changes to the cypress commands like 'verifyPageTitle'. They did not get caught because these tests don't run in the CI pipeline yet.


nixocio (approved):
I got this code locally, and the tests passed a local dev server. Good job @vidyanambiar 🚀

PR #1044, opened 2023-10-16
[WIP] Create global project

akus062381:
This PR creates an AWX global project that can be reused across all AWX spec files.
One goal is to prevent the tests from creating a huge number of projects that have to be synced, which adds a lot of time to the overall test run.
It also adds JS doc to explain how the user can utilize the global resource.


vishsanghishetty (approved):
🚀

PR #1046, opened 2023-10-17
repositoryBasePath function

ShaiahWren:
RepositoryBasePath function for download tarball function in Install tab of collection details.

### DIFF
--- a/frontend/hub/api/utils.tsx
+++ b/frontend/hub/api/utils.tsx
@@ -9,44 +9,50 @@ import {
 } from './request';
 import { useTranslation } from 'react-i18next';
 import { requestGet } from '../../common/crud/Data';
+import { AnsibleAnsibleRepositoryResponse as Repository } from '../api-schemas/generated/AnsibleAnsibleRepositoryResponse';
+import { AnsibleAnsibleDistributionResponse as Distribution } from '../api-schemas/generated/AnsibleAnsibleDistributionResponse';
 
 function getBaseAPIPath() {
   return process.env.HUB_API_PREFIX;
 }
 
-export function useRepositoryBasePath(name: string, pulp_href?: string): Promise<unknown> {
+type Results = { data: { results: Repository[] | Distribution[] } };
+
+export function useRepositoryBasePath(name: string, pulp_href?: string): Promise<string> {
   const { t } = useTranslation();
-  return Promise.all([
-    pulp_href
-      ? Promise.resolve({ name, pulp_href })
-      : requestGet<{ data: { results: unknown } }>(
-          pulpAPI`/repositories/ansible/ansible/?name=${name}&limit=1`
-        ).then(firstResult),
-    requestGet<{ data: { results: unknown } }>(
-      pulpAPI`/distributions/ansible/ansible/?name=${name}&limit=1`
-    ).then(firstResult),
-  ]).then(async ([repository, distribution]) => {
+
+  const repoRequest = pulp_href
+    ? Promise.resolve({ name, pulp_href } as Repository)
+    : requestGet<Results>(pulpAPI`/repositories/ansible/ansible/?name=${name}&limit=1`).then(
+        firstResult
+      );
+
+  const distroRequest = requestGet<Results>(
+    pulpAPI`/distributions/ansible/ansible/?name=${name}&limit=1`
+  ).then(firstResult);
+
+  return Promise.all(repoRequest, distroRequest).then(async ([repository, distribution]) => {
     if (!repository) {
-      return Promise.reject(t`Failed to find repository ${name}`);
+      throw new Error(t`Failed to find repository ${name}`);
     }
 
     if (distribution && distribution.repository === repository.pulp_href) {
       return distribution.base_path;
     }
 
-    distribution = await requestGet(
+    distribution = await requestGet<Results>(
       pulpAPI`/distributions/ansible/ansible/?repository=${repository.pulp_href}&ordering=pulp_created&limit=1`
     ).then(firstResult);
 
     if (!distribution) {
-      return Promise.reject(t`Failed to find a distribution for repository ${name}`);
+      throw new Error(t`Failed to find a distribution for repository ${name}`);
     }
 
     return distribution.base_path;
   });
 }
 
-function firstResult({ data: { results } }) {
+function firstResult({ data: { results } }: Results) {
   return results[0];
 }
 
--- a/frontend/hub/api/utils.tsx
+++ b/frontend/hub/api/utils.tsx
@@ -18,7 +18,7 @@ function getBaseAPIPath() {
   return process.env.HUB_API_PREFIX;
 }
 
-type Results = { data: { results: Repository[] | Distribution[] } };
+type Results = { results: Repository[] | Distribution[] };
 export function useRepositoryBasePath(name: string, pulp_href?: string | undefined) {
   const { t } = useTranslation();
 
--- a/frontend/hub/api/utils.tsx
+++ b/frontend/hub/api/utils.tsx
@@ -19,13 +19,12 @@ function getBaseAPIPath() {
 }
 
 type Results = { data: { results: Repository[] | Distribution[] } };
-export function useRepositoryBasePath(name: string | undefined, pulp_href?: string | undefined) {
-  console.log(name, pulp_href, 'useRepositoryBasePath');
+export function useRepositoryBasePath(name: string, pulp_href?: string | undefined) {
   const { t } = useTranslation();
 
-  const [basePath, setBasePath] = useState<string | null>('blahahhh');
+  const [basePath, setBasePath] = useState<string>('');
   const [loading, setLoading] = useState<boolean>(true);
-  const [error, setError] = useState<string | null>(null);
+  const [error, setError] = useState<string>('');
 
   useEffect(() => {
     setLoading(true);
@@ -75,7 +74,6 @@ export function useRepositoryBasePath(name: string | undefined, pulp_href?: stri
   }, []);
 
   const x = { basePath, loading, error };
-  console.log(x, 'useRepositoryBasePath result');
   return x;
 }
 
--- a/frontend/hub/api/utils.tsx
+++ b/frontend/hub/api/utils.tsx
@@ -7,9 +7,47 @@ import {
   postHubRequest,
   putHubRequest,
 } from './request';
+import { useTranslation } from 'react-i18next';
+import { requestGet } from '../../common/crud/Data';
 
 function getBaseAPIPath() {
   return process.env.HUB_API_PREFIX;
+}
+
+export function useRepositoryBasePath(name: string, pulp_href?: string): Promise<unknown> {
+  const { t } = useTranslation();
+  return Promise.all([
+    pulp_href
+      ? Promise.resolve({ name, pulp_href })
+      : requestGet<{ data: { results: unknown } }>(
+          pulpAPI`/repositories/ansible/ansible/?name=${name}&limit=1`
+        ).then(firstResult),
+    requestGet<{ data: { results: unknown } }>(
+      pulpAPI`/distributions/ansible/ansible/?name=${name}&limit=1`
+    ).then(firstResult),
+  ]).then(async ([repository, distribution]) => {
+    if (!repository) {
+      return Promise.reject(t`Failed to find repository ${name}`);
+    }
+
+    if (distribution && distribution.repository === repository.pulp_href) {
+      return distribution.base_path;
+    }
+
+    distribution = await requestGet(
+      pulpAPI`/distributions/ansible/ansible/?repository=${repository.pulp_href}&ordering=pulp_created&limit=1`
+    ).then(firstResult);
+
+    if (!distribution) {
+      return Promise.reject(t`Failed to find a distribution for repository ${name}`);
+    }
+
+    return distribution.base_path;
+  });
+}
+
+function firstResult({ data: { results } }) {
+  return results[0];
 }
 
 export function apiTag(strings: TemplateStringsArray, ...values: string[]) {
--- a/frontend/hub/api/utils.tsx
+++ b/frontend/hub/api/utils.tsx
@@ -23,11 +23,12 @@ export function useRepositoryBasePath(name: string | undefined, pulp_href?: stri
   console.log(name, pulp_href, 'useRepositoryBasePath');
   const { t } = useTranslation();
 
-  const [basePath, setBasePath] = useState<string | null>(null);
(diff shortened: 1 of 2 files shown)


frontend/hub/api/utils.tsx
```diff
@@ -7,11 +7,44 @@ import {
...
 }
 
+export function repositoryBasePath(name: string, pulp_href?: string): Promise<string> {
+  const { t } = useTranslation();
+  return Promise.all([
+    pulp_href
+      ? Promise.resolve({ name, pulp_href })
+: requestGet(`repositories/ansible/ansible/?name=${name}?limit=1`).then(firstResult),
+
+      requestGet(`distributions/ansible/ansible/?name=${name}?limit=1`).then(firstResult),
```
himdel, line 24:
```suggestion
: requestGet(pulpAPI`/repositories/ansible/ansible/?name=${name}&limit=1`).then(firstResult),

      requestGet(pulpAPI`/distributions/ansible/ansible/?name=${name}&limit=1`).then(firstResult),
```

frontend/hub/api/utils.tsx
```diff
@@ -7,11 +7,44 @@ import {
...
+  ]).then(async ([repository, distribution]) => {
+    if (!repository) {
+      return Promise.reject(t`Failed to find repository ${name}`);
+    }
+
+    if (distribution && distribution.repository === repository.pulp_href) {
+      return distribution.base_path;
+    }
+
+    distribution = await requestGet(`distributions/ansible/ansible/?repository=${repository.pulp_href}?ordering=pulp_created?limit=1`).then(firstResult);
```
himdel, line 34:
```suggestion
    distribution = await requestGet(pulpAPI`distributions/ansible/ansible/?repository=${repository.pulp_href}&ordering=pulp_created&limit=1`).then(firstResult);
```

ShaiahWren:
@himdel Thanks for the suggestions, the issue now is the eslinter which shows many pesky errors.

frontend/hub/api/utils.tsx
```diff
@@ -7,11 +7,45 @@ import {
...
   return process.env.HUB_API_PREFIX;
 }
 
+export function useRepositoryBasePath(name: string, pulp_href?: string): Promise<unknown> {
+  const { t } = useTranslation();
+  return Promise.all([
+    pulp_href
+      ? Promise.resolve({ name, pulp_href })
+      : requestGet(pulpAPI`/repositories/ansible/ansible/?name=${name}?limit=1`).then(firstResult),
+    requestGet(pulpAPI`/distributions/ansible/ansible/?name=${name}?limit=1`).then(firstResult),
```
himdel, line 23:
```suggestion
      : requestGet(pulpAPI`/repositories/ansible/ansible/?name=${name}&limit=1`).then(firstResult),
    requestGet(pulpAPI`/distributions/ansible/ansible/?name=${name}&limit=1`).then(firstResult),
```

himdel:
> `Argument of type '({ data: { results } }: { data: { results: any; }; }) => any' is not assignable to parameter of type '(value: unknown) => any'.`

I guess you may need to experiment a bit to see if the function is even getting called with an object that has a data key inside, or if requestGet unwraps data, or even data & results.

But.. calling `requestGet(something)` will produce the `unknown` type, but you can also call `requestGet<type>(something)` instead.

So .. in your case, `requestGet<{ data: { results: any; } }>(pulpAPI....` should do the trick :).

And then you'll have to figure out which type to use instead of any, presumably it's already autogenerated in `api-schemas/generated/`

ShaiahWren:
@himdel do you mind looking at the linter one more time? :)

frontend/hub/api/utils.tsx
```diff
@@ -7,11 +7,49 @@ import {
...
+  ]).then(async ([repository, distribution]) => {
+    if (!repository) {
+      return Promise.reject(t`Failed to find repository ${name}`);
+    }
+
+    if (distribution && distribution.repository === repository.pulp_href) {
+      return distribution.base_path;
+    }
+
+    distribution = await requestGet(
```
himdel, line 37:
```suggestion
    distribution = await requestGet<{ data: { results: unknown } }>(
```

frontend/hub/api/utils.tsx
```diff
@@ -7,11 +7,49 @@ import {
...
+
+    if (!distribution) {
+      return Promise.reject(t`Failed to find a distribution for repository ${name}`);
+    }
+
+    return distribution.base_path;
+  });
+}
+
+function firstResult({ data: { results } }) {
```
himdel, line 49: yeaaah, if you want to type this right, this will have to be 

```suggestion
function firstResult<T>({ data: { results: T[] } }) {
```

and then instead of calling `.then(firstResult)` it'd be `.then(firstResult<AnsibleRepository>)` (and the type is different depending on distro/repo).

Or you could try 

```suggestion
function firstResult({ data: { results } }: { data: { results: unknown[] } }) {
```

himdel: (might also be a good idea to take all those `{ data: { results: unknown } }` and `{ data: { results: unknown[] } }` and move the definition up into a `type Results = { data: { results: unknown[] } };` and use that instead)

frontend/hub/api/utils.tsx
```diff
@@ -7,11 +7,49 @@ import {
...
+      ? Promise.resolve({ name, pulp_href })
+      : requestGet<{ data: { results: unknown } }>(
+          pulpAPI`/repositories/ansible/ansible/?name=${name}&limit=1`
+        ).then(firstResult),
+    requestGet<{ data: { results: unknown } }>(
+      pulpAPI`/distributions/ansible/ansible/?name=${name}&limit=1`
+    ).then(firstResult),
+  ]).then(async ([repository, distribution]) => {
+    if (!repository) {
+      return Promise.reject(t`Failed to find repository ${name}`);
```
nixocio, line 30: For what I have seen in the code basis, we are adding the standard like like below.
```suggestion
      return Promise.reject(new Error(t`Failed to find repository ${name}`));
```

frontend/hub/api/utils.tsx
```diff
@@ -7,11 +7,49 @@ import {
...
+  return Promise.all([
+    pulp_href
+      ? Promise.resolve({ name, pulp_href })
+      : requestGet<{ data: { results: unknown } }>(
+          pulpAPI`/repositories/ansible/ansible/?name=${name}&limit=1`
+        ).then(firstResult),
+    requestGet<{ data: { results: unknown } }>(
+      pulpAPI`/distributions/ansible/ansible/?name=${name}&limit=1`
+    ).then(firstResult),
+  ]).then(async ([repository, distribution]) => {
```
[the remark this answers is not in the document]
nixocio: @ShaiahWren I took a look at the code and a possible change would something like

```

export async function useRepositoryBasePath(name: string, pulp_href?: string): Promise<unknown> {

  try {
    let repository;
    if (pulp_href) {
      repository = { name, pulp_href };
    } else {
      const repositoryResponse = await requestGet<{ data: { results: unknown } }>(
        pulpAPI`/repositories/ansible/ansible/?name=${name}&limit=1`
      );
      repository = firstResult(repositoryResponse);
    }
    if (!repository) {
      throw new Error(t`Failed to find repository ${name}`);
    }
    const distributionResponse = await requestGet<{ data: { results: unknown } }>(
      pulpAPI`/distributions/ansible/ansible/?name=${name}&limit=1`
    );
    let distribution = firstResult(distributionResponse);

    if (distribution && distribution.repository === repository.pulp_href) {
      return distribution.base_path;
    }

    const newDistResponse = await requestGet<{ data: { results: unknown } }>(
      pulpAPI`/distributions/ansible/ansible/?repository=${repository.pulp_href}&ordering=pulp_created&limit=1`
    );
    distribution = firstResult(newDistResponse);

    if (!distribution) {
      throw new Error(t`Failed to find a distribution for repository ${name}`);
    }

    return distribution.base_path;
  } catch (error) {
    if (isRequestError(error)) {
      throw error;
    } else {
      throw new Error(t`Failed to find repository base path`);
    }
  }
}


```

It will be required to verify that we are achieving the same with this code. Just an idea how we could refactor this one.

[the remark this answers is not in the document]
himdel: This will take twice as long to wait on the server, won't it?

And any exception will cause "Failed to find repository base path", none of the error messages in the if statements will work since those are now exceptions?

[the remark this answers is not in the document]
himdel: `await` is not great when working with multiple promises,
and it makes it hard to re-throw the exception on failure.

If you want a more readable version with the same semantics, what about

```js
type Results = { data: { results: Repository[] | Distribution[] } };

export function useRepositoryBasePath(name: string, pulp_href?: string): Promise<string> {

  const repoRequest = pulp_href ? Promise.resolve({ name, pulp_href }) : requestGet<Results>(
    pulpAPI`/repositories/ansible/ansible/?name=${name}&limit=1`
  ).then(firstResult);

  const distroRequest = requestGet<Results>(
    pulpAPI`/distributions/ansible/ansible/?name=${name}&limit=1`
  ).then(firstResult);

  return Promise.all(repoRequest, distroRequest).then(async ([ repository, distribution ]) => {
    if (!repository) {
      throw new Error(t`Failed to find repository ${name}`);
    }

    if (distribution && distribution.repository === repository.pulp_href) {
      return distribution.base_path;
    }

    distribution = await requestGet<Results>(
      pulpAPI`/distributions/ansible/ansible/?repository=${repository.pulp_href}&ordering=pulp_created&limit=1`
    ).then(firstResult);

    if (!distribution) {
      throw new Error(t`Failed to find a distribution for repository ${name}`);
    }

    return distribution.base_path;
  });
}

function firstResult({ data: { results } }: Results) {
  return results[0];
}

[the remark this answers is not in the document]
Comment (nixocio) [reply frontend/hub/api/utils.tsx]: Yeah, if we want to keep the requests in parallel we could add back the Promise.all. Not sure how much performative will be that, but if we can make those requests in parallel, +1.

The requests client will raise HTTP RequestError if there is any error underline on the fetch itself.

I like your idea @himdelI it does not need to be part of this PR, but we could have small functions to fetch some resources that we will need to fetch over and over. Something like: 


```js
 const getDistribution = async (distroName: string) => {
    const response = await requestGet<Results>(
      pulpAPI`/distributions/ansible/ansible/?name=${distroName}&limit=1`
    );
    return firstResult(response);
  };

```

Thanks @ShaiahWren and @himdel for looking into improve readability here.

frontend/hub/api/utils.tsx
```diff
@@ -7,11 +7,55 @@ import {
...
+    ? Promise.resolve({ name, pulp_href } as Repository)
+    : requestGet<Results>(pulpAPI`/repositories/ansible/ansible/?name=${name}&limit=1`).then(
+        firstResult
+      );
+
+  const distroRequest = requestGet<Results>(
+    pulpAPI`/distributions/ansible/ansible/?name=${name}&limit=1`
+  ).then(firstResult);
+
+  return Promise.all(repoRequest, distroRequest).then(async ([repository, distribution]) => {
```
nixocio, line 34: This needs to be an array
```suggestion
  return Promise.all([repoRequest, distroRequest]).then(async ([repository, distribution]) => {
```

frontend/hub/api/utils.tsx
ShaiahWren: Hey Kersom, 
I've added some console logs in CollectionDetails.tsx that use `useRepositoryBasePath()` function. The issue (I think) is that the Promise returns the object not the result, and I wanted ask where it should be resolved. 

The console logs in the download() function can be ignored for now, just stuck them in there to confirm values. 

Thank you~

nixocio: Hey @ShaiahWren the useRepositoryBasePath is returning a Promised that was not yet resolved yet.

With the current implementation we could do something like


```js
useRepositoryBasePath(name, pulp_href).then(response => console.log(response))

```

We could either update `useRepositoryBasePath` to return the resolved promised, for instance adding a `useEffect` and  `useState` to that hook. We could also add, error, and isLoading for example as part of the response.

Or on the component consuming useRepositoryBasePath add an `useEffect` and `useState` to fetch the data once the component mounts then fetch data, and update the state.

There are other hooks fetching data that they are adding an inner async function using useGetRequest as well, take a look at useRepoQueryOptions for example.

himdel:
This can't just be `useRepoBasePath` .. we need it in a function form.

Feel free to add a hook wrapper, that can be nice, but most of the uses won't be able to use it.

nixocio:
> This can't just be `useRepoBasePath` .. we need it in a function form.
> 
> Feel free to add a hook wrapper, that can be nice, but most of the uses won't be able to use it.


I like this idea, @ShaiahWren perhaps the core function inside the useEffect could be extracted as the funciton form mentioned by @himdel.

frontend/hub/collections/CollectionDetails.tsx
```diff
@@ -272,15 +280,28 @@ export function CollectionDetails() {
...
   const { t } = useTranslation();
+  const downloadLinkRef = React.useRef<HTMLAnchorElement>(null);
   const { collection } = props;
+  const { basePath, error, loading } = useRepositoryBasePath(
+    collection.repository?.name ?? '',
+    collection.repository?.pulp_href
+  );
+
+  if (loading) {
+    return <>{loading}</>;
```
nixocio, line 299:
```suggestion
    return <LoadingPage breadcrumbs tabs />;
``

File: frontend/hub/collections/CollectionDetails.tsx [hunk 10]
```diff
@@ -272,15 +280,28 @@ export function CollectionDetails() {
...
+  }
+  if (error) {
+    return <>{error}</>;
```
nixocio, line 302:
```suggestion
  if (error) {
    return <AwxError error={new Error(error as string)} />;
  }
```

nixocio (commented):
@ShaiahWren we could add those components as part of the details page.

frontend/hub/api/utils.tsx
```diff
@@ -7,11 +7,80 @@ import {
...
+      } finally {
+        setLoading(false);
+      }
+    };
+
+    void fetchRepositoryBasePath();
+  }, [name, pulp_href, t]);
+
+  const x = { basePath, loading, error };
+  return x;
```
himdel, line 77:
```suggestion
  return { basePath, loading, error };
```

nixocio (approved):

MilanPospisil:
> This can't just be `useRepoBasePath` .. we need it in a function form.
> 
> Feel free to add a hook wrapper, that can be nice, but most of the uses won't be able to use it.

Now Im trying to use it and hook is not good in my situation :D

PR #1054, opened 2023-10-18
Execution Environment add or edit

MilanPospisil:


AlexSCorey (changes requested):
Milan and I discussed these changes off line.  This is a great start.  Let make some changes to help all AAP UI projects unify around some conventions.

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
```diff
@@ -0,0 +1,346 @@
...
+import { ExecutionEnvironment } from './ExecutionEnvironment';
+import { HubPageForm } from '../HubPageForm';
+import { HubItemsResponse } from '../useHubView';
+import { useState, useEffect, useCallback } from 'react';
+
+import { patchHubRequest, putHubRequest } from './../api/request';
+import { PageFormAsyncSelect } from '../../../framework/PageForm/Inputs/PageFormAsyncSelect';
+import { useSelectRegistrySingle } from './hooks/useRegistrySelector';
+
+export function ExecutionEnvironmentForm(props: { mode: 'add' | 'edit' }) {
```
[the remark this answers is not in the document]
MilanPospisil: Isnt that antipattern? You have to maintain some copy paste code?

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
```diff
@@ -0,0 +1,346 @@
...
+  );
+}
+
+function validateName(name: string, t: TFunction<'translation', undefined>) {
+  const regex = /^([0-9A-Za-z._-]+\/)?[0-9A-Za-z._-]+$/;
+  if (regex.test(name)) {
+    return undefined;
+  } else {
+    return t(
+      `Container names can only contain alphanumeric characters, ".", "_", "-" and a up to one "/".`
```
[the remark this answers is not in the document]
MilanPospisil: Its from old application, unless UX changes this, I dont feel I could.

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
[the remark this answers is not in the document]
MilanPospisil: However, I solved the problem of conditional useGet, when you pass different URL, it will do another request and when you pass empty string, it will do nothing. So I was able to rewrite it using mode. If you are comfortable with this solution, it would be totaly ok for me:

I can keep mode under the hood, but I will expose two components Add and Edit form, which will contains component with mode. But outside of form file, there will be no mode.

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
```diff
@@ -0,0 +1,346 @@
...
+}) {
+  const [tagsText, setTagsText] = useState<string>('');
+  const { tags, setTags, mode } = props;
+  const { t } = useTranslation();
+
+  const label = mode == 'exclude' ? t('Add tag(s) to exclude') : t('Add tag(s) to include');
+  const label2 = mode == 'exclude' ? t('Currently excluded tags') : t('Currently included tags');
+
+  const chipGroupProps = () => {
+    const count = '${remaining}'; // pf templating
```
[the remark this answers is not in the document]
MilanPospisil: No idea :-D Copy pasted from old code :-)

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
```diff
@@ -0,0 +1,346 @@
...
+          )}
+
+          <PageFormTextArea<ExecutionEnvironmentFormProps>
+            name="description"
+            label={t('Description')}
+            placeholder={t('Enter a description')}
+            isDisabled={mode == 'add'}
+          />
+
+          {error && <AwxError error={new Error(error)} />}
```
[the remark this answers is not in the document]
MilanPospisil: Yes, Kersom explained me the errors handling.

frontend/hub/useHubNavigation.tsx
```diff
@@ -114,6 +115,16 @@ export function useHubNavigation() {
                 label: t('Execution Environments'),
                 path: 'execution-environments',
                 children: [
+                  {
+                    id: HubRoute.CreateExecutionEnvironment,
+                    path: 'create',
+                    element: <ExecutionEnvironmentForm mode="add" />,
```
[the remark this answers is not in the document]
MilanPospisil: Already handled.

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
```diff
@@ -0,0 +1,346 @@
...
+            validate={(name: string) => validateName(name, t)}
+          />
+
+          {!isRemote && (
+            <PageFormTextInput<ExecutionEnvironmentFormProps>
+              name="namespace.name"
+              label={t('Namespace')}
+              placeholder={t('Enter a namespace name')}
+              isRequired
+              isDisabled
```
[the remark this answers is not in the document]
MilanPospisil: Thats copy paste mistake :)

nixocio (approved):
@MilanPospisil I got the latest version locally, and it seems to be working as expected. 

Good job.

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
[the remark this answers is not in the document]
AlexSCorey: I like the new approach.

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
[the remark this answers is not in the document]
AlexSCorey:
```suggestion
      `Container names can only contain alphanumeric characters, ".", "_", "-" and up to one "/".`
```

AlexSCorey (changes requested):
Just update that 1 string and I think we will be good to go.

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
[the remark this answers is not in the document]
himdel: it's to make sure nobody rewrites that to `` `${remaining}` `` - it *looks* like it should be a template string, but isn't, patternfly is doing it's own template string handling with this

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+}) {
+  const [tagsText, setTagsText] = useState<string>('');
+  const { tags, setTags, mode } = props;
+  const { t } = useTranslation();
+
+  const label = mode == 'exclude' ? t('Add tag(s) to exclude') : t('Add tag(s) to include');
+  const label2 = mode == 'exclude' ? t('Currently excluded tags') : t('Currently included tags');
+
+  const chipGroupProps = () => {
+    const count = '${remaining}';
```
himdel, line 274: maybe keep the comment please? This looks like a bug without some kind of comment that it isn't

frontend/common/crud/useGet.tsx
```diff
@@ -11,8 +11,10 @@ export function useGet<T>(
   url: string | undefined,
   query?: Record<string, string | number | boolean>,
   swrConfiguration: SWRConfiguration = {}
+  // disables running the request, useful when you dont want to actualy run the hook in certain conditions
```
himdel, line 14: what does this comment belongs with?

if it's about swrConfiguration, put it above,
if about useGet, put it inside

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+  const getPageUrl = useGetPageUrl();
+  const mode = props.mode;
+  const params = useParams<{ id?: string }>();
+
+  const [tagsToInclude, setTagsToInclude] = useState<string[]>([]);
+  const [tagsToExclude, setTagsToExclude] = useState<string[]>([]);
+  const [tagsSet, setTagsSet] = useState<boolean>(false);
+
+  const registry = useGet<HubItemsResponse<Registry>>(
+    hubAPI`/_ui/v1/execution-environments/registries?page_size=${page_size.toString()}`
```
himdel, line 48: this is not filtering by name

the registry selector should be a typeahead select, so that it can work with >50 repos

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+      collapsedText: t(`{{count}} more`, count.toString()),
+      expandedText: t(`Show Less`),
+    };
+  };
+
+  const addTags = () => {
+    if (tagsText == '' || !tagsText.trim().length) {
+      return;
+    }
+    const tagsArray = tagsText.split(',');
```
himdel, line 285:
```suggestion
    const tagsArray = tagsText.split(/\s+|\s*,\s*/).filter(Boolean);
```

split on any combination of commas and spaces, not just a singular comma (and throw away any empty fragments)

`"  foo    bar,baz,   quux  ,"` should resolve to `foo`, `bar`, `baz` and `quux` (instead of `  foo    bar`, `baz`, `  quux  ` and empty string)

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+
+  const registry = useGet<HubItemsResponse<Registry>>(
+    hubAPI`/_ui/v1/execution-environments/registries?page_size=${page_size.toString()}`
+  );
+
+  const eeUrl = useMemo(
+    () =>
+      mode == 'edit'
+        ? hubAPI`/v3/plugin/execution-environments/repositories/${params?.id || ''}/`
+        : '',
```
himdel, line 55: `params?.id || ''` implies params.id can be an empty string...

such an url would return a list of repositories in dev mode, and a 404 in production => not what we want :)

```suggestion
      mode == 'edit' && params?.id
        ? hubAPI`/v3/plugin/execution-environments/repositories/${params.id}/`
        : '',
```

himdel: (also, strings are compared by value, not object identity, so the `useMemo` is probably superfluous, but up to you :))

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+    [mode, params?.id]
+  );
+
+  const executionEnvironment = useGet<ExecutionEnvironment>(eeUrl);
+
+  const singleRegistryUrl = useMemo(
+    () =>
+      mode == 'edit' && executionEnvironment.data
+        ? hubAPI`/_ui/v1/execution-environments/registries/${
+            executionEnvironment.data?.pulp?.repository?.remote?.registry || ''
```
himdel, line 65: same issue as above .. if registry is falsy, the url is wrong

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
himdel: (also, `chilGroupProps` should probably be a shared helper, since it's needed for any ChipGroup component to be localized)

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
```diff
@@ -0,0 +1,333 @@
...
+    ? !!executionEnvironment.data?.pulp?.repository?.remote
+    : true;
+
+  const query = useCallback(async () => {
+    return Promise.resolve({
+      total: registry?.data?.meta?.count || 0,
+      values: registry?.data?.data || [],
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [registry.data]);
```
himdel, line 95: this is suspicious .. it doesn't actually call any apis, I asume because useGet already handles that...

and `registry` is not filtering by anything, so this just returns the total count of registries and the first 50 registries

Sounds like this is old code, and both the registry `useGet` and this `query` function should be removed and are replaced by `useRegistrySelector`?

himdel:
The problem where clicking on the "Show less" button in the LabelGroup causes a reload is unrelated to this PR, that's a patternfly bug:

* `LabelGroup` renders items as `<Label ...`, and the last one as `<Label isOverflowLabel ...`,
* `Label` renders a `<span>` .. except with `isOverflowLabel` where it renders a `<button>`
* `<button>` inside form submits the form

The problem is that pattenrfly Label renders a raw `<button>` (which submits forms) instead of a `<button type="button">` (which doesn't).

So, the solution is to fix patternfly, either of these works:

```diff
diff --git a/packages/react-core/src/components/Label/Label.tsx b/packages/react-core/src/components/Label/Label.tsx
index 0bab0335a..101b5c894 100644
--- a/packages/react-core/src/components/Label/Label.tsx
+++ b/packages/react-core/src/components/Label/Label.tsx
@@ -180,7 +180,8 @@ export const Label: React.FunctionComponent<LabelProps> = ({
     }
   };
 
-  const LabelComponent = (isOverflowLabel ? 'button' : 'span') as any;
+  const safeButton = ({ children }) => <button type="button">{children}</button>;
+  const LabelComponent = (isOverflowLabel ? safeButton : 'span') as any;
 
   const button = closeBtn ? (
     closeBtn
```

```diff
diff --git a/packages/react-core/src/components/LabelGroup/LabelGroup.tsx b/packages/react-core/src/components/LabelGroup/LabelGroup.tsx
index 82e8d0376..ba449b3d9 100644
--- a/packages/react-core/src/components/LabelGroup/LabelGroup.tsx
+++ b/packages/react-core/src/components/LabelGroup/LabelGroup.tsx
@@ -104,7 +104,9 @@ export class LabelGroup extends React.Component<LabelGroupProps, LabelGroupState
     });
   }
 
-  toggleCollapse = () => {
+  toggleCollapse = (e) => {
+    e.preventDefault();
+
     this.setState(prevState => ({
       isOpen: !prevState.isOpen,
       isTooltipVisible: Boolean(
```

(not sure how that interacts with the pf5 switch)

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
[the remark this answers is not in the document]
MilanPospisil: Ok, thats minor change.

frontend/common/crud/useGet.tsx
MilanPospisil: Oh, thanks for catching that! I tried to implemenent conditional request, but it showed that useSWR actualy can do this by passing empty URL.

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
MilanPospisil: Explain please more in depth what should this comment says, me and Alex dont know what this is. I will add extended comment into the PR.

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
MilanPospisil: It is not ment to filter by name. useRegistrySelector handles modal with PageTable that allows the filtering. I had typeahead component, but it was banned by James and others that it could affect performance of application. Instead they decided that we will add Browse button into the dropdown that will open modal with filter. I have written function that constructs this modal dynamicaly by passing columns, filters and parameters for useGet for fetching data. But framework also needs actual data for the dropdown. Im planning to return this "query" from my function, because it can be generated by the data already passed for modal. But for now, it is not implemented and so it has to be two separated calls - one for modal and other for dropdown list.

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
MilanPospisil: You are right, it works. I was not sure how it is compared in SWR (if it was identity, it would need useMemo). I was there a little paranoid.

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
MilanPospisil: It will be replaced in the future, I have to add this into the framework first. For now, I discovered I can remove async, but query requires  function that will fetch data and return promise, so thats why it looks suspicious. I had useGetRequest for that inside the function with await, but since Kersom and Alex told me, we should use useGet everywhere, I fetch data outside and returned its result there.

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
nixocio: Ideally we should add a chipgroup wrapper around PF providing the selected strings translated. That one was done for AWX, to avoid to have provide the translations over and over. IMO this could a future enhancement for the framework itself. Something like below.

```jsx
import React from 'react';

import { t } from '@lingui/macro';
import { ChipGroup as PFChipGroup } from '@patternfly/react-core';
import { number } from 'prop-types';

function ChipGroup({ numChips, totalChips, ...props }) {
  return (
    <PFChipGroup
      {...props}
      numChips={numChips}
      expandedText={t`Show less`}
      collapsedText={t`${totalChips - numChips} more`}
    />
  );
}

ChipGroup.propTypes = {
  numChips: number.isRequired,
  totalChips: number.isRequired,
};

export default ChipGroup;```

File: frontend/common/crud/useGet.tsx [hunk 14]
```diff
@@ -23,12 +27,21 @@ export function useGet<T>(
...
       isLoading: response.isLoading,
+      /**
+       * Renders error or loading element, can be used to simplify loading and error logic in application.
+       */
+      renderErrorOrLoading: error ? (
+        <AwxError error={error} handleRefresh={refresh} />
+      ) : response.isLoading ? (
+        <LoadingPage breadcrumbs tabs />
+      ) : null,
     }),
```
nixocio, line 45: @jamestalton please, can you review those changes?

jamestalton: I'm not wild about the useGet which currently only returns data, returning a function that returns react components. But I do see the benefits of it and it solves the problems of consistency in the pages for loading and error handling.

While the name `renderErrorOrLoading` is more descriptive, it would read clearer in the pages if it was `loadingOrError`.
```
if( loadingOrError ) return loadingOrError
```

MilanPospisil: Ok, I will do it old way without any changes in useGet. We will discuss this in future.

jamestalton: I'm not against it. Like I said, I do see the benefits it provides.

frontend/hub/execution-environments/ExecutionEnvironmentForm.tsx
MilanPospisil: Can we do this in separate PR? I would like this to be merged and to separate framework changes.

nixocio (approved):
@MilanPospisil informed that will the last framework changes in a follow up PR.

AlexSCorey (approved):
Looks good.  All of my concerns have been addressed. Well done!

PR #1062, opened 2023-10-18
[AAP-17392] Add hidden field component

nixocio:
Add hidden field component

### DIFF
--- a/framework/PageInputs/PageHiddenInput.cy.tsx
+++ b/framework/PageInputs/PageHiddenInput.cy.tsx
@@ -1,0 +1,68 @@
+/* eslint-disable i18next/no-literal-string */
+import { PageHiddenInput } from './PageHiddenInput';
+import { Flex, FlexItem } from '@patternfly/react-core';
+
+const TestComponent = () => {
+  return (
+    <Flex>
+      <FlexItem>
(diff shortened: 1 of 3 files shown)


framework/PageInputs/HiddenField.tsx
```diff
@@ -0,0 +1,81 @@
...
+      <InputGroup>
+        <TextInput
+          aria-label={t('hidden value')}
+          placeholder="••••••••••••••••••••••"
+          type="password"
+          autoComplete="off"
+          isDisabled={isValueHidden}
+        />
+        {isValueHidden && (
+          <Button variant="control" onClick={() => onClear()}>
```
himdel, line 74: clear should not be enabled if ther is no value

nixocio: @himdel taking a look. Perhaps, renaming the variable, and making sure we have unit tests.

nixocio: @himdel update the logic. And there are unit tests covering both cases.

nixocio:
Perhaps change name since we have PageFormHidden - to be verified.

marshmalien (approved):
LGTM 👍

himdel (approved):

PR #1064, opened 2023-10-18
Awx credential credential tab

keithjgrant:
Adds the credentials list to the credentials tab of a Credential Type page

### DIFF
--- a/frontend/awx/administration/credential-types/CredentialTypePage/CredentialTypeCredentials.cy.tsx
+++ b/frontend/awx/administration/credential-types/CredentialTypePage/CredentialTypeCredentials.cy.tsx
@@ -1,0 +1,29 @@
+import { CredentialTypeCredentials } from './CredentialTypeCredentials';
+
+describe('CredentialTypeCredentials', () => {
+  beforeEach(() => {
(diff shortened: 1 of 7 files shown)


frontend/awx/administration/credential-types/CredentialTypePage/CredentialTypeCredentials.cy.tsx
```diff
@@ -0,0 +1,29 @@
...
+      { method: 'GET', url: '/api/v2/credential_types/*' },
+      { fixture: 'credentialType' }
+    );
+  });
+
+  it('fetches credentials for the correct credential type', () => {
+    cy.intercept(
+      {
+        method: 'GET',
+        url: '/api/v2/credential_types/*/credentials/*',
```
marshmalien, line 15:
```suggestion
        url: '/api/v2/credential_types/1/credentials/*',
```

I don't think we need to include `path` and `initialEntries` on mount if you include the id here. 

```
cy.mount(<CredentialTypeCredentials />);
```

marshmalien (approved):
lgtm!

PR #1080, opened 2023-10-19
Add create remote registry screen

nixocio:
Add create remote registry screen

image


jamestalton (approved):

frontend/hub/remote-registries/RemoteRegistry.tsx
```diff
@@ -19,7 +19,7 @@ export interface RemoteRegistry {
   download_concurrency: number | null;
   proxy_url: string | null;
   write_only_fields: {
-    name: 'client_key' | 'username' | 'password' | 'proxy_username';
+    name: 'client_key' | 'username' | 'password' | 'proxy_username | proxy_password';
```
himdel, line 22:
```suggestion
    name: 'client_key' | 'username' | 'password' | 'proxy_username' | 'proxy_password';
```

frontend/hub/remote-registries/RemoteRegistries.tsx
```diff
@@ -18,7 +18,7 @@ export function RemoteRegistries() {
     toolbarFilters,
     tableColumns,
   });
-  const toolbarActions = useRemoteRegistriesActions();
+  const toolbarActions = useRemoteRegistriesToolbarActions();
```
himdel, line 21: why "toolbar actions"?

This seems misleading because the actions are not only available in toolbars, there's the list entries kebab too.

Didn't we have a convention to use `useFoosActions` for the list screen and `useFooActions` for the detail screens?

PR #1088, opened 2023-10-19
Fetch and display a workflow visualizer node

marshmalien:
https://issues.redhat.com/browse/AAP-16994

node


AlexSCorey (approved):
Lets try to add some tests to this. Then I think it looks good

PR #1104, opened 2023-10-23
[AAP-15683] Convert the Rule Audit pages to use PageRoutedTabs. Add Back to Rule Audit tab

lgalis:
- Convert the Rule Audit pages to use PageRoutedTabs. 
- Add Back to Rule Audit tab
Screenshot from 2023-10-23 15-15-38


tiyiprh:
This would apply to all sections within EDA correct?

vishsanghishetty (approved):

PR #1110, opened 2023-10-24
Login layout, form PF5 updates

keithjgrant:
* Rebuilds login layout from scratch (rather than co-opting modal styles)
* Fixes PF5 issues in Page Form

This includes a refactor of `PageForm` to put error handling in a hook. I've added a `GenericForm` component that provides similar functionality to PageForm, but with a more stripped down layout


framework/PageFramework.css
```diff
@@ -42,16 +40,3 @@
 :root:where(.pf-v5-theme-dark) .bg-lighten {
   background-image: linear-gradient(#ffffff08, #ffffff08);
 }
-
```
jamestalton, line 45: I would like to leave there until we figure out if pf5 needs them for the toolbar to wrap properly.

framework/PageFramework.css
```diff
@@ -1,7 +1,5 @@
 #app {
-  position: fixed;
-  width: 100%;
-  height: 100%;
+  min-block-size: 100dvh;
```
jamestalton, line 2: i think 100dvh breaks mobile. You have to use 100%. Needs testing.

framework/PageForm/GenericForm.tsx
```diff
@@ -0,0 +1,70 @@
+import { useContext, ReactNode, BaseSyntheticEvent } from 'react';
```
jamestalton, line 1: Do we really need GenericForm? I'd like to keep it so that we do not have 2 ways to do the same thing.

framework/PageForm/PageForm.tsx
```diff
@@ -17,16 +16,26 @@ import {
...
 import { useBreakpoint } from '../components/useBreakPoint';
 import { SettingsContext } from '../Settings';
 import { useFrameworkTranslations } from '../useFrameworkTranslations';
 import { ErrorAlert } from './ErrorAlert';
 import { genericErrorAdapter } from './genericErrorAdapter';
 import { ErrorAdapter } from './typesErrorAdapter';
+import { PageFormSubmitButton, PageFormCancelButton } from './PageFormButtons';
+import styled from 'styled-components';
+
+const FormContainer = styled(PageSection)`
```
jamestalton, line 30: why is the form body using m-action--MarginTop?

jamestalton (changes requested):

framework/PageForm/GenericForm.tsx
keithjgrant: It gets us the same function form behavior as `PageForm`, but without the layout behaviors -- which is exactly what the login form needs.

framework/PageForm/PageForm.tsx
keithjgrant: switched to a more appropriate variable

marshmalien (approved):
Looks great and this resolves the wizard footer pf5 upgrade issue! https://issues.redhat.com/browse/AAP-17378
Screenshot 2023-10-24 at 2 33 45 PM

framework/PageFramework.css
keithjgrant: I've reverted this change for now... it's unrelated to the forms stuff. I'll take a closer look at this in a separate PR

framework/PageForm/PageForm.tsx
```diff
@@ -17,16 +16,26 @@ import {
...
+  margin-block-end: var(--pf-v5-global--spacer--xl);
+`;
+
+const FormActionGroup = styled(ActionGroup)`
+  && {
+    margin-block-start: unset;
+  }
+`;
 
 export interface PageFormProps<T extends object> {
```
nixocio, line 40: Could we extend GenericFormProps here? Since they have so many common properties.

framework/PageForm/PageForm.tsx
```diff
@@ -82,6 +86,19 @@ export function PageForm<T extends object>(props: PageFormProps<T>) {
...
   };
 
+  return { form, handleSubmit, error, setError, handleSubmitError, setFieldError };
+}
+
+export function PageForm<T extends object>(props: PageFormProps<T>) {
+  const { errorAdapter = genericErrorAdapter } = props;
+
+  const { form, handleSubmit, error, setError, handleSubmitError, setFieldError } =
+    useFormErrors<T>(props.defaultValue, errorAdapter);
```
nixocio, line 96: +1 for useFormErrors.

framework/PageForm/GenericForm.tsx
```diff
@@ -0,0 +1,70 @@
...
+    <FormProvider {...form}>
+      <Form
+        // eslint-disable-next-line @typescript-eslint/no-misused-promises
+        onSubmit={handleSubmit(
+          async (
+            data,
+            event?: BaseSyntheticEvent & { nativeEvent: { submitter?: { innerHTML: string } } }
+          ) => {
+            setError(null);
+            let isSecondaryButton = false;
```
nixocio, line 44: When I worked on the error on PageForm I got in doubt about this logic, it is for one specific use case - where was necessary to pass the whole form down. Perhaps we could simplify that one, and remove this part logic and associated code.
```suggestion
   
```

framework/PageForm/GenericForm.tsx
```diff
@@ -0,0 +1,70 @@
...
+            if (event !== undefined && event?.nativeEvent?.submitter) {
+              isSecondaryButton =
+                event.nativeEvent.submitter?.innerHTML === props.additionalActionText;
+            }
+            try {
+              await props.onSubmit(
+                data,
+                (error) => setError(error),
+                setFieldError,
+                isSecondaryButton ? form : undefined
```
nixocio, line 55:
```suggestion
                isSecondaryButton ? form : undefined
```

framework/PageForm/PageForm.tsx
keithjgrant: I don't think they need to share an interface just because they happen to be similar. I kind of prefer seeing the whole of PageFormProps listed out in one place

PR #1112, opened 2023-10-24
Reparallel tests

AlexSCorey:


kialam (changes requested):
I'd like a more thorough test plan as part of this PR. If we can capture total run times for one container vs 3 I think that'd be useful data to collect.

jamestalton:
Screenshot 2023-10-30 at 9 35 15 AM

jamestalton:
It would have been nice to have been asked for thoughts/review on disabling the parallelization before it had been merged.
I was around and available. The initial PR name "testing test" gave no indication that this change was going in.

Bumping up the timeout on the cleanup of resources increases the problems with the tests. I would have pointed that out and we would have been able to have a discussion. In addition the problems we were seeing were with the tests, not the parallelization. Investigations into the test results would have identified that.

There has been a request that this PR only get merged to reenable parallelization only after a merge meeting. It seems like we should have had a merge meeting on the initial PR if anything.

The resource cleanup needs to be fixed. It is frustrating that is was merged in the first place and it is frustrating that getting the fix in is blocked by waiting on a merge meeting.

jamestalton (approved):

vidyanambiar (approved):
It seems like the tests are at a stable point now after issues in the test code were recently addressed. 
It would be a good time to turn concurrency back on - the faster runtime, shorter timeout and quicker resource cleanups as a result of it may keep the server from clogging up with resources. 
If we encounter issues, we can revert and investigate further. 👍

keithjgrant (approved):

kialam (approved):

PR #1120, opened 2023-10-25
Adds support for workflow approval list in AWX

mabashian:
Adds the workflow approval list including support for row and table level approve, deny and delete.  The list comes with two basic filters (name and id).  Three of the columns are sortable (name, started and status).  I'm planning on a follow-on PR with the details view implemented.  Here's a screenshot of the list:

Screenshot 2023-10-25 at 12 44 39 PM

I've included component tests for the following scenarios:
    Workflow approvals list renders
    Workflow approvals list has filters for Name and ID
    Filter workflow approvals by name
    Filter workflow approvals by id
    Clicking name table header sorts workflow approvals by name
    Clicking started table header sorts workflow approvals by start time
    Clicking status table header sorts workflow approvals by status
    Delete workflow approval row action is disabled if the user does not have permission to delete workflow approvals
    Delete workflow approval row action is enabled if the user has permission to delete workflow approvals
    Approve row action is enabled if the user has permission to approve
    Deny row action is enabled if the user has permission to deny
    Approve row action is disabled if the user does not have permission to approve
    Deny row action is disabled if the user does not have permission to deny
    Approve row action calls correct API endpoint
    Deny row action calls correct API endpoint
    Delete row action calls correct API endpoint
    Displays error if workflow approvals are not successfully loaded
    Empty state is displayed correctly

### DIFF
--- a/cypress/fixtures/workflowApprovals.json
+++ b/cypress/fixtures/workflowApprovals.json
@@ -1,0 +1,496 @@
+{
+  "count": 6,
+  "next": null,
+  "previous": null,
+  "results": [
+    {
+      "id": 130,
+      "type": "workflow_approval",
+      "url": "/api/v2/workflow_approvals/130/",
+      "related": {
+        "created_by": "/api/v2/users/2/",
+        "unified_job_template": "/api/v2/workflow_approval_templates/10/",
+        "source_workflow_job": "/api/v2/workflow_jobs/129/",
+        "workflow_approval_template": "/api/v2/workflow_approval_templates/10/",
+        "approve": "/api/v2/workflow_approvals/130/approve/",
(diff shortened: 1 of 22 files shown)


akus062381 (commented):
I'm thinking we can add a component test for pagination....?

mabashian:
@akus062381 that's a great idea - talking with @vidyanambiar I think that test coverage would be better suited on the framework component itself rather than in each individual list.  We're already confirming that the endpoint is properly configured because the list loads.  If we added a component test that ensured that `?page=` was properly added to the payload I think that would be sufficient but I don't think we need to block this PR for that specific use case.  We can define a task to add component test coverage to that framework component.

vidyanambiar (approved):

vishsanghishetty (approved):
🚀

PR #1125, opened 2023-10-25
Replace PageForm with HubPageForm

nixocio:
Replace PageForm with HubPageForm

See:

https://github.com/ansible/ansible-ui/blob/main/framework/PageForm/README.md


ShaiahWren (approved):
Much improved error handling, this looks good.

PR #1131, opened 2023-10-26
Dark theme scrollable support

jamestalton:


keithjgrant (approved):
We need to discuss styling with the team at large and settle on styled-components vs stylesheets. I know the Controller team mostly prefers the former

PR #1134, opened 2023-10-26
[AAP-15657] - Update eda details routes. Action menu only in details pages.

lgalis:
Update eda details routes. 
Action menu only in details pages for User and Rulebook activations:

Screenshot from 2023-10-26 16-44-00
Screenshot from 2023-10-26 16-43-53
Screenshot from 2023-10-26 16-40-11
Screenshot from 2023-10-27 10-42-44


tiyiprh (approved):
Looks great! TY

PR #1136, opened 2023-10-26
Restructure oauth options to allow duplicate types

keithjgrant:
Instead of an object keyed by the oAuth type, this restructures the auth options to an array of objects where `type` is an item in the object. This way two different auth options can use the same type

This also adds a flag for hiding the username/password field in cases when only oAuth login should be allowed


vidyanambiar (approved):
lgtm!

jamestalton (approved):

PR #1137, opened 2023-10-27
[AAP-11584] Update enable/disable rulebook activation interactions

lgalis:
Update enable/disable rulebook activation interactions:

The following applies to both single rulebook activation action as wel as bulk:

Enable will not display a confirmation modal.
Disable will display a confirmation modal 
Screenshot from 2023-10-30 15-49-49
Screenshot from 2023-10-30 15-49-35
Screenshot from 2023-10-30 15-49-14
Screenshot from 2023-10-30 15-48-49


tiyiprh (approved):
LGTM!

PR #1140, opened 2023-10-30
Rename AwxError to PageErrorState and move to framework

nixocio:
Rename AwxError to PageErrorState and move to framework


nixocio:
> The framework should not import anything from the frontend.
> 
> ```
> import { RequestError } from '../../frontend/common/crud/RequestError';
> ```

Based on that, every project should create its own version of the AwxError, instead of importing that one.  Closing this PR.

MilanPospisil:
> The framework should not import anything from the frontend.
> 
> ```
> import { RequestError } from '../../frontend/common/crud/RequestError';
> ```

I see common folder in frontend problematic, because sometimes we need to use those common things in framework itself. So we should move those things like RequestError into the framework.

PR #1151, opened 2023-10-31
Fix remove second PageLayout - each page had two PageLayouts

jamestalton:


keithjgrant (approved):
Are we sure there aren't any pages now that are missing the PageLayout wrapper?

jamestalton:
> Are we sure there aren't any pages now that are missing the PageLayout wrapper?

Not 100% sure. Every page should be. I've run though a bunch manually just to do a visual check.

PR #1158, opened 2023-10-31
Monaco Webpack - Setup Monaco using getWorkerUrl

jamestalton:
Following WebPacks Example for setup
https://github.com/microsoft/monaco-editor/blob/main/samples/browser-esm-webpack

Uses getWorkerUrl vs getWorker and lets Monaco handle creating the workers.

This may fix some of the issues we are seeing in the Cypress E2E tests.

### DIFF
--- a/cypress.base.config.ts
+++ b/cypress.base.config.ts
@@ -1,4 +1,3 @@
-import codeCoverage from '@cypress/code-coverage/task';
 import pkg from 'webpack';
 import env from './webpack/environment.cjs';
 const { DefinePlugin } = pkg;
@@ -13,7 +12,7 @@ export const baseConfig: Cypress.ConfigOptions = {
     testIsolation: false,
     setupNodeEvents(on, config) {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-var-requires
-      codeCoverage(on, config);
+      // codeCoverage(on, config);
       return config;
     },
     retries: { runMode: 2, openMode: 0 },
@@ -23,7 +22,7 @@ export const baseConfig: Cypress.ConfigOptions = {
     setupNodeEvents(on, config) {
       // implement node event listeners here
       // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-var-requires
-      codeCoverage(on, config);
+      // codeCoverage(on, config);
       return config;
     },
     retries: { runMode: 2, openMode: 0 },
--- a/cypress.base.config.ts
+++ b/cypress.base.config.ts
@@ -11,8 +11,6 @@ export const baseConfig: Cypress.ConfigOptions = {
   e2e: {
     testIsolation: false,
     setupNodeEvents(on, config) {
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-var-requires
-      // codeCoverage(on, config);
       return config;
     },
     retries: { runMode: 2, openMode: 0 },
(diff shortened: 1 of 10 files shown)


vishsanghishetty (approved):

cypress.base.config.ts
```diff
@@ -13,7 +12,7 @@ export const baseConfig: Cypress.ConfigOptions = {
     testIsolation: false,
     setupNodeEvents(on, config) {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-var-requires
-      codeCoverage(on, config);
+      // codeCoverage(on, config);
```
AlexSCorey, line 15: Lets remove this code all together rather than commenting out.

cypress.base.config.ts
```diff
@@ -23,7 +22,7 @@ export const baseConfig: Cypress.ConfigOptions = {
     setupNodeEvents(on, config) {
       // implement node event listeners here
       // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-var-requires
-      codeCoverage(on, config);
+      // codeCoverage(on, config);
```
AlexSCorey, line 25: Lets remove this code all together rather than commenting out.

cypress/support/e2e.ts
```diff
@@ -18,3 +18,10 @@ before(function () {
 // HUB Component Port: 4202
 // EDA E2E Port: 4103
 // EDA Component Port: 4203
+
+Cypress.on('uncaught:exception', (_err, _runnable) => {
```
AlexSCorey, line 22: Will this prevent all uncaught excepts from causing test failures?

AlexSCorey (changes requested):

cypress/support/e2e.ts
jamestalton: This will not catch excepts from tests.

jamestalton: Needed this for the cypress **component** tests to work.
I was unable to get the cypress component test webpack config to build the webworkers.
This ignores the error that it cannot load them, and uses the inline support for handling the json and yaml that monaco has.

cypress.base.config.ts
jamestalton: done

cypress.base.config.ts
jamestalton: done

webpack/webpack.awx.cjs
```diff
@@ -5,7 +5,12 @@ const proxyUrl = new URL(AWX_SERVER);
 module.exports = function (env, argv) {
   const config = webpackConfig(env, argv);
 
-  config.entry = './frontend/awx/Awx.tsx';
+  config.entry = {
+    app: './frontend/awx/Awx.tsx',
+    'editor.worker': 'monaco-editor/esm/vs/editor/editor.worker',
+    'json.worker': 'monaco-editor/esm/vs/language/json/json.worker',
+    'yaml.worker': 'monaco-yaml/yaml.worker',
```
marshmalien, line 12: 👍

marshmalien (approved):

AlexSCorey (approved):
Looks good. I'm going to approve, but I do think we need more than 3 test runs on this PR.  Can we launch that workflow that fires of multiple test runs?  IIRC the failures that caused the work were flaky so we should exercise these changes to be sure that fix it.

jamestalton:
Screenshot 2023-11-01 at 5 34 18 PM

Last run failed on the "deletes a job from the jobs list row" which is a known flaky test and is being discussed in the Ansible-UI channel.

PR #1161, opened 2023-11-01
[Analytics] Add error state to Analytics builder

ZitaNemeckova:
## Error state:
Before:
Screenshot 2023-11-01 at 14 16 19

After:
Screenshot 2023-11-01 at 14 07 53

## Loading state:
Screenshot 2023-11-01 at 14 39 52

## PageHeader:
Screenshot 2023-11-01 at 15 14 25


h-kataria:
@MilanPospisil can you please review

MilanPospisil:
So far, good quick fix. But this will be refactored in the future according to my issue https://issues.redhat.com/browse/AAP-15074. I will create also hooks usePost, that will return { error, data, isLoading }, so we can handling those data loading the same over the whole platform. But for now, lets merge this.

LGTM :+1:

MilanPospisil (approved):

PR #1167, opened 2023-11-01
Adds empty state to workflow visualizer

AlexSCorey:
This addresses 16902 and adds Read Only empty state to workflow visualizer.


marshmalien (approved):
🎉

PR #1170, opened 2023-11-01
Add run status indicators to workflow visualizer edges 

marshmalien:
https://issues.redhat.com/browse/AAP-16995
https://issues.redhat.com/browse/AAP-16996
https://issues.redhat.com/browse/AAP-16997

Screenshot 2023-11-01 at 4 26 44 PM

### DIFF
--- a/frontend/awx/resources/templates/WorkflowVisualizer/types.ts
+++ b/frontend/awx/resources/templates/WorkflowVisualizer/types.ts
@@ -1,4 +1,4 @@
-import { NodeModel } from '@patternfly/react-topology';
+import { NodeModel, GraphElement, ElementModel, NodeStatus } from '@patternfly/react-topology';
 import { WorkflowNode } from '../../../interfaces/WorkflowNode';
 
 export type GraphNode = Omit<NodeModel, 'data'> & {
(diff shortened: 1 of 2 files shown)


frontend/awx/resources/templates/WorkflowVisualizer/types.ts
```diff
@@ -8,3 +8,14 @@ export type GraphNode = Omit<NodeModel, 'data'> & {
 export type LayoutNode = WorkflowNode & {
   runAfterTasks?: string[];
 };
+
+export interface CustomEdgeProps {
+  element: GraphElement<
+    ElementModel,
+    {
+      tag: string;
```
AlexSCorey, line 16: Does topology support the tag being a react element?  If so, perhaps we could pass it a context menu instead of just a string.

AlexSCorey (approved):
I think this looks good.

frontend/awx/resources/templates/WorkflowVisualizer/types.ts
marshmalien: It currently doesn't support `ReactNode`. I brought this up with the PatternFly team and they will look into making this prop a `ReactNode` instead of a `string`.

PR #1173, opened 2023-11-02
Move AWX routes into subdirectory

keithjgrant:
* Fixes a few rogue `*` in some routes
* Moves all awx route hooks into their own subdirectory to unclutter the awx directory
* removes unneeded "get" from all the route hook names


kialam (approved):
⭐

PR #1175, opened 2023-11-02
Renaming UserDetails base component and its usages

vidyanambiar:
PR to adopt the convention recommended by Keith:

Renames the "base" component `UserDetailsBase` to `UserDetails`, and then each platform component's specific wrapper `AwxUserDetails`, `EdaUserDetails` etc.


lgalis (approved):
LGTM

PR #1179, opened 2023-11-02
Update css to avoid line to wrap text on MultiSelectDialog

nixocio:
Update css to avoid line to wrap text on MultiSelectDialog

With this change

image


marshmalien (approved):
😎

PR #1186, opened 2023-11-03
[AAP-17462] Display error on Create activation form

lgalis:
Add eda error adapter:
Screenshot from 2023-11-09 11-04-14
Screenshot from 2023-11-09 10-48-24

### DIFF
--- a/framework/PageForm/README.md
+++ b/framework/PageForm/README.md
@@ -16,13 +16,15 @@ If no adapter is provided, the `PageForm` uses a generic adapter.
 
 ### Error Adapters
 
-- **AWX Error Adapter:** Tailored for AWX-specific errors structures.
+- **AWX Error Adapter:** Tailored for AWX-specific error structures.
 - **HUB Error Adapter:** Tailored for HUB-specific error structures.
+- **EDA Error Adapter:** Tailored for EDA-specific error structures.
 - **Generic Error Adapter:** A generic adapter that can be used as a fallback.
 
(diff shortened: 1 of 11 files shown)


nixocio:
@lgalis Do we want to show the object with the error at form level or just the error itself?

Just ```No controller token specified```

For hub we did like this 

image

lgalis:
@nixocio - I agree that on the message should be displayed - the problem is that at the moment, the backend return is {"errors":"{'errors': 'No controller token specified'}"} - this will be fixed in the backend to instead be just {'errors': 'No controller token specified'}, and then the UI will display only the message. The UI would stay as it is in this PR.

jamestalton:
The framework supports showing an error on a specific field if the backend returns the field name and error as part of the payload.

```
{
  "name": [
    "A project with this name already exists."
  ]
}
```

Screenshot 2023-11-08 at 11 52 24 AM

Would it make sense for the backend to return the error on the input field?

lgalis:
@jamestalton - the controller token in the image above is not an input field error. I am waiting for a backend PR to change the return to mark the error as such, then I will update this PR to display the error value only.

nixocio (dismissed):

lgalis:
Updated after the latest API changes to return: 
field errors:
{
        name: ['Name is required'],
        email: ['Email is invalid'],
}

non field errors:
{"non_field_errors":["No controller token specified"]}

nixocio (approved):
@lgalis Good job.

PR #1190, opened 2023-11-07
upgrade packages and remove unused packages

jamestalton:
Upgrades minor versions of packages dependencies.
Also removes some unused dependencies.

At this moment, after this, the only poackages that are still out of date are

 p-limit              4.0.0  →   5.0.0
 react-i18next      11.18.6  →  13.3.1
 styled-components   5.3.11  →   6.1.1
 
 which are all major version upgrades and should be tested before merge


nixocio (approved):
If tests are green. 🚢

PR #1198, opened 2023-11-07
[AAP-17392] Update remote and remote registry to use PageHiddenInput

nixocio:
For remote certain fields are hidden fields - once they are set the API returns `hidden_fields` only, and not the value of those fields.

For remote registry, similar concept is also present, but the API returns `write_only_fields`.

image


jamestalton:
**PageHiddenInput**
Is this only for forms? If so, I have tried to follow the standard `PageFormXXX` for the component names.
We already have a `PageFormHidden` - A component that conditionally renders its children based on the value of a watched form field.
What about renaming it to `PageFormSecret` and moving it to the page form inputs directory?

jamestalton (approved):

PR #1202, opened 2023-11-08
[don't merge yet] Fix Controller downstream test failures

akus062381:
Controller downstream tests have 19 failures currently. this PR is meant to fix those.

### DIFF
--- a/cypress/e2e/awx/access/teams.cy.ts
+++ b/cypress/e2e/awx/access/teams.cy.ts
@@ -27,10 +27,10 @@ describe('teams', function () {
     });
   });
 
-  this.afterEach(function () {
+  afterEach(function () {
     cy.deleteAwxUser(user1, { failOnStatusCode: false });
     cy.deleteAwxUser(user2, { failOnStatusCode: false });
-    cy.deleteAwxTeam(team, { failOnStatusCode: false });
+    // cy.deleteAwxTeam(team, { failOnStatusCode: false });
   });
 
   it('can render the teams list page', function () {
--- a/cypress/e2e/awx/views/jobs.cy.ts
+++ b/cypress/e2e/awx/views/jobs.cy.ts
@@ -114,7 +114,7 @@ describe('jobs', () => {
 describe('job delete', () => {
   let inventory: Inventory;
   let jobTemplate: JobTemplate;
-  let jobList: UnifiedJobList;
+  // let jobList: Unified/JobList;
 
   before(() => {
     cy.awxLogin();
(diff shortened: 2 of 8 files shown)


vishsanghishetty (approved):

cypress/e2e/awx/access/teams.cy.ts
```diff
@@ -27,10 +27,10 @@ describe('teams', function () {
     });
   });
 
-  this.afterEach(function () {
+  afterEach(function () {
     cy.deleteAwxUser(user1, { failOnStatusCode: false });
     cy.deleteAwxUser(user2, { failOnStatusCode: false });
-    cy.deleteAwxTeam(team, { failOnStatusCode: false });
+    // cy.deleteAwxTeam(team, { failOnStatusCode: false });
```
vidyanambiar, line 33: What is the reason for commenting this out? This could clog the test server with undeleted teams.

cypress/e2e/awx/views/jobs.cy.ts
```diff
@@ -133,18 +133,18 @@ describe('job delete', () => {
         jobTemplate = jt;
 
         // Launch job to populate jobs list
-        cy.awxRequestPost(`/api/v2/job_templates/${jobTemplate.id.toString()}/launch/`, {}).then(
-          (jl) => {
-            jobList = jl;
-          }
-        );
+        // cy.awxRequestPost(`/api/v2/job_templates/${jobTemplate.id.toString()}/launch/`, {}).then(
```
vidyanambiar, line 136: Why are we commenting out the launch? This will cause the[ test for relaunching a job](https://github.com/ansible/ansible-ui/blob/047379b537534d60c92bcee520aea74cb387230a/cypress/e2e/awx/views/jobs.cy.ts#L58) to fail.

cypress/support/e2e.ts
```diff
@@ -5,8 +5,28 @@ import './commands';
 import { createGlobalOrganization, createGlobalProject } from './global-project';
 
 before(function () {
+  let splitLocalhost;
   const devBaseUrlPort = Cypress.config().baseUrl?.split(':').slice(-1).toString();
-  if (devBaseUrlPort === '4101') {
+  const baseUrl = Cypress.config().baseUrl;
+  let localhost;
+  // Assuming baseUrl = "https://localhost.com"
```
vidyanambiar, line 12: Could you add a brief comment here to explain this code block please?

cypress/e2e/awx/access/teams.cy.ts
akus062381: @vidyanambiar, the reason we commented it out is because a failure is occurring downstream. We need to see whether this fixes it. We had trouble recreating the failure locally, but this is what makes the most sense based on what we were looking at.

cypress/e2e/awx/views/jobs.cy.ts
akus062381: Upon review of the code, this launch is only called in the afterEach block of this spec file. It is superfluous.

cypress/support/e2e.ts
akus062381: @vidyanambiar I am in the process of changing this code block. we may not need it. pending checks.

cypress/e2e/awx/views/jobs.cy.ts
jamestalton: Isn't this in the `beforeEach`?

cypress/support/e2e.ts
```diff
@@ -5,8 +5,30 @@ import './commands';
...
+  //   }
+  // } else {
+  //   cy.log('Base URL is not defined.');
+  // }
+  // if (localhost) {
+  //   splitLocalhost = localhost.split('/').slice(-1).toString();
+  // }
+
+  // if (devBaseUrlPort === '4101' || splitLocalhost !== 'localhost') {
+  if (Cypress.env('AWX_SERVER')) {
```
jamestalton, line 30: I think AWX_SERVER is always defined that this code is trying to be run always.
That would fail the EDA E2E tests.

PR #1203, opened 2023-11-08
[AA-1834] Reports list view toolbar

skochay330:


ZitaNemeckova:
Looks good 👍  I found few minor problems:

There should be an Empty state with `Clear all filters` when there's no results like when you select `Migration` tag
Screenshot 2023-11-14 at 12 33 06

There's nothing happening when you click on the cross inside the selected tag.

ZitaNemeckova (approved):
LGTM :+1: I'm ok with two mentioned issues being fixed in follow-up PR. I'm leaving that decision up to Salma.

skochay330:
Broken filter chip 'x' is a larger component issue (not related to this PR). Empty state will be addressed in a follow up PR.

PR #1207, opened 2023-11-09
[AAH-2494] Update edit remote test

nixocio:
Update edit remote test to delete the edited remote at the end of the test.


nixocio:
Looking at the test failure, locally I can only see the failure in headless mode.

vishsanghishetty (approved):
🎸

PR #1209, opened 2023-11-09
[AAP-18147] Display the status_message as an alert for a rulebook activation Failed status 

lgalis:
Display the status_message as an alert for a rulebook activation Failed status and as a regular field in the rulebook activation details page for every status other than Error or Failed

Screenshot from 2023-11-09 17-13-05
Screenshot from 2023-11-09 17-12-56


tiyiprh (approved):
Looks good!

PR #1217, opened 2023-11-13
docs tab for Collection Detail

ShaiahWren:
Draft pr for the Documentation tab on Collection Detail. 

Screenshot 2023-11-13 at 1 40 39 PM

### DIFF
--- a/frontend/hub/collections/CollectionDetails.tsx
+++ b/frontend/hub/collections/CollectionDetails.tsx
@@ -374,7 +374,7 @@ function CollectionDocumentationTab(props: { collection?: CollectionVersionSearc
       collection?.collection_version?.namespace || ''
     }&name=${collection?.collection_version?.name || ''}&version=${
       collection?.collection_version?.version || ''
-    }&offset=0&limit=10`
+    }&offset=0&limit=1`
   );
 
   const readme = useMemo(() => {
--- a/frontend/hub/collections/CollectionDetails.tsx
+++ b/frontend/hub/collections/CollectionDetails.tsx
@@ -61,7 +61,7 @@ import { AwxError } from '../../awx/common/AwxError';
 import { StatusCell } from '../../common/Status';
 import { useGetRequest, useGet } from '../../common/crud/useGet';
 import { HubRoute } from '../HubRoutes';
-import { hubAPI, pulpAPI } from '../api/utils';
+import { hubAPI, pulpAPI } from '../api/formatPath';
 import { HubItemsResponse } from '../useHubView';
 import { PageSingleSelect } from './../../../framework/PageInputs/PageSingleSelect';
 import { CollectionVersionSearch } from './Collection';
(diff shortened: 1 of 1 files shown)


frontend/hub/collections/CollectionDetails.tsx
```diff
@@ -368,16 +369,24 @@ function CollectionDocumentationTab(props: { collection?: CollectionVersionSearc
...
-  const { data } = useGet<CollectionDocs>(
-    hubAPI`/_ui/v1/repo/published/${collection?.collection_version?.name || ''}/${
-      collection?.collection_version?.name || ''
-    }/?include_related=my_permissions`
+  const { data } = useGet<CollectionVersionListResponse>(
+    pulpAPI`/content/ansible/collection_versions/?namespace=${
+      collection?.collection_version?.namespace || ''
+    }&name=${collection?.collection_version?.name || ''}&version=${
+      collection?.collection_version?.version || ''
+    }&offset=0&limit=10`
```
himdel, line 377: might want to use `limit=1` if we're only looking at `data.results[0]`

frontend/hub/collections/CollectionDetails.tsx
```diff
@@ -368,16 +369,24 @@ function CollectionDocumentationTab(props: { collection?: CollectionVersionSearc
 
   const [content, setContent] = useState<IContents>();
 
-  const { data } = useGet<CollectionDocs>(
-    hubAPI`/_ui/v1/repo/published/${collection?.collection_version?.name || ''}/${
-      collection?.collection_version?.name || ''
-    }/?include_related=my_permissions`
+  const { data } = useGet<CollectionVersionListResponse>(
```
nixocio, line 372: We could add error, and isLoading here.

MilanPospisil: You can look here https://issues.redhat.com/browse/AAP-15074. This is the way we want to handle errors using useGet.

frontend/hub/collections/CollectionDetails.tsx
```diff
@@ -368,16 +369,24 @@ function CollectionDocumentationTab(props: { collection?: CollectionVersionSearc
...
+    pulpAPI`/content/ansible/collection_versions/?namespace=${
+      collection?.collection_version?.namespace || ''
+    }&name=${collection?.collection_version?.name || ''}&version=${
+      collection?.collection_version?.version || ''
+    }&offset=0&limit=1`
   );
 
+  const readme = useMemo(() => {
```
MilanPospisil, line 380: @himdel Do you think this useMemo is necessary? The data comes from useGet, which memoizes the data anyway, so direct reference to the readme blob should be always the same.

himdel: oh, good catch, but moot point, `readme` itself is unused :) we can remove the whole block

MilanPospisil:
This has to be rebased, because of changes in collection tabs. Maybe it will be faster to create new PR and copy changes from this PR rather than rebase it.

ShaiahWren:
Replacing with new pr for docs tab.

PR #1218, opened 2023-11-13
Update AWX api calls to use the awxAPI util

keithjgrant:
Update all AWX API calls to use the `awxAPI` url utility


keithjgrant:
I've added a fallback value to awxAPI ... this is needed for the time being since tests use `getJobsAPIUrl()` during the jobs cleanup.

PR #1226, opened 2023-11-14
[AAH-2532] Add sync per row for remote registry

nixocio:
Add sync per row for remote registry.

Also, modify task system to the return the task response.

After

image

Before

image


skochay330:
@nixocio Can you please add before/after screenshots? Also is it possible to add sync as a bulk action as well?

nixocio:
> @nixocio Can you please add before/after screenshots? Also is it possible to add sync as a bulk action as well?

Update @skochay330,  bulk action will be addressed in a follow up, since requires a different UX.

PR #1229, opened 2023-11-15
[AAP-18277] status message as extended column

lgalis:
Screenshot from 2023-11-15 14-30-07

Screenshot from 2023-11-15 18-04-00


vishsanghishetty (approved):
🚀

jamestalton (approved):

PR #1231, opened 2023-11-16
Fix DataEditor to support two editors on the same page (credential-type)

jamestalton:
The data editor was referencing the the first editor model using
```
monaco.editor.getModels()[0]
```

When there were two editors on the page, this was causing errors and highlighting based on the first editor.
This fixes that issue.

This was found because the tests on this page were marked Flaky by Cypress.
Hopefully this resolves that.

Screenshot 2023-11-16 at 10 21 37 AM


vishsanghishetty (approved):
🚀

PR #1237, opened 2023-11-16
Upgrade react-i18next to v13

keithjgrant:
- Upgrades react-i18next to the latest version (13.5)
- Adds i18next as a direct dependency (types now need to be imported from here rather than the react-i18next wrapper)


vidyanambiar (approved):
lgtm!

jamestalton (approved):

PR #1238, opened 2023-11-16
Remove mobx-react

marshmalien:
* Remove mobx-react dependency. React-topology exports `observer` and `action` from mobx and mobx-react.


AlexSCorey (approved):
Looks good.

PR #1260, opened 2023-11-23
Repair delete collection from repository

MilanPospisil:
Loads all collection versions that needs to be deleted.

However, there can be hundreds of collections and API can return only 100 in one call. So we make repeated calls that can load 300 of them at max. When there are more, it will throw this exception and notify user:

image

### DIFF
--- a/frontend/hub/api/request.ts
+++ b/frontend/hub/api/request.ts
@@ -2,6 +2,7 @@ import { createRequestError } from '../../common/crud/RequestError';
 import { getCookie } from '../../common/crud/cookie';
 import { TaskResponse } from '../tasks/Task';
 import { parseTaskResponse } from './utils';
+import { requestGet } from '../../common/crud/Data';
 
 interface Options {
   method: string;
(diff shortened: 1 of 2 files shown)


frontend/hub/api/request.ts
```diff
@@ -179,3 +180,26 @@ export async function getHubRequest<ResponseBody extends object | TaskResponse>(
...
+  settings: {
+    getData: (res: ResponseBody) => ResponseItem[] | undefined;
+    getNextUrl: (res: ResponseBody) => string | undefined;
+  }
+) {
+  let actualUrl = url;
+  let results: ResponseItem[] = [];
+
+  while (actualUrl) {
+    const data = await requestGet<ResponseBody>(actualUrl);
```
nixocio, line 198: We could use `hubRequestCommon` it is more suitable for HUB API peculiarities.

Could we limit the amount of data fetched to avoid an infinite loop?

Could we pass a `signal?: AbortSignal` to abort the cancel the requests in case user navigates away from the page, and so on?

Also, I looked at the https://github.com/ansible/ansible-ui/blob/10633ccb1fcb33179539bef92b2baa5f87b2f7d9/frontend/awx/common/useAwxGetAllPages.tsx#L2 Perhaps we could leverage useSWRInfinite to deal with pagination here as well.

MilanPospisil: Yeah, I was thinking about some watchdog. However, we are waiting for the API to complete this as single Task. But we dont know when (and if) will be completed...

nixocio (approved):

PR #1263, opened 2023-11-24
[AAP-15696] Add default timeout to failure toast for job relaunch

prat98:
This PR is regarding [AAP-15696](https://issues.redhat.com/browse/AAP-15696), it fixes the issue with the failure toast not fading away by adding a timeout.

### DIFF
--- a/frontend/awx/views/jobs/hooks/useRelaunchJob.tsx
+++ b/frontend/awx/views/jobs/hooks/useRelaunchJob.tsx
@@ -90,7 +90,7 @@ export function useRelaunchJob(jobRelaunchParams?: JobRelaunch) {
         variant: 'danger',
         title: t('Failed to relaunch job'),
         children: error instanceof Error && error.message,
-        timeout: true,
+        timeout: 2000,
       });
     }
   };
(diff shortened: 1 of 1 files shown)


frontend/awx/views/jobs/hooks/useRelaunchJob.tsx
```diff
@@ -90,6 +90,7 @@ export function useRelaunchJob(jobRelaunchParams?: JobRelaunch) {
         variant: 'danger',
         title: t('Failed to relaunch job'),
         children: error instanceof Error && error.message,
+        timeout: true,
```
AlexSCorey, line 93: This should be a number representing the miliseconds that the toast should be visible on the screen.

prat98: So I saw setting the timeout to `true` default the timeout to 8000ms. Do we have a standard timeout duration for toasts?

AlexSCorey: Did was not aware of an default timeout for the toast, though I do think that 8000ms is too long.  A user could potentially relaunch several jobs at once, or in succession and their screen could quickly get filled with toasts if they all fail.  Lets go with a shorter timeout.  2000 ms seems comfortable to me.

AlexSCorey (approved):
Looks good to me.

PR #1265, opened 2023-11-27
Update eslint lint rule

nixocio:
See: https://eslint.org/docs/latest/rules/eqeqeq


vidyanambiar (approved):
👍 Thanks for this, @nixocio!

PR #1266, opened 2023-11-27
Update navigation for AWX, EDA, and HUB

jamestalton:
Simplifies the navigation. Needed for platform.


frontend/awx/AwxMain.tsx
```diff
@@ -30,6 +30,7 @@ export default function AwxMain() {
       }
       masthead={<AwxMasthead />}
       navigation={navigation}
+      basename={process.env.AWX_ROUTE_PREFIX}
```
nixocio, line 33: Do we need a fallback here?

jamestalton: should not need a fallback. if empty it will just not a a basename which is the desired functionality

vidyanambiar (approved):

PR #1269, opened 2023-11-27
use approve and sign wording when autoSign is true

ShaiahWren:
Hello @MilanPospisil ,

This pr is ancillary to your good work on the approve and sign feature and simply adds some ux consistency to the button, which should appear as "Approve and sign collections" if `autoSign` flag is true and just "Approve collections" if otherwise. Of course, this change applies with and without using the bulk action dialog.

Before:
Screenshot 2023-11-27 at 2 45 32 PM

After:
Screenshot 2023-11-27 at 2 41 43 PM


MilanPospisil (approved):
LGTM :+1:

PR #1276, opened 2023-11-28
add conditional to disable approve button

ShaiahWren:
Hello @MilanPospisil,

Here is another quick fix to help with the Approvals buttons. This pr adds the conditional that the "Approve and sign" thumb button is enabled when a collection is in the `staging` or `rejected` repository, but not otherwise. This prevents the user from trying to approve collections that are already approved. 

Note in the "After" image that the thumbs up button on an approved collection has been greyed out.

Before:
Screenshot 2023-11-28 at 3 25 50 PM

After:
Screenshot 2023-11-28 at 3 25 21 PM

### DIFF
--- a/frontend/hub/approvals/hooks/useApprovalActions.tsx
+++ b/frontend/hub/approvals/hooks/useApprovalActions.tsx
@@ -57,7 +57,8 @@ export function useApprovalActions(callback?: (collections: CollectionVersionSea
         onClick: (collection) => approveCollections([collection]),
         isDanger: false,
         isDisabled: (collection) =>
-          can_upload_signatures && require_upload_signatures && collection.is_signed
+          (can_upload_signatures && require_upload_signatures && collection.is_signed) ||
+          (collection.repository?.name !== 'staging' && collection?.repository?.name !== 'rejected')
             ? t`You do not have rights to this operation`
             : undefined,
       },
--- a/frontend/hub/approvals/hooks/useApprovalActions.tsx
+++ b/frontend/hub/approvals/hooks/useApprovalActions.tsx
@@ -58,8 +58,7 @@ export function useApprovalActions(callback?: (collections: CollectionVersionSea
         isDanger: false,
         isDisabled: (collection) =>
           (can_upload_signatures && require_upload_signatures && collection.is_signed) ||
-          (collection?.repository?.name !== 'rejected' &&
-            collection?.repository?.pulp_labels?.pipeline !== 'staging')
+          collection?.repository?.pulp_labels?.pipeline === 'approved'
             ? t`You do not have rights to this operation`
             : undefined,
       },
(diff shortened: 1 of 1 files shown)


nixocio (approved):
@ShaiahWren good job. Should we have similar logic to the reject button?

frontend/hub/approvals/hooks/useApprovalActions.tsx
```diff
@@ -57,7 +57,8 @@ export function useApprovalActions(callback?: (collections: CollectionVersionSea
         onClick: (collection) => approveCollections([collection]),
         isDanger: false,
         isDisabled: (collection) =>
-          can_upload_signatures && require_upload_signatures && collection.is_signed
+          (can_upload_signatures && require_upload_signatures && collection.is_signed) ||
+          (collection.repository?.name !== 'staging' && collection?.repository?.name !== 'rejected')
```
himdel, line 61: for staging, this is not correct, we can't use the names, it has to be the pipeline label

anything with `pipeline: staging` is also a staging repo

(there is no `pipeline: rejected`, so rejected is fine by name)

ShaiahWren: This should be cleaned up now. :)

frontend/hub/approvals/hooks/useApprovalActions.tsx
```diff
@@ -57,7 +57,8 @@ export function useApprovalActions(callback?: (collections: CollectionVersionSea
         onClick: (collection) => approveCollections([collection]),
         isDanger: false,
         isDisabled: (collection) =>
-          can_upload_signatures && require_upload_signatures && collection.is_signed
+          (can_upload_signatures && require_upload_signatures && collection.is_signed) ||
+          collection?.repository?.pulp_labels?.pipeline === 'approved'
             ? t`You do not have rights to this operation`
             : undefined,
```
himdel, line 63: Looks like we may also want to separate the messages, not sure if you want to do it now...

Something like..

```
       (can_upload_signatures && require_upload_signatures && collection.is_signed) ? t`...rights...` :
       collection?.repository?.pulp_labels?.pipeline === 'approved' ? t`...already there...` : undefined
```

... except that `can_upload_signatures && require_upload_signatures && collection.is_signed` is also a bug we should fix...  Approval should absolutely be possible even if a collection is already signed.

It looks like it could be a misinterpretation of the insights mode check where a collection must be signed before it can be approved.


Sooo... the button should be disabled when ANY of these is true:

* pipeline is approved (`collection?.repository?.pulp_labels?.pipeline === 'approved'`)
* a signature is required but not present (`can_upload_signatures && require_upload_signatures && !collection.is_signed` (note the extra `!`))


What about

```js
  (collection?.repository?.pulp_labels?.pipeline === 'approved') ? true :
  (can_upload_signatures && require_upload_signatures && !collection.is_signed) ? t`Signature must be uploaded first` : undefined
```

frontend/hub/approvals/hooks/useApprovalActions.tsx
```diff
@@ -70,6 +71,10 @@ export function useApprovalActions(callback?: (collections: CollectionVersionSea
         label: t('Reject'),
         onClick: (collection) => rejectCollections([collection]),
         isDanger: true,
+        isDisabled: (collection) =>
+          collection?.repository?.pulp_labels?.pipeline === 'rejected'
+            ? t`You do not have rights to this operation`
```
himdel, line 76: I'd go with `true` if it can be used to disable without a message;
if not  .. this is not about rights, but more about ... this is already done.

So.... `` t`Collection is already rejected` `` or something like that?

MilanPospisil (approved):
LGTM :+1:

himdel (approved):

PR #1278, opened 2023-11-28
Fix login and logout

jamestalton:


frontend/awx/AwxLogin.tsx
```diff
@@ -25,7 +22,5 @@ export function AwxLogin() {
     });
   }
 
-  return (
-    <Login authOptions={authOptions} apiUrl="/api/login/" onLoginUrl={getPageUrl(AwxRoute.Login)} />
-  );
+  return <Login authOptions={authOptions} apiUrl="/api/login/" onLoginUrl="/dashboard" />;
```
keithjgrant, line 25: is the dashboard URL always `/dashboard` regardless of the path prefix?

jamestalton: at the moment yes, i had an issue with getPageUrl as I think the login might exist outside the navigation provider

PR #1282, opened 2023-11-29
PageDashboardCard support canCollapse

jamestalton:
Allows enabling collapsing on a page dashboard card.


vidyanambiar (commented):
Can we move the collapse icon to the start of the card (to the left of the title) rather than at the end of it? That seems like a more familiar experience for expand/collapse. Something like this? https://www.patternfly.org/components/card#expandable-cards

image

vidyanambiar (approved):

PR #1285, opened 2023-11-30
Repair approve and sign

MilanPospisil:
Repaired approve and sign in approval list - it should sign now. 

The problem was that it signs only when there is more than one approval repository and approval modal appears.
Now signing was also implemented in signing without approval modal (in case of single approved repo).

Function for approval was moved from CopyToRepositoryModal (which is used for both copy and approve action) and its shared and used also in single repo approval.


ShaiahWren (approved):
Much better, it works for me now. :)

PR #1290, opened 2023-11-30
Enhance useAwxRoles for downstream

jamestalton:
Essentially adds a resource type name to the structure.


vidyanambiar (changes requested):
Seeing an error when trying to add a role to a user/team, using the Add Roles button on the roles list. This is on the first field (dropdown for selecting the resource type). 
image
image

vidyanambiar (approved):
👍

PR #1291, opened 2023-11-30
CLI config on ns detail

ShaiahWren:
This pr add the CLI config tab to the ns details page. 

Screenshot 2023-11-30 at 5 43 52 PM

### DIFF
--- a/frontend/hub/namespaces/HubNamespacePage/HubNamespaceCLI.tsx
+++ b/frontend/hub/namespaces/HubNamespacePage/HubNamespaceCLI.tsx
@@ -1,34 +1,10 @@
-import { useParams } from 'react-router-dom';
-import React from 'react';
-import { LoadingPage } from '../../../../framework';
+import { ClipboardCopy } from '@patternfly/react-core';
 import ExternalLinkAltIcon from '@patternfly/react-icons/dist/esm/icons/external-link-alt-icon';
 import { Trans } from 'react-i18next';
-import { ClipboardCopy } from '@patternfly/react-core';
-import { useGet } from '../../../common/crud/useGet';
-import { HubItemsResponse } from '../../useHubView';
-import { HubNamespace } from '../HubNamespace';
-import { hubAPI } from '../../api/formatPath';
-import { HubError } from '../../common/HubError';
 import { getRepoURL } from '../../api/utils';
 
 export function HubNamespaceCLI() {
   const repositoryUrl = getRepoURL('published');
-  const params = useParams<{ id: string }>();
-  const {
-    data: response,
-    error,
-    refresh,
-  } = useGet<HubItemsResponse<HubNamespace>>(
-    hubAPI`/_ui/v1/namespaces/?limit=1&name=${params.id ?? ''}`
-  );
-
-  if (!response || !response.data || (response.data.length === 0 && !error)) {
-    return <LoadingPage />;
-  }
-
-  if (error) {
-    return <HubError error={error} handleRefresh={refresh} />;
-  }
 
   return (
     <div style={{ padding: '24px' }}>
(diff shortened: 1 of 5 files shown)


frontend/hub/namespaces/HubNamespacePage/HubNamespaceCLI.tsx
```diff
@@ -0,0 +1,54 @@
...
+export function HubNamespaceCLI() {
+  const repositoryUrl = getRepoURL('published');
+  const params = useParams<{ id: string }>();
+  const {
+    data: response,
+    error,
+    refresh,
+  } = useGet<HubItemsResponse<HubNamespace>>(
+    hubAPI`/_ui/v1/namespaces/?limit=1&name=${params.id ?? ''}`
+  );
```
ZitaNemeckova, line 24: Is this needed? `data` is never used to show anything to the user. If it's not needed there's no need for `const params = useParams<{ id: string }>();` either.

frontend/hub/namespaces/HubNamespacePage/HubNamespaceCLI.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  if (!response || !response.data || (response.data.length === 0 && !error)) {
+    return <LoadingPage />;
+  }
+
+  if (error) {
```
ZitaNemeckova, line 30: This is not needed if we don't load data. And this page seems to be pretty much static.

frontend/hub/namespaces/HubNamespacePage/HubNamespaceCLI.tsx
```diff
@@ -0,0 +1,54 @@
...
+    return <HubError error={error} handleRefresh={refresh} />;
+  }
+
+  return (
+    <div style={{ padding: '24px' }}>
+      <ClipboardCopy isReadOnly>{repositoryUrl}</ClipboardCopy>
```
ZitaNemeckova, line 36: It's better to use `CopyCell` from framework for styling purposes :)

frontend/hub/namespaces/HubNamespacePage/HubNamespaceCLI.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { useGet } from '../../../common/crud/useGet';
+import { HubItemsResponse } from '../../useHubView';
+import { HubNamespace } from '../HubNamespace';
+import { hubAPI } from '../../api/formatPath';
+import { HubError } from '../../common/HubError';
+import { getRepoURL } from '../../api/utils';
+import styled from 'styled-components';
+
+export function HubNamespaceCLI() {
+  const repositoryUrl = getRepoURL('published');
```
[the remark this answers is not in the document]
ShaiahWren: If we are trying to replicate what is on the [old ui](https://github.com/ansible/ansible-hub-ui/blob/d1f9f7e309bd0b6e301e0bdec445da670f6eaa17/src/containers/namespace-detail/namespace-detail.tsx#L287), yes I believe so. :)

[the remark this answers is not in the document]
himdel: Yeaah.. right now it's the best we can do, though I think that changes after the backend namespace refactor (I recall some per-repository namespace metadata should be there).

Right now, a namespace is completely orthogonal to repositories, and can have collections from multiple repositories. And the published repo is the one repo you would never push to...(well, together with rejected and validated) 

But `getRepoURL` special-cases `published` to just return the server URL without the repository parts, making this work. (And is used from multiple places.)

---

How it should actually work..

URLs for **pulling**
* if the repository is literally `published`, return the api base path (ie `http://host/api/galaxy/`)
* for any other repository, return the distribution base path (ie `http://host/api/galaxy/content/foobar/`)

URLs for **pushing**
* if the repository is a `pipeline:approved` it has to go through the approval process, return the api base path
* if the repository is a `pipeline:staging` it has to go through the process, but the repo is a valid place to start the process, return the distribution base path
* if the repository is a `pipeline:rejected`, there is no reason to push there, return nothing (or go with api base path if it makes more sense than nothing)
* any other repository, we push there directly, use the distribution base path (but also add an object permission check for that repo before displaying it)

ZitaNemeckova (approved):
:+1:

PR #1294, opened 2023-12-01
Workflow visualizer remove all nodes and screen state management 

marshmalien:
Cleanup close button location: https://issues.redhat.com/browse/AAP-18142

Remove link: https://issues.redhat.com/browse/AAP-17003

Remove node: https://issues.redhat.com/browse/AAP-17068

delete


frontend/awx/resources/templates/WorkflowVisualizer/hooks/useRemoveAllNodes.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const nameColumn = useMemo<ITableColumn<GraphNode>>(
+    () => ({
+      header: t('Name'),
+      cell: (item: GraphNode) => <TextCell text={item.getLabel() || t('DELETED')} iconSize="sm" />,
+    }),
+    [t]
+  );
+  const idColumn = useMemo<ITableColumn<GraphNode>>(
+    () => ({
+      header: t('Identifier'),
```
AlexSCorey, line 26: The identifier becomes that same value as "alias" - which is a field in the node form - if the user adds a value there. In the case where they do not, the identifier comes from the API and may not be of much use for the user.  

To help us with this we do have a util function called `stringIsUUID`

Perhaps in this delete modal we should have a alias/idenifier column, a resource name column..... what do you think?

AlexSCorey (changes requested):
I do think we should add the remove element in this PR.  The other comments should be added in this PR, or become issues (if you agree with them) so we can track them.

frontend/awx/resources/templates/WorkflowVisualizer/components/DeletedNode.tsx
```diff
@@ -1,14 +1,25 @@
 import { FC } from 'react';
 import { TrashIcon } from '@patternfly/react-icons';
-import { DefaultNode, WithDragNodeProps } from '@patternfly/react-topology';
+import { DefaultNode, WithDragNodeProps, WithSelectionProps } from '@patternfly/react-topology';
 import type { CustomNodeProps } from '../types';
 
-export const DeletedNode: FC<CustomNodeProps & WithDragNodeProps> = ({ element, ...rest }) => {
+export const DeletedNode: FC<CustomNodeProps & WithDragNodeProps & WithSelectionProps> = ({
```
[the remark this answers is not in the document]
marshmalien: Let's do it in a future iteration

frontend/awx/resources/templates/WorkflowVisualizer/hooks/useRemoveAllNodes.tsx
AlexSCorey: @marshmalien and I spoke offline and my concerns have been addressed.  I agree with this approach.

AlexSCorey (approved):
This looks great!

PR #1299, opened 2023-12-01
Repair collection detail version list

MilanPospisil:
Adds browse modal to the collection detail version dropdown and ability to load more items.

It also adds some change into the framework - PageAsyncSingleSelect (and multi variant) now allows for Browse button that runs the modal. Previously, this was available only to ToolbarAsyncSingleSelect (and multi variant).

image

Browse modal:
image

TODO:
Filters may be in the future available also in form of typing into the dropdown and filter them on API side, not in memory over results fetched by API. This was part of the PR orginaly, but i reset the changes, so I can merge the PR now. This must be communicated first.

The dropdown menu lies over the modal.

Another TODOs may be:
- use useView hook to load items in loadmore, because the browse modal has the same data as dropdown data
- unify PageAsyncSingleSelect and PageAsyncMultiSelect - they both should use the same core component, because those two component are nearly similar and there is lots of copy paste here.

### DIFF
--- a/framework/PageInputs/PageAsyncMultiSelect.tsx
+++ b/framework/PageInputs/PageAsyncMultiSelect.tsx
@@ -5,12 +5,6 @@ import { useTranslation } from 'react-i18next';
 import { PageMultiSelect, PageMultiSelectProps } from './PageMultiSelect';
 import { PageSelectOption } from './PageSelectOption';
 
-/** A function to open a single selection browse modal for a filter. */
-type PageOpenMultiSelectBrowse<T> = (
-  onSelect: (values: T[]) => void,
-  defaultSelections?: T[]
-) => void;
-
 export interface PageAsyncMultiSelectQueryResult<ValueT> {
   total: number;
   options: PageSelectOption<ValueT>[];
@@ -31,9 +25,6 @@ export interface PageAsyncMultiSelectProps<ValueT>
 
   /** The placeholder to show if the query fails. */
   queryErrorText?: string | ((error: Error) => string);
-
-  /** The function to open the browse modal. */
-  openBrowse?: PageOpenMultiSelectBrowse<ValueT>;
 }
 
 /**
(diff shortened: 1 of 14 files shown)


framework/PageInputs/PageAsyncMultiSelect.tsx
```diff
@@ -157,6 +166,18 @@ export function PageAsyncMultiSelect<
...
+          variant="link"
+          onClick={() => {
+            props.openBrowse?.((selections) => {
+              props.onSelect(() => selections);
+            }, props.values || undefined);
+          }}
+        >
+          {t`Browse`}
+        </Button>
+      ) : undefined}
```
nixocio, line 180: @jamestalton can you review the changes related to the framework?

framework/PageInputs/PageAsyncMultiSelect.tsx
```diff
@@ -25,6 +31,9 @@ export interface PageAsyncMultiSelectProps<ValueT>
 
   /** The placeholder to show if the query fails. */
   queryErrorText?: string | ((error: Error) => string);
+
+  /** The function to open the browse modal. */
+  openBrowse?: PageOpenMultiSelectBrowse<ValueT>;
```
jamestalton, line 36: I'd like to keep the browse button logic out of the PageAsyncMultiSelect.
PageAsyncMultiSelect works fine with dynamically loading items in it's current state.
Trying to keep PageAsyncMultiSelect as simple as possible.

For any places that use PageAsyncMultiSelect and want to support a 'browse',
It can be passed in as the 'footer' prop to the PageAsyncMultiSelect.
Might be good to have a PageSelectBrowseFooter component that encapsulates that support.

Makes the components more composable.

jamestalton (changes requested):

nixocio:
@MilanPospisil I got this code locally, and I noticed that when a I clicked the browse, and the following is shown

image

Do we want the modal and search overlapping like that?

framework/PageInputs/PageAsyncMultiSelect.tsx
MilanPospisil: Done, but I choose to implement it inside the component params directly, no need to move it to shared functions, since we would need to pass half of parameters anyway (default selection, on select) only to render button :)

ZitaNemeckova:
@jamestalton Please have a look at this again. I depend on those Collections URL changes.

jamestalton (approved):

PR #1305, opened 2023-12-04
Add subtitle support in navigation items

jamestalton:
This is needed to give some context to nav items as we transition to new terminology in the navigation.
i.e. Helps the user understand the terminology change.


vidyanambiar (changes requested):
When a subtitle is added for an expandable nav item, the label shifts a little to the right. (see _Access_)
image

Could we increase the subtitle's font size by a bit?

PR #1311, opened 2023-12-04
Role descriptions support wrapping of description in column.

jamestalton:
For columns there is a "description" type.

When a column is a description, this makes it so that the description has a min width of 200px but also wraps when needed. This is useful for really long descriptions shown in the table.

This switches the roles columns for EDA and AWX to use type 'description' for the description column.


MilanPospisil (approved):
LGTM :+1:

PR #1318, opened 2023-12-06
Feature parity for Hub role creation, edit and details

vidyanambiar:
## Existing UI

image

## New UI

image

### Role details

The existing UI does not have a role details page but this was added to keep the roles UI consistent across components.

image


frontend/hub/access/roles/Roles.tsx
```diff
@@ -29,7 +28,7 @@ export function HubRolesTable() {
 
   const view = usePulpView<Role>({
     url: pulpAPI`/roles/`,
-    keyFn: nameKeyFn,
+    keyFn: (role: { pulp_href: string }) => role.pulp_href,
```
jamestalton, line 31: This line might cause issues as the keyFn needs to not change between rerenders.

nixocio (commented):
@vidyanambiar I got this code locally, and it works great. 💯 

I left a couple of minor suggestions, and questions.

frontend/hub/access/roles/hooks/useRoleActions.tsx
```diff
@@ -68,8 +70,11 @@ export function useRoleRowActions(onComplete: (roles: Role[]) => void) {
             : t(
                 'You do not have permission to edit this role. Please contact your organization administrator if there is an issue with your access.'
               ),
-        // eslint-disable-next-line @typescript-eslint/no-unused-vars
-        onClick: (role) => alert('TODO'),
+        href: (role) => {
+          return getPageUrl(HubRoute.EditRole, {
+            params: { id: parsePulpIDFromURL(role.pulp_href) ?? '' },
```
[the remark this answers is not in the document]
vidyanambiar: Thanks for catching that. Looks like role names are unique so I updated the code to use name instead.

frontend/hub/access/roles/RolePage/RoleForm.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const postRequest = usePostRequest<RoleRequestBody>();
+  const onSubmit: PageFormSubmitHandler<RoleInput> = async (values) => {
+    const { name, description, permissionCategories } = values;
+    const permissions = getSelectedPermissions(permissionCategories);
+    await postRequest(pulpAPI`/roles/`, {
+      name,
+      description,
+      permissions,
+    });
+    navigate(-1);
```
nixocio, line 46: I just noticed that one now. I believe we are redirecting to the details page after creating resources in general.

nixocio (approved):
LGTM

frontend/hub/access/roles/RolePage/RoleForm.tsx
vidyanambiar: updated 👍

PR #1320, opened 2023-12-06
[AAH-2995] Add signed only and sync dependencies for remotes

nixocio:
Update remotes to reflect remotes latest changes in HUB

Update details page, expand unit tests to cover new changes

image

### DIFF
--- a/cypress/e2e/hub/remotes.cy.ts
+++ b/cypress/e2e/hub/remotes.cy.ts
@@ -62,13 +62,44 @@ describe('Remotes', () => {
     cy.clickButton(/^Clear all filters$/);
   });
 
+  it('display alert when creating a remote with community URL and checking select `signed collections only`', () => {
+    cy.navigateTo('hub', 'remotes');
+    const remoteName = `test-remote-${randomString(5, undefined, { isLowercase: true })}`;
+    cy.get('[data-cy="create-remote"]').should('be.visible').click();
+    cy.get('[data-cy="name"]').type(remoteName);
+    cy.get('[data-cy="url"]').type(Remotes.remoteCommunityURL);
+    cy.get('[data-cy="signed_only"]').check();
+    cy.get('[data-cy="signed-only-warning"]').should('be.visible');
+    cy.contains(Remotes.showAdvancedOptions).click();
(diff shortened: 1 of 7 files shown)


frontend/hub/remotes/RemoteForm.tsx
```diff
@@ -494,18 +579,34 @@ const TranslationLabelHelp = () => {
   );
 };
 
-function RequirementsFile() {
+function RequirementsFile({ isCommunityRemote }: IRequirementsFile) {
   const { t } = useTranslation();
+  const requirementsFile = useWatch({ name: 'requirements_file' }) as string;
+  const signedOnlyInput = useWatch({ name: 'signed_only' }) as boolean;
+  const isRequired =
+    isCommunityRemote && signedOnlyInput && requirementsFile === yamlRequirementsTemplate;
```
himdel, line 587: requirements being required depends only on isCommunityRemote, but not signedOnlyInput or the file being the same as the template, right?

nixocio: Thanks @himdel I will remove the signeOnlyInput logic from here, but the same template is to avoid the template to be saved as default since it will be a required one now.

nixocio: @himdel logic was updated.

cypress/e2e/hub/remotes.cy.ts
```diff
@@ -86,16 +117,29 @@ describe('Remotes', () => {
     cy.get('[data-cy="download-concurrency"]').type(Remotes.downloadConcurrency);
     cy.get('[data-cy="rate-limit"]').type(Remotes.rateLimit);
     cy.get('[data-cy="tls_validation"]').click();
+    cy.get('[data-cy="requirements_file"]')
+      .click()
+      .focused()
+      .type('{ctrl}a')
```
himdel, line 123: this is probably platform-dependent, might not work on a mac.. would `{home}` be better?

nixocio: @himdel updated.

himdel (approved):

PR #1330, opened 2023-12-07
Update remote registry tests

nixocio:
Add test to sync remote registry, and test to explore views and pagination

image


vishsanghishetty (approved):
🚀  LGTM

PR #1334, opened 2023-12-07
Update keyboard navigation for checkbox on BukConfirmationDialog

nixocio:
Update keyboard navigation for checkbox on BukConfirmationDialog

Selection done via keyboard

image

### DIFF
--- a/framework/PageDialogs/BulkConfirmationDialog.tsx
+++ b/framework/PageDialogs/BulkConfirmationDialog.tsx
@@ -31,6 +31,10 @@ const ConfirmBoxDiv = styled.div`
   height: 64px;
   display: flex;
   align-items: center;
+
+  &:focus {
+    outline: none;
+  }
 `;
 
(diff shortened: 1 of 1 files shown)


framework/PageDialogs/BulkConfirmationDialog.tsx
```diff
@@ -31,6 +31,10 @@ const ConfirmBoxDiv = styled.div`
   height: 64px;
   display: flex;
   align-items: center;
+
+  &:focus {
+    outline: none;
```
nixocio, line 36: @jamestalton any suggestions here? I used that to avoid to show the focus on the div itself

nixocio: As per discussion with @jamestalton that the current behavior seems to be the proper one based on https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role#keyboard_interactions Closing this PR.

PR #1339, opened 2023-12-08
Update plural and singular title for bulk actions

nixocio:
Update plural and singular title for bulk actions, there were previous caseas using the length to decide about plural or singular. Standarize this one.

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -12,3 +12,7 @@
 ## Testing
 
 See: [CYPRESS.md](../cypress/CYPRESS.md)
+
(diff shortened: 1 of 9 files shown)


nixocio:
There are other places that same concept should be applied, such as ```actionButtonText``` but it might be addressed in a follow up PR.

marshmalien (approved):

nixocio:
Followed up with @jamestalton and the usage of count variable for plural should be the one used whenever possible, I will add an entry to README to deal with that one, and update this PR.

lgalis (approved):

jamestalton (approved):

PR #1342, opened 2023-12-12
Collection tab distributions

MilanPospisil:
Adds new distribution tab:

image

it needed also to add HUB_SERVER environment variable, so you need to rebuild the webpack using npm run build and webpack will replace process.env.HUB_SERVER, which will not normally work in the browser.


frontend/hub/collections/CollectionPage/CollectionDistributions.tsx
```diff
@@ -1,7 +1,88 @@
...
+        sort: 'base_path',
+      },
+      {
+        header: t('Created'),
+        type: 'datetime',
+        value: (item) => item.pulp_created,
+        sort: 'pulp_created',
+      },
+      {
+        header: t('	CLI configuration'),
```
himdel, line 72:
```suggestion
        header: t('CLI configuration'),
```

himdel (approved):

PR #1344, opened 2023-12-12
HUB add initial support for collection approvals notifications

jamestalton:


ZitaNemeckova:
@jamestalton please add a code example how to call it.

ZitaNemeckova (changes requested):
No idea why we should show collections in staging in Alerts. Needs more clarifications.

jamestalton:
Updated description for the PR and linked to UXD designs.

>  I see it calls some API. I believe Hub doesn't have API for this

This is hooked up to the Hub API and works as expected.

> Is it possible to have variant for success, danger as well?

it is, but the notifications are only for things that need action.

tiyiprh:
This work is part of the Gateway work which includes consolidating the notifications from the different areas of AAP into a consolidated notifications area, which includes collection approvals from Hub. This is slated for the 2.5 release. Notifications like other areas within AAP should be paired with the corresponding permissions. If a user does not have the permissions to approve collections then these notifications should not display for them because they cannot act on them. We are  essentially lifting the current experience in Hub into the Gateway but also adding it to a consolidated notifications area.

himdel:
No objections to the notifications mechanism/ui :). That part is common ui :+1: 

Just to using the list of unpublished collections as a source of notifications.

We'll just need to define how we use these notifications and what for,
and if we're talking about notifying users of unpublished collections,
we'll need to define which kinds of user would see this. 
Notifying users of unapproved collections would be considered a new feature.

frontend/hub/HubMasthead.tsx
```diff
@@ -91,3 +98,40 @@ export function HubMasthead() {
...
+  const { t } = useTranslation();
+  const getPageUrl = useGetPageUrl();
+
+  const { hasPermission } = useHubContext();
+
+  const canApprove = hasPermission('ansible.modify_ansible_repo_content');
+
+  const { data: result } = useGet<HubItemsResponse<CollectionVersionSearch>>(
+    canApprove ? hubAPI`/v3/plugin/ansible/search/collection-versions/` : undefined,
+    { page_size: 200, repository_label: 'pipeline=staging' },
```
himdel, line 112: (max page_size is 100, but the api will silently ignore this)

himdel (approved):

ZitaNemeckova (approved):

PR #1350, opened 2023-12-13
Collection tab Imports

MilanPospisil:


frontend/hub/collections/CollectionPage/CollectionImportLog.tsx
```diff
@@ -27,67 +28,134 @@ export function CollectionImportLog() {
...
+  const ref = useRef<HTMLDivElement>(null);
 
   const { data: collectionImport } = useGet<CollectionImport>(
     collectionImportsResponse && collectionImportsResponse.data.length
       ? hubAPI`/_ui/v1/imports/collections/${collectionImportsResponse.data[0].id}/`
       : ''
   );
+
+  const StyledArrowDown = styled.div`
+    color: white;
```
nixocio, line 40: This could be moved outside the main component definition. Avoid to re-create the component every time the main one renders, and create all custom ones in the same place.

https://styled-components.com/docs/basics#define-styled-components-outside-of-the-render-method

MilanPospisil: Definetly, I wrote this before you commented on previous PR. I was going to redo this now :-)

MilanPospisil: Definetly, I was going to redo this there as well.

frontend/hub/collections/CollectionPage/CollectionImportLog.tsx
```diff
@@ -27,67 +30,195 @@ export function CollectionImportLog() {
...
+`;
+
+function NavigationArrow(props: { direction: 'up' | 'down'; onClick: () => void }) {
+  const className = `fa fa-arrow-circle-${props.direction} clickable`;
+
+  const Component = props.direction === 'up' ? StyledArrowUp : StyledArrowDown;
+
+  return (
+    <Component>
+      <span
```
nixocio, line 208: Could we use a native button here to avoid any accessibility issues? Did I misread the intent here?

nixocio (approved):
@MilanPospisil overall good job. Is there a related issue?

I left one question.

frontend/hub/collections/CollectionPage/CollectionImportLog.tsx
MilanPospisil: Is it necessary? This is from old code, but changing this to button broke the css of element (it looks badly), which will need some effort to repair it.

MilanPospisil: I will merge it for now and we can consult it after you return from PTO, I dont want this PR to be blocked for too long.

PR #1351, opened 2023-12-13
Replace project sync with project cancel when a project is syncing

kialam:
Per convo with UXD, we now replace the sync icon with the cancel icon when a project is syncing:
project-detail-sync
project-list-sync


vishsanghishetty (approved):
🚀  LGTM

PR #1354, opened 2023-12-14
Framework - Fix page layout

jamestalton:
For our pages to support scrolling of tables and such, the main page container needs to have a max height of 100%.
The new drawer added for notifications needs this set to enable us to handle scrolling properly.


kialam (changes requested):
Can we add this as a style prop to the Drawer component directly? I'm concerned this will have unintended side effects since it's a PF class that's being modified.

kialam (approved):

PR #1367, opened 2023-12-18
Fix a couple template form credential field errors

kialam:
This PR adds the following fixes to the `credentials` field in the Template form:
1. Fix field error not showing up when trying to associate an incorrect credential type to a template.
2. Fix incorrect API call to dissociate a credential.

Before:

cred-error-before

After:
cred-error-after


nixocio (approved):
🥇

PR #1374, opened 2023-12-19
Instance forks check

kialam:
This PR fixes a TypeError thrown when trying to adjust forks capacity for an instance with 0 capacity. It also disables the slider if the user is not a superuser.

Before:
instance-forks-before

After:
istance-forks-disabled


vidyanambiar (approved):
LGTM 👍 

Likely unrelated to these changes but when I login as a normal user, the Topology view appears empty. I think in the existing UI we hide it from the main menu.
image

PR #1380, opened 2023-12-20
Shaiah edit ns

ShaiahWren:
Edit namespace detail page. 
See https://issues.redhat.com/browse/AAH-3026

Screenshot 2023-12-21 at 1 23 48 PM

### DIFF
--- a/frontend/hub/namespaces/HubNamespaceForm.tsx
+++ b/frontend/hub/namespaces/HubNamespaceForm.tsx
@@ -10,7 +10,7 @@ import {
   usePageNavigate,
 } from '../../../framework';
 import { useGet } from '../../common/crud/useGet';
-import { usePatchRequest } from '../../common/crud/usePatchRequest';
+import { usePutRequest } from '../../common/crud/usePutRequest';
 import { usePostRequest } from '../../common/crud/usePostRequest';
 import { HubRoute } from '../HubRoutes';
 import { hubAPI } from '../api/formatPath';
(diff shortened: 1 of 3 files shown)


vishsanghishetty:
@ShaiahWren looks like the translation for breadcrumbs and Namespace field (label and the placeholder) are missing

nixocio (approved):
Good job @ShaiahWren, Is there a related ticket? Please, add a reference here.

ShaiahWren:
Hi! @vishsanghishetty just wanted to clarify your comment about the translation. I left that actual name of the namespace without a translation because I didn't think it was dependent on the language in use, but if it it, I will add it.

vishsanghishetty:
> Hi! @vishsanghishetty just wanted to clarify your comment about the translation. I left that actual name of the namespace without a translation because I didn't think it was dependent on the language in use, but if it it, I will add it.

thanks for clarifying, @ShaiahWren

frontend/hub/namespaces/HubNamespaceForm.tsx
```diff
@@ -59,10 +59,10 @@ export function EditHubNamespace() {
...
     error,
     refresh,
-  } = useGet<HubNamespace>(hubAPI`/_ui/v1/namespaces/${name ?? ''}/`);
-  const patchRequest = usePatchRequest<HubNamespace, HubNamespace>();
+  } = useGet<HubNamespace>(hubAPI`/_ui/v1/my-namespaces/${name ?? ''}/`);
+  const putRequest = usePutRequest<HubNamespace, HubNamespace>();
   const onSubmit: PageFormSubmitHandler<HubNamespace> = async (namespace) => {
-    await patchRequest(hubAPI`/_ui/v1/namespaces/`, namespace);
+    await putRequest(hubAPI`/_ui/v1/my-namespaces/${name ?? ''}/`, namespace);
     navigate(-1);
```
ShaiahWren, line 66: @jamestalton @vishsanghishetty could you take a look at this use of `navigate(-1)` on the ns edit form? If you _edit namespace_ in the ui, you can see that navigate only takes you back to the previous page and does not trigger a reload, which we need for the fields to be updated. Do you know of anything we can replace this with that will trigger reload? Thank you~

jamestalton (approved):

ShaiahWren:
Following up on the navigate(-1) issue in the next pr.

PR #1390, opened 2024-01-02
Update username HubMastHead to use optional chain

nixocio:
Update username HubMastHead to use optional chain


ShaiahWren (approved):
Works for me.

PR #1397, opened 2024-01-03
hub/api: deduplicate; fix HUB_API_PREFIX example in README

himdel:
`getBaseAPIPath`: never imported, inline
`getAPIHost`: never imported, inline
`getRepoURL`: duplicate between formatPath & utils, only one has the right trailing slashes to deal with no trailing slash on `HUB_API_PREFIX`, but only the other one respects `HUB_SERVER`, fix, merge
`apiTag`: duplicate between formatPath & utils, remove from utils

README: update HUB_API_PREFIX example to match the use - remove trailing slash


ShaiahWren (approved):
LGTM

PR #1408, opened 2024-01-04
Hub - merge useHubView & usePulpView

himdel:
`useHubView` & `usePulpView` were *mostly* the same, differing in name, types, default `sortKey`, `sortKey` only being implemented in `useHubView`, and reading count, items & next from different places in the response.

The idea that sort, pagination and the shape of the response depends on hub vs pulp is not quite true, see AAH-3007 for a full table.

Thus:

* merge `useHubView` & `usePulpView` into one `useHubView`
* detect `sortKey` based on `url` (returns `sort`, `ordering` and `order_by`)
* add `pageKey` to set `offset` + `limit` vs `page` + `page_size`
* detect `pageKey` based on `url` (`offset`, `page`)
* handle both `HubItemsResponse<T>` and `PulpItemsResponse<T>` transparently (based on if it has `.meta`)

(follows #1397)

Related to AAH-3007, but does not address the list APIs used with useGet/request, only useView.


nixocio (approved):
👍

PR #1414, opened 2024-01-04
Collections tab ns page

ShaiahWren:
This pr ads the collections tab to the namespace page alongside "details" and "CLI config". It uses the same logic from the collections list view and adds `params.id` as a namespace query so collections stay relevant to the given namespace.
See https://issues.redhat.com/browse/AAH-2476

Screenshot 2024-01-04 at 11 58 12 AM


nixocio (approved):
@ShaiahWren

Good job on this implementation so far.

Is there an existing ticket related to this change?

ShaiahWren:
@nixocio just added the ticket - https://issues.redhat.com/browse/AAH-2476

PR #1430, opened 2024-01-09
Webpack: Fix hosting of static assets based on ROUTE_PREFIX.

jamestalton:
This fixes the case where the EDA_ROUTE_PREFIX was being set to '/eda' but static assets were deployed from '/'.
The default routes are all now '/' but they can still be overridden if needed.

Example for AWX hosting the UI
```
PUBLIC_PATH=/static/awx/ AWX_ROUTE_PREFIX='/ui_next' npm run build:awx
```
That is because AWX is setup to serve the files from '/static/awx/' but wants the route of the UI to be on 'ui_next'

### DIFF
--- a/webpack/webpack.awx.cjs
+++ b/webpack/webpack.awx.cjs
@@ -4,8 +4,12 @@ const { AWX_SERVER } = env;
 const proxyUrl = new URL(AWX_SERVER);
 module.exports = function (env, argv) {
   const config = webpackConfig(env, argv);
-
   config.entry = './frontend/awx/Awx.tsx';
+
+  // publicPath is the path where the bundle is served from
+  // https://webpack.js.org/guides/public-path/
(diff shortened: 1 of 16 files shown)


webpack/webpack.awx.cjs
```diff
@@ -4,9 +4,13 @@ const { AWX_SERVER } = env;
 const proxyUrl = new URL(AWX_SERVER);
 module.exports = function (env, argv) {
   const config = webpackConfig(env, argv);
-
   config.entry = './frontend/awx/Awx.tsx';
 
+  // publicPath is the path where the bundle is served from
+  // https://webpack.js.org/guides/public-path/
+  // it nees to match the route prefix of the app
```
jamesmarshall24, line 11: typo "nees" to "needs"

vishsanghishetty (approved):

PR #1454, opened 2024-01-15
Add tests to API token

ZitaNemeckova:
How the page works:
Screenshot 2024-01-15 at 19 47 25
After clicking Generate token:
Screenshot 2024-01-15 at 19 47 32

### DIFF
--- a/cypress/e2e/hub/api-tokens.cy.ts
+++ b/cypress/e2e/hub/api-tokens.cy.ts
@@ -4,8 +4,19 @@ describe('API Tokens', () => {
   before(() => {
     cy.hubLogin();
   });
-  it('it should render the api tokens page', () => {
+  it('should render the api tokens page', () => {
     cy.navigateTo('hub', APITokens.url);
     cy.verifyPageTitle(APITokens.title);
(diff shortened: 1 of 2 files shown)


cypress/e2e/hub/api-tokens.cy.ts
```diff
@@ -1,11 +1,22 @@
 import { APITokens } from './constants';
 
 describe('API Tokens', () => {
-  before(() => {
+  beforeEach(() => {
     cy.hubLogin();
   });
```
nixocio, line 6: Is there a particular reason to login before every test? If we can re-use the session that would make the tests a bit faster.

ZitaNemeckova: This may not be needed. I'm debugging on GH so it's still WIP. I want to get it green before I try to remove it :)

ZitaNemeckova: And it wasn't. Nice catch!

nixocio (approved):
🥇

PR #1455, opened 2024-01-15
Add tests to Signature keys

ZitaNemeckova:
How the page works:
Screenshot 2024-01-15 at 20 33 12


cypress/e2e/hub/signature-keys.cy.ts
```diff
@@ -1,12 +1,42 @@
...
-  it('it should render the signature keys page', () => {
+  it('should render the signature keys page', () => {
     cy.navigateTo('hub', SignatureKeys.url);
     cy.verifyPageTitle(SignatureKeys.title);
   });
+  it('should show no content when no data', () => {
+    cy.contains('No signature keys yet').should('be.visible');
+  });
+  it('should contain correct table headers', () => {
+    // galaxykit not supporting signature keys yet, mocking instead
```
himdel, line 17: it's not galaxykit, it's that the api instance is running 4.7.1 without signing services configured

himdel (approved):

nixocio (approved):
🚢

cypress/e2e/hub/signature-keys.cy.ts
ZitaNemeckova: Fixed the comment

PR #1464, opened 2024-01-16
Async select filters, support query for chip label

jamestalton:
This allows async filter which have values like "1" or "2" to have a callback to query the label for the value and show that label in the filter chips.

Screenshot 2024-01-16 at 12 28 01 PM


vidyanambiar (approved):
lgtm

PR #1469, opened 2024-01-16
Simplify column management modal

jamestalton:
The column management dialog is too busy.
This is a simple tweak to make the column management dialog only show the setting for the currently selected view type.


nixocio (approved):
Since tests are passing. 👍

PR #1476, opened 2024-01-17
Table Filters - Support limiting filters to what is supported by the API

jamestalton:
In general table filters support multiple values which are `OR`ed together and then the filters are `AND`ed together

image

> WHERE (Name = 'Job1' OR Name = 'Job2') AND (State = 'Pending' OR State = 'Failed')

There are API limitations where this does not work.

- AWX only support a single OR filter.
- EDA does not support OR filters.
- HUB does not support OR filters, but "should" after a pulp upgrade.
- Analytic does not support OR filters in some cases.

Unfortunately the APIs are not going to be enhanced for the next release.
The next best solution is for the UI to limit the filtering to only what is supported by the API.

This PR adds support for that, adding
- a `TextSingleFilter` for when a text field can only have one value
- a flag `limitFiltersToOneOrOperation` which will be used by AWX to limit to only one OR operation

https://github.com/ansible/ansible-ui/assets/6277895/42856e86-e831-4dad-b7b8-8c2a05d8c8ca


framework/PageToolbar/PageToolbarFilters/ToolbarTextFilter.tsx
```diff
@@ -102,3 +109,83 @@ export function ToolbarTextFilter(props: IToolbarTextFilterProps) {
     </InputGroup>
   );
 }
+
+export function ToolbarSingleTextFilter(props: {
+  id?: string;
+  placeholder?: string;
+  comparison: 'contains' | 'startsWith' | 'endsWith' | 'equals';
+  setValue: (value: string) => void;
+  value: string;
```
MilanPospisil, line 118: This may be better called initialValue, value implies that if I pass value and then change it, it will change in component too. But thats probably already established and in many places in the framework, so we will probably leave it as it is?

frontend/hub/collections/hooks/useCollectionFilters.tsx
```diff
@@ -26,21 +26,21 @@ export function useCollectionFilters() {
       {
         key: 'keywords',
         label: t('Name'),
-        type: ToolbarFilterType.Text,
+        type: ToolbarFilterType.MultiText,
```
MilanPospisil, line 29: MultiText does not work in this filter, should we check (after merge of this PR) and change some of those into SingleText? Im not sure if backend will support this, but you are probably better informed.

framework/PageToolbar/PageToolbarFilter.tsx
```diff
@@ -321,6 +377,7 @@ function ToolbarFilterComponent(props: {
     case ToolbarFilterType.AsyncMultiSelect:
       return (
         <PageAsyncMultiSelect<string>
```
MilanPospisil, line 379: Forgotten limitFiltersToOneOrOperation?

framework/PageToolbar/PageToolbarFilter.tsx
```diff
@@ -266,10 +281,49 @@ function ToolbarFilterComponent(props: {
...
+          placeholder={filter.placeholder}
+          comparison={filter.comparison}
+          setValue={(value) => setFilterValues(() => (value ? [value] : []))}
+          value={filterValues && filterValues?.length > 0 ? filterValues[0] : ''}
+          hasKey={!!filterState?.[filter.key]}
+        />
+      );
+
+    case ToolbarFilterType.MultiText:
+      if (isHasOrFilter && props.limitFiltersToOneOrOperation) {
```
MilanPospisil, line 311: When I turn limitFiltersToOneOrOperation to true, isHasOrFilter is false every time, so I still have multi version. Is it because in PR description, there is note about that it will be used for AWX?

Because isHasOrFilter is evaluated to false when I have active filter, thus the condition is skipped and  limitFiltersToOneOrOperation  does not work.

MilanPospisil:
Summary:
Good Job, I have tested both Single and Multi text filters and they both work as expected in Hub.

Found problems:
limitFiltersToOneOrOperation does not seems to work correctly
one forgotten condition for AsyncMultiSelect

framework/PageToolbar/PageToolbarFilters/ToolbarTextFilter.tsx
jamestalton: left as it is. added jsdoc comments

framework/PageToolbar/PageToolbarFilter.tsx
jamestalton: thanks! fixed

framework/PageToolbar/PageToolbarFilter.tsx
jamestalton: This is working for me. Updated documentation for
limitFiltersToOneOrOperation

```
  /**
   * Limits the filters so that only one filter can be set to an OR operation.
   *
   * Example: AWX can either have an OR on type or status but not both.
   * So once one has 2 selections, the other becomes a single select.
   * Example: (Status is pending or success) and type is inventory.
   */
```

jamestalton: Add tests that verify this functionality.

MilanPospisil: If I understand it correctly, it should automatically convert all multiselects to single select when this is set?

Try pls that in collections.tsx:
      <PageTable<CollectionVersionSearch>
        id="hub-collection-version-search-table"
        toolbarFilters={toolbarFilters}
        toolbarActions={toolbarActions}
        tableColumns={tableColumns}
        rowActions={rowActions}
        errorStateTitle={t('Error loading collections')}
        emptyStateTitle={t('No collections yet')}
        emptyStateDescription={t('To get started, upload a collection.')}
        emptyStateButtonText={t('Upload collection')}
        emptyStateButtonClick={() => pageNavigate(HubRoute.UploadCollection)}
        {...view}
        defaultTableView="list"
        defaultSubtitle={t('Collection')}
        limitFiltersToOneOrOperation={true}
      />
      
      But in the page, I still can select multiple filter:
      
image

MilanPospisil: Oh, ok, one or operation, I though it should disable any or operation :-) Thus convert multiselect to single select. That makes sense. Approving.

MilanPospisil (approved):
LGTM :+1:

PR #1478, opened 2024-01-18
update the overview document link

reidliu41:
it would be better to show the user detail document too, it introduced every part of the EDA.
https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.4/html/event-driven_ansible_controller_user_guide/index


keithjgrant:

We’re sorry for the extra work this may cause. We welcome you to check the new `devel` branch to see if this change is still pertinent in the updated codebase; if so, please port your changes to a new branch and open a new PR against `devel`.

PR #1479, opened 2024-01-18
Add delete to repository list and detail view

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AAH-2484

Repositories -> delete from row dropdown
Repositories -> delete from toolbar dropdown
Repository detail page -> delete from toolbar dropdown


nixocio (approved):
I tested this one manually and it is working properly.

PR #1482, opened 2024-01-18
test Approve a collection

acruzgon:


akus062381 (approved):

cypress/e2e/hub/approvals.cy.ts
```diff
@@ -1,12 +1,46 @@
...
+  after(() => {
+    cy.deleteCollection(thisCollectionName, 'ibm', 'published');
   });
 
   it('it should render the approvals page', () => {
+    cy.verifyPageTitle(Approvals.title);
+  });
+
+  it('user can upload a new collection and approve it', () => {
+    thisCollectionName = 'test' + randomString(4).toLowerCase();
```
MilanPospisil, line 23: Maybe the string can be little longer, for example 5? Just to be sure its not the same as other collection. The name of any entity should start with 'e2e' string - we discussed it in monday, this would help us identifying our testing data and batch deleting old data in script outside of tests.

cypress/e2e/hub/approvals.cy.ts
```diff
@@ -1,12 +1,46 @@
...
+    namespace = 'ibm';
+
+    cy.uploadCollection(thisCollectionName, namespace);
```
MilanPospisil, line 26: This is using galaxykit, correct.

cypress/e2e/hub/approvals.cy.ts
```diff
@@ -1,12 +1,46 @@
...
+    cy.deleteCollection(thisCollectionName, 'ibm', 'published');
   });
 
   it('it should render the approvals page', () => {
+    cy.verifyPageTitle(Approvals.title);
+  });
+
+  it('user can upload a new collection and approve it', () => {
+    thisCollectionName = 'test' + randomString(4).toLowerCase();
+    namespace = 'ibm';
```
MilanPospisil, line 24: Is ibm standard namespace in the system, what if its deleted?
I rather suggest creating own namespace with random string, but its up to discussion. The solution with ibm namespace will be faster, but it can cause potential troubles. For example when two tests will run over cleaned database, both will query namespaces, both will at the same time recieve that ibm not exist, both will attempt to create it, both will fail.

cypress/support/hub-commands.ts
```diff
@@ -305,3 +305,25 @@ Cypress.Commands.add('createRemoteRegistry', (remoteRegistryName: string) => {
...
+  (
+    collectionName: string,
+    namespaceName: string,
+    repository: string,
+    options?: {
+      /** Whether to fail on response codes other than 2xx and 3xx */
+      failOnStatusCode?: boolean;
+    }
+  ) => {
+    cy.requestDelete(
```
MilanPospisil, line 320: This is failing in my local computer due to CSRF token missing. But this can be replaced by calling galaxykit command delete collection.

cypress/e2e/hub/approvals.cy.ts
```diff
@@ -1,12 +1,46 @@
...
+
+    cy.uploadCollection(thisCollectionName, namespace);
     cy.navigateTo('hub', Approvals.url);
     cy.verifyPageTitle(Approvals.title);
+    cy.clickTableRowPinnedAction(thisCollectionName, 'approve');
```
MilanPospisil, line 29: This command is not working in my local tests, there may be some problem with syncing. My tip is that it gets the old row, but then filter is applied and old row is removed, so it cant click it...

PR #1493, opened 2024-01-22
[AAP-8169] Add Templates List to Inventory

daphnemaeve:
Jira ticket: https://issues.redhat.com/browse/AAP-8169

Adds Templates List component to Job Templates tab in Inventories Page. Only shows Job Templates related to the selected Inventory.

Before:
Screenshot 2024-01-22 at 10 44 45 AM

After:
Screenshot 2024-01-22 at 10 35 13 AM


frontend/awx/resources/inventories/InventoryPage/InventoryJobTemplates.tsx
```diff
@@ -0,0 +1,8 @@
+import { useParams } from 'react-router-dom';
+import { awxAPI } from '../../../common/api/awx-utils';
+import { TemplatesList } from '../../templates/TemplatesList';
+
+export function InventoryJobTemplates() {
+  const { id = '' } = useParams<{ id: string }>();
+  return <TemplatesList url={awxAPI`/job_templates/`} inventoryId={id} />;
```
skochay330, line 7: The row actions and filter categories are a bit different here than what I see in the old UI- was the `TemplatesList` component changed intentionally or was something missed?

skochay330 (commented):
Great work!

frontend/awx/resources/inventories/InventoryPage/InventoryJobTemplates.tsx
daphnemaeve: Thanks for noticing those! I've added the copy button to the row actions. Just double checked to confirm, and the missing filter categories will likely be added as part of Advanced Search, so for now it should be okay without those.

kialam (changes requested):
This is looking really good. I'm noticing a bug where when you copy a template from the list view, the list does not show the newly copied template until after a refresh.

kialam (approved):
LGTM! Pulled this down and tested locally.

PR #1498, opened 2024-01-23
[do not merge] Access Tabs for Hub resources (testable with namespaces)

vidyanambiar:
### Jira Issues
[AAH-3052](https://issues.redhat.com/browse/AAH-3052)
[AAH-3053](https://issues.redhat.com/browse/AAH-3053)
[AAH-3054](https://issues.redhat.com/browse/AAH-3054)

### Design changes from existing UI:
- In the existing UI, the Access tab for a resource contains 2 lists, one for users and one for teams (groups). In the new UI, the single access tab is broken into 2 tabs "User access" and "Team access" each containing a list of users and teams respectively.
- Clicking on a user/team in the "User/Team access" tab brings up the list of roles for that user/team. The tabs would change to reflect a "Back to user/team access" tab and a "Roles" tab. The breadcrumbs reflect the current selected tab.

### Changes in this PR:
- **Hub**
    - Reusable components to implement lists for users, teams and roles in the access tabs for a hub resource
    - This PR _does not_ include the actions for adding or deleting a user/team/role. It will be added separately.
    - Updates to Namespace page adding tabs for User access and Team access 
- **Framework**
    - updated `useInMemoryView` to handle text-based filters correctly
    - PageBreadcrumbs: Provider and hook for managing breadcrumbs in the PageHeader and PageRoutedTabs components. This allows the currently active tab to be displayed in the breadcrumbs.

https://github.com/ansible/ansible-ui/assets/43621546/2719bb03-821b-4974-9116-8afc121c7980


frontend/hub/access/resource-access/HubResourceAccessRoles.tsx
```diff
@@ -0,0 +1,159 @@
...
+        query: 'name',
+        comparison: 'contains',
+        placeholder: t('Filter by name'),
+        isPinned: true,
+      },
+    ];
+    return filters;
+  }, [t]);
+
+  const view = useInMemoryView<HubAccessRole>({
```
nixocio, line 126: @vidyanambiar I am  just starting to look at this PR. It does look pretty good.

I could not follow why we need to use the useInMemoryView, neither why the changes were required there. Is it something that we will need to use in order to use the new approach for breadcrumbs? 

Could we document this new hook, and how to use them?

vidyanambiar: @nixocio Thanks for taking a look!

Reason for using the `useInMemoryView` hook: Unlike most other list UIs, this one does not have a direct API call for loading the users/teams table so we cannot utilize `usePulpView` or `useHubView` which are associated with an API call for fetching the data for the table. We need to user the `users` and `groups` array that is received within the namespace data response. So the table is constructed using these static arrays. The `useInMemoryView` is a hook provided in the framework to create a PageTable from static data. And it provides the utilities required like selectable rows, filtering etc. I needed to make a small update for making the filtering work correctly for text-based filters.

framework/PageHeader.tsx
```diff
@@ -107,6 +109,31 @@ export function PageHeader(props: PageHeaderProps) {
   const isXl = useBreakpoint('xl');
   const isMdOrLarger = useBreakpoint('md');
   const [translations] = useFrameworkTranslations();
+  const { pageBreadcrumbs, baselineBreadcrumbs, setBaselineBreadcrumbs, setPageBreadcrumbs } =
+    usePageBreadcrumbs();
+
+  useEffect(() => {
+    if (breadcrumbs?.length) {
+      setBaselineBreadcrumbs(breadcrumbs ?? []);
+      setPageBreadcrumbs([...baselineBreadcrumbs]);
```
himdel, line 118: not sure I understand this bit...

if breadcrumbs were passed from props, this sets baseline breadcrumbs to those, and then uses the previous value of baseline breadcrumbs to set page breadcrumbs?

himdel:
Something is broken, I can access both Access tabs under namespaces,
but:

1. be on https://localhost:4102/namespaces/abc/user-access?page=1&perPage=10&sort=username
2. click Add user
3. https://localhost:4102/namespaces/:id/user-access-add-user renders, with a Under Development screen

(same for groups)

(looks like it doesn't set the namespace id in the url)

framework/PageHeader.tsx
vidyanambiar: I had added this for the cases where PageHeader is used independently from PageRoutedTabs. So that pageBreadcrumbs default to the baseline breadcrumbs if there are no PageRoutedTabs to go along with it.

vidyanambiar:
@himdel Yes, https://localhost:4102/namespaces/:id/user-access-add-user does not work yet in this PR. I'm fixing up the routing for that along with hooking up the add user action.

PR #1502, opened 2024-01-24
Additional tests - Signature-keys

vishsanghishetty:
## Backend on CI not supporting signing yet so it's not possible to create signature key, mocking instead
This PR is for [ticket](https://issues.redhat.com/browse/AAH-3063) 

**Tests**
- [x] should display fingerprint-mocked data on the list page 
- [x] should display name and public key mocked data in the list page

### DIFF
--- a/cypress/e2e/hub/signature-keys.cy.ts
+++ b/cypress/e2e/hub/signature-keys.cy.ts
@@ -6,12 +6,9 @@ describe('Signature Keys', () => {
     cy.hubLogin();
   });
 
-  it('should render the signature keys page', () => {
+  it('should render the signature keys page and display empty state when no content when no data', () => {
     cy.navigateTo('hub', SignatureKeys.url);
     cy.verifyPageTitle(SignatureKeys.title);
-  });
-  it('should display empty state when no content when no data', () => {
-    cy.navigateTo('hub', SignatureKeys.url);
     cy.get('[data-cy="empty-state-title"]')
       .should('contain', 'No signature keys yet')
       .should('be.visible');
--- a/cypress/e2e/hub/signature-keys.cy.ts
+++ b/cypress/e2e/hub/signature-keys.cy.ts
@@ -2,7 +2,7 @@ import { SignatureKeys } from './constants';
 import { pulpAPI } from '../../support/formatApiPathForHub';
 
 describe('Signature Keys', () => {
-  before(() => {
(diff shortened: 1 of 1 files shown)


cypress/e2e/hub/signature-keys.cy.ts
```diff
@@ -2,16 +2,59 @@ import { SignatureKeys } from './constants';
 import { pulpAPI } from '../../support/formatApiPathForHub';
 
 describe('Signature Keys', () => {
-  before(() => {
+  beforeEach(() => {
```
nixocio, line 5: `before` should do that job here, since we are leveraging the `cy.sesssion` to avoid to login every single test, right? Did I miss something else?

cypress/e2e/hub/signature-keys.cy.ts
```diff
@@ -2,16 +2,59 @@ import { SignatureKeys } from './constants';
...
     cy.hubLogin();
   });
 
   it('should render the signature keys page', () => {
```
nixocio, line 9: The test `should display empty state when no content when no data` could be combined here.

cypress/e2e/hub/signature-keys.cy.ts
```diff
@@ -2,16 +2,59 @@ import { SignatureKeys } from './constants';
...
-  it('should show no content when no data', () => {
-    cy.contains('No signature keys yet').should('be.visible');
+  it('should display empty state when no content when no data', () => {
+    cy.navigateTo('hub', SignatureKeys.url);
+    cy.get('[data-cy="empty-state-title"]')
+      .should('contain', 'No signature keys yet')
+      .should('be.visible');
+  });
+  it('should display name and public key mocked data in the list page', () => {
+    // backend on CI not supporting signing yet so it's not possible to create signature key, mocking instead
```
nixocio, line 20: 👍

nixocio (approved):
@vishsanghishetty 🥇 a few minor suggestions.

akus062381 (approved):

cypress/e2e/hub/signature-keys.cy.ts
vishsanghishetty: yes, thanks for bringing it up Kersom
with before hook, tests fail in this case, only with beforeEach hook the two tests I wrote pass.

PR #1512, opened 2024-01-25
Repository detail add remove collection

MilanPospisil:
Divide collection version in repository into two separate tabs:
Remove collection version and Add collection version.
image

It is because, in old application, after clicking add collection, the modal with selector collection table appear. But we have in collection table toolbar filters with repository selector and Browse will open modal within modal, thats not what we want. So the adding is now one of the tabs. May check with UX if thats ok.

Remove collection version from repository as batch operation was removed from collections and repository collections because it fails - this may be problem in pulp that cant handle multiple modify operations over one repo in backend. Only first operation succeeds, others fail. But bulk operations were not present in old UI, so thats ok.
image

Add tab:
image

After clicking the add, classic confirmation dialog appears:
image

While the old functionality has only collection tab with collections list to delete, when add is clicked, it opens modal:
image

I decided to change the UI, because we cant have modal for this. In toolbar filter, there is option to open new modal for filtering collections by repository with another table for selections. We dont want two modals above each other. So I decided to add new tab.

Another fix - delete and add collection from/to repository also load latest version of repo in order to modify it. It now does not rely on latest version that previously loaded the user interface. For example, bulk actions will not work without it (in all collection list) - the first delete will modify repo, which will create new version (if there is not retain repo versions set to 1), and the other pending deletes will then reference old repo version and will fail. This is not database transaction, so it will still fail occasionaly, but this fix reduce the number of indicents.


MilanPospisil:
Closing, will be replaced by two separate PRs, one will be this: https://github.com/ansible/ansible-ui/pull/1590

PR #1513, opened 2024-01-25
Refactor deleted node to use Custom Node

AlexSCorey:
Currently, a node whose resource has been deleted shows as Deleted with a Trash Icon, but no context menu.  This PR refactors Topology.tsx to remove the deleted node type and simply uses the custom node, and all of its functionality that already exists.  The first image is a Deleted node as it currently is.  The second image is a Deleted node from this branch.
Screenshot 2024-01-25 at 11 44 31 AM
Screenshot 2024-01-25 at 11 44 11 AM


marshmalien (approved):
🎉

PR #1525, opened 2024-01-26
Repositories tab versions action revert

MilanPospisil:
Repository detail, tab versions and action revert. It reverts repository to actual version.

How to test it - you need repository with max retain versions for example 100. Default is 1, so there is only 1 version and you cant test it.

Then you can add some collection to the repository (for example using old UI - add collection to repository). Then new repo version will appear and we can revert repo to original one. Then the collections should dissapear from repo again.

We cant revert to latest version of repo.


ZitaNemeckova (approved):
LGTM :+1: Tested in UI.

PR #1528, opened 2024-01-26
Cleanup workflow circular dependencies and refactor hooks

marshmalien:
* Remove circular dependencies within "/WorkflowVisualizer/*"
* Remove unused hooks
* Move shared constants and interfaces from components to shared files 
* Save RBAC on controller state
* Introduce useCloseSidebar hook to handle updating the sidebar mode and resetting selected ids


AlexSCorey (approved):
I think this looks good.

PR #1529, opened 2024-01-27
[AAP-8285] Add inventory group add page

matoval:
Jira ticket: https://issues.redhat.com/browse/AAP-8285

Add inventory group add page to create new groups

Before:
image

After:
image


frontend/awx/interfaces/InventoryGroupCreate.ts
```diff
@@ -0,0 +1,6 @@
+export interface InventoryGroupCreate {
```
AlexSCorey, line 1: Lets consider exporting this from the InventoryGroup interface, which already exists.

frontend/awx/resources/inventories/inventoryGroup/GroupPageAdd.cy.tsx
```diff
@@ -0,0 +1,38 @@
+import { GroupPageAdd } from './GroupPageAdd';
```
AlexSCorey, line 1: We should probably have some RBAC tests as part of this PR also.  

- We should test that the the Create group button on the list is disabled.  
- We should also have some e2e tests that assert that we can get from login, to the form, create a group, and then delete the group.  (Consider reaching out to QE for their thoughts on other e2e tests).

frontend/awx/resources/inventories/inventoryGroup/GroupPageAdd.tsx
```diff
@@ -0,0 +1,54 @@
+import { useTranslation } from 'react-i18next';
```
AlexSCorey, line 1: The name of the file...  Generally our conventions has been to name the file that wraps the form fields something like InventoryGroupForm.  Inside it we export 2 functions 1 for add and another for edit  (See ProjectForm for an example).  

While this request is super nitpicky I think it would improve developer experience if we can keep our file name in line with the conventions.

AlexSCorey (changes requested):
This looks really good.  I have some really small nitpicky suggestions.

Note:  Not part of your work in this PR, but I did notice that after creating an inventory group it takes you to the details page, as it should.  On that page, the `Back to groups` tabs sends you to a weird url.  If you are doing other inventory group views you will probably come across this.  See below.
Screenshot 2024-01-29 at 9 07 31 AM

frontend/awx/resources/inventories/InventoryPage/InventoryGroups.cy.tsx
```diff
@@ -109,4 +120,14 @@ describe('InventoryGroups', () => {
         cy.clearAllFilters();
       });
   });
+  it('disables Create group button when user dose not have permissions', () => {
```
AlexSCorey, line 123:
```suggestion
  it('disables Create group button when user does not have permissions', () => {
```

AlexSCorey (approved):
Just 1 small spelling change.  Fix that and you are good to go. Well done!

PR #1530, opened 2024-01-28
Hub - check trailing and double slashes when constructing API urls, better page/sort param support

himdel:
Fixes AAH-3007, follows #1408 

`hubAPI` and `pulpAPI` will now throw when any of these happen:
* `HUB_API_PREFIX` ends with a trailing slash
* prefix-relative url doesn't start with a slash (already happening)
* url doesn't end in a slash (end of string or before a ?)
* url contains a double slash
* url contains a double ?
* the wrong param is used based on the url (`page, limit, offset, page_size, sort, ordering, order_by`)

add `hubQueryString` to wrap `normalizeQueryString` and:
* escape values
* convert page/limit/offset/page_size/sort/ordering/sort_by to the right name

allow a `` hubAPI`/foo/?${{ param1: 'abc', param2: 123 }}` `` syntax, calling `hubQueryString`

fixes all instances of `hubAPI` & `pulpAPI` missing a trailing slash

also, this allows us to fix all those `|| ''` and `?? ''` and `as string` in urls

and fix tests

---

In e2e tests, `hubAPI` and `pulpAPI` retain their simpler versions, I'm not sure how to handle the use of `*` wildcards without allowing them in regular code.

### DIFF
--- a/cypress/e2e/hub/approvals.cy.ts
+++ b/cypress/e2e/hub/approvals.cy.ts
@@ -12,6 +12,7 @@ describe('Approvals', () => {
     cy.hubLogin();
     cy.getNamespace(namespace);
     cy.uploadCollection(thisCollectionName, namespace);
+    cy.galaxykit('task wait all');
 
     cy.navigateTo('hub', Approvals.url);
   });
@@ -24,7 +25,6 @@ describe('Approvals', () => {
   it('user can upload a new collection and approve it', () => {
     cy.navigateTo('hub', Approvals.url);
     cy.verifyPageTitle(Approvals.title);
-    cy.clickButton(/^Clear all filters$/);
     cy.clickTableRowPinnedAction(thisCollectionName, 'approve');
 
     //Verify Approve and sign collections modal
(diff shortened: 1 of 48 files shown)


cypress/support/formatApiPathForHub.ts
```diff
@@ -9,7 +9,11 @@
 import { apiTag } from '../../frontend/hub/common/api/formatPath';
 
 function getBaseAPIPath() {
-  return (Cypress.env('HUB_API_PREFIX') as string) || '/api/galaxy';
+  const base = (Cypress.env('HUB_API_PREFIX') as string) || '/api/galaxy';
+  if (base.endsWith('/')) {
+    throw new Error(`Invalid HUB_API_PREFIX - must NOT end with a slash`);
```
nixocio, line 14: 🥇

cypress/e2e/hub/approvals.cy.ts
```diff
@@ -12,6 +12,7 @@ describe('Approvals', () => {
     cy.hubLogin();
```
nixocio, line 12: @acruzgon Please review the approval tests changes.

acruzgon (approved):
lgtm 
image

cypress/e2e/hub/approvals.cy.ts
acruzgon: all tests passed
image

himdel:
rebased, and updated the offset <-> page conversion to fix an off-by-one (page starts at 1, offset at 0: https://github.com/ansible/ansible-ui/pull/1530/commits/b600638eab405d8ab74731d06896ec5ad967ff7b)

ZitaNemeckova (approved):

PR #1534, opened 2024-01-29
Add Execution Environment detail page with tabs and actions

jerabekjiri:
Screenshot from 2024-01-29 15-58-17

### DIFF
--- a/frontend/hub/execution-environments/ExecutionEnvironmentPage/hooks/useExecutionEnvironmentPageActions.tsx
+++ b/frontend/hub/execution-environments/ExecutionEnvironmentPage/hooks/useExecutionEnvironmentPageActions.tsx
@@ -32,19 +32,19 @@ export function useExecutionEnvironmentPageActions() {
         type: PageActionType.Button,
         selection: PageActionSelection.Single,
         label: t('Use in controller'),
-        onClick: () => alert('WIP'),
+        onClick: () => {},
       },
       {
         type: PageActionType.Button,
         selection: PageActionSelection.Single,
         label: t('Delete'),
-        onClick: () => alert('WIP'),
+        onClick: () => {},
       },
       {
         type: PageActionType.Button,
         selection: PageActionSelection.Single,
         label: t('Sign'),
-        onClick: () => alert('WIP'),
+        onClick: () => {},
       },
     ];
     return actions;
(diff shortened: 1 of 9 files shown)


frontend/hub/execution-environments/ExecutionEnvironmentPage/hooks/useExecutionEnvironmentPageActions.tsx
```diff
@@ -0,0 +1,52 @@
...
+        label: t('Edit'),
+        onClick: (ee) =>
+          pageNavigate(HubRoute.EditExecutionEnvironment, { params: { id: ee?.name } }),
+      },
+      { type: PageActionType.Seperator },
+      {
+        type: PageActionType.Button,
+        selection: PageActionSelection.Single,
+        label: t('Use in controller'),
+        onClick: () => alert('WIP'),
```
himdel, line 35: just make sure to remove all the `alert`s soon :)

himdel (approved):

frontend/hub/execution-environments/ExecutionEnvironmentPage/hooks/useExecutionEnvironmentPageActions.tsx
jerabekjiri: I'll remove that, wasn't paying attention

jerabekjiri:
@himdel Is it okay to change EE paths to `execution-environments/alpine/activity`, the old UI is `containers/alpine/_content/activity/`?

himdel:
yeah, UI paths will be different in the new UI, that's fine :)

(just make sure it still works when the container is `foo/bar` instead of `alpine`)

jerabekjiri:
oh, that's a good point, I have to test what happens with `foo/bar` format, probably needs to be fixed.

jerabekjiri:
Looks like react router escapes into `foo%2Fbar`, not sure if that's the desired output

himdel:
Yeah, as long as `foo%2Fbar` in the url works, and the rest of the ui works as expected, I think that's close enough for now :)

The old ui implements it as 2 sets of routes, one with `/:namespace/:name/`, other with just `/:name/` but it's a bit of a hack, it could be replaced with optional params now (in the old ui, along the lines of `/:namespace?/:name/`), but the hierarchical router here may need some work to support that...

We can leave supporting foo/bar in the url as a separate issue, as long as foo%2Fbar works :)

PR #1539, opened 2024-01-30
Repair collection upload

MilanPospisil:
Fixes:
   load only one namespace instead of loading all of them (the API will not return all of them anyway)
   error handling in clicking on submit instead of page watch

### DIFF
--- a/frontend/hub/collections/UploadCollection.tsx
+++ b/frontend/hub/collections/UploadCollection.tsx
@@ -182,11 +182,12 @@ export function UploadCollectionByFile() {
     let lastError = '';
 
     try {
-      lastError = t('Can not find namespace {{namespaceName}}', { namespaceName });
+      lastError = t('Error in loading namespace {{namespaceName}}', { namespaceName });
       const namespace = await requestGet<HubItemsResponse<HubNamespace>>(
         hubAPI`/_ui/v1/namespaces/?limit=1&name=${namespaceName}`
       );
       if (namespace.data.length === 0) {
+        lastError = t('Can not find namespace {{namespaceName}}', { namespaceName });
         throw new Error('');
       }
 
--- a/frontend/hub/collections/UploadCollection.tsx
+++ b/frontend/hub/collections/UploadCollection.tsx
@@ -1,11 +1,10 @@
-import { Alert, Radio } from '@patternfly/react-core';
+import { Radio } from '@patternfly/react-core';
 import React, { useEffect, useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useNavigate } from 'react-router-dom';
 import {
   ITableColumn,
   IToolbarFilter,
-  PageDetails,
   PageHeader,
   PageLayout,
   PageTable,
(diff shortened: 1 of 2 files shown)


frontend/hub/collections/UploadCollection.tsx
```diff
@@ -170,82 +162,70 @@ export function UploadCollectionByFile() {
...
 
-    return hubPostRequestFile(
-      hubAPI`/v3/plugin/ansible/content/${base_path}/collections/artifacts/`,
-      data.file as Blob
-    ).then(() => pageNavigate(HubRoute.Approvals));
+    let lastError = '';
+
+    try {
+      lastError = t('Error in loading namespace {{namespaceName}}', { namespaceName });
+      const namespace = await requestGet<HubItemsResponse<HubNamespace>>(
```
nixocio, line 186: This error handle is very different from the other fields, in general we are either showing the error associated with a certain field, or displaying an error at the form level. The styling also does not fit with the remaining errors one. 

image



If a HTTP request error is raising inside the HubPageForm it will be parsed to either be displayed at form level, or at field level. It might be worth to explore raising a RequestError and pass the proper strings there.

MilanPospisil: However, this is not using PageForm... It was written before we started to do any PageForms and also the basics were written by James and he also did not used PageForm. Do we want to rewrite it? Mind that we dont have enough time. Some smaller fix to make it look more similar will be nicer.

nixocio: @MilanPospisil The submit is wrapped on the HubPageForm submit and any HTTP error raised inside that will bubble up and catch up by the HubPageForm - as long we do not catch the error inside the submit.

At least for the upload field itself we could do something like below, perhaps we could a function raiseFieldError, and raiseFormError to the framework, or as helpers.

I added  a function

 ```

async function raiseFieldError(fieldName: string, message: string, status = 404) {
  const responseBody = JSON.stringify({ [fieldName]: [message] });
  const response = new Response(responseBody, {
    status,
    statusText: httpStatusCodes[status],
    headers: { 'Content-Type': 'application/json' },
  });
  throw await createRequestError(response);
}

const httpStatusCodes: { [key: number]: string } = {
  200: 'OK',
  201: 'Created',
  204: 'No Content',
  400: 'Bad Request',
  401: 'Unauthorized',
  403: 'Forbidden',
  404: 'Not Found',
  500: 'Internal Server Error',
  502: 'Bad Gateway',
  503: 'Service Unavailable',
};
```

Then inside the onSubmit I added 

```

  if (namespace.data.length === 0) {
      await raiseFieldError(
        'file',
        t('Can not find namespace {{namespaceName}}', { namespaceName })
      );
    }

```


image

MilanPospisil: You are right, I missed that it is using hub page form component.

MilanPospisil: James forked https://github.com/ansible/ansible-ui/pull/1546

himdel (approved):
LGTM,

looks like https://github.com/ansible/ansible-ui/pull/1546 got stalled,
and this is still the first step :+1:

himdel:
Might be a good idea to rebase this one before merging and letting the tests pass again, just to make sure there are no interactions with the recent approval, signing and worker fixes.

PR #1541, opened 2024-01-30
Dialogs - Support stack of active dialogs

jamestalton:
We have use cases that need to be able to open dialogs on top of dialogs.
Example having a dialog with a table, which has a filter, which has a browse button, which opens a second dialog.


MilanPospisil (approved):
Thanks James, looking good. I will test it in more depth when I will use it. So lets merge!

PR #1547, opened 2024-01-31
[AAP-8310] Add inventory group edit form

matoval:
Jira ticket: https://issues.redhat.com/browse/AAP-8310

Add inventory group edit form

Before:
image

After:
image


kialam:
Very nice. This is almost there. I noticed that if I try to edit an inventory group, it does not take me to the edit form standalone page:

inv-group-detail-tab-edit
Also note that we have two headers here "demo Inv test group" and "Inventory Groups." Tab list views should not have header titles.

kialam:
We should add integration tests for this feature to make sure a user can add/edit a group from either the list view or tab view.

kialam (changes requested):
Well done. There's a few UI discrepancies when the user tries to edit from a tabbed view. We also should add e2e tests to this feature.

frontend/awx/resources/groups/GroupEdit.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const onSubmit: PageFormSubmitHandler<InventoryGroupCreate> = async (groupInput) => {
+    const { name, description, variables } = groupInput;
+    const editGroup: InventoryGroupCreate = {
+      name,
+      description,
+      variables,
+      inventory: group.inventory,
+    };
+    const editedGroup = await putRequest(awxAPI`/groups/${group.id.toString()}`, editGroup);
```
kialam, line 31: Looking at the other forms in our project, we should follow the pattern of using `patchRequest` instead of `putRequest` here.

kialam: The request URL is missing a trailing `/` in the url.

cypress/e2e/awx/resources/inventories.cy.ts
```diff
@@ -127,6 +127,21 @@ describe('inventories', () => {
...
+      cy.verifyPageTitle(inventory.name);
+      cy.clickLink(/^Groups$/);
+      cy.clickTableRowKebabAction(group.name as string, 'edit-group', true);
+      cy.verifyPageTitle('Edit group');
+      cy.get('[data-cy="name-form-group"]').type('-changed');
+      cy.get('[data-cy="Submit"]').click();
+      cy.verifyPageTitle(group.name + '-changed');
+    });
+  });
+
```
kialam, line 144: Let's add another test case of editing from the detail page as well. This case handles editing from a table row.

kialam (changes requested):
A few more things to adjust, but this is looking good. Notably there is a redirect bug when clicking away from the edit form.

kialam (approved):
🎉 Awesome job! Good to merge once tests pass.

PR #1553, opened 2024-01-31
Collections additional info

MilanPospisil:
Adds roles, plugins and dependecies count to collection list (list and cards view). Also adds clickable links to namespaces and repositories.

image


himdel (approved):

himdel:
(lgtm, but these are also missing in collection detail)

MilanPospisil:
> (lgtm, but these are also missing in collection detail)

Where exactly? Collection detail is probably generated from the collection list, so it contains those links too.

himdel:
Oh never mind, this fixed both :)

PR #1554, opened 2024-01-31
fix collection names and approvals links

MilanPospisil:


himdel:
Looks good,

old ui - list ,detail




new ui - list, detail




LGTM :)

(but if you wanted to get 100% there, the "provided by" bit should link to the namespace,
and the detail screen breadcrumbs should have a namespace too (though maybe the breadcrumbs are supposed to be different here :shrug: )

PR #1555, opened 2024-01-31
[Hub > Execution Environment] Fix small EE issues

jerabekjiri:
No-Issue

Fixes:
- Infinite loading on locally pushed images in EE/edit screen
- Create/edit EE: `Error Invalid param - use "limit", not "page_size" (url: /_ui/v1/execution-environments/registries/?page_size=50)`
- Editing empty EE description: `RequestError Bad Request`
- Signing status badge: 
Screenshot from 2024-02-05 00-11-14
- uncaught error: 
```
{
    "errors": [
        {
            "status": "400",
            "code": "invalid",
            "title": "Invalid input.",
            "detail": "The provided base path contains forbidden characters.",
            "source": {
                "parameter": "base_path"
            }
        }
    ]
}
``` 
(returned parameter is `base_path` instead of `name`, which doesn't show up in the form errors and has to be mapped to `name`) 
- wait for task finish after creating/editing EE


himdel (approved):

himdel (approved):
Still looking good :)

Cc @nixocio for `mappedKeys`, sounds like this could be useful in more forms.

nixocio (commented):
@jerabekjiri @himdel `mappedKeys` look great. Ty.

frontend/hub/common/adapters/hubErrorAdapter.tsx
```diff
@@ -46,7 +49,10 @@ export const hubErrorAdapter: ErrorAdapter = (error): ErrorOutput => {
       // Handling Galaxy errors
       for (const e of data.errors as GalaxyError[]) {
         if (e.source && e.source.parameter) {
-          fieldErrors.push({ name: e.source.parameter, message: e.detail || e.title });
+          if (mappedKeys) {
```
nixocio, line 52: I might be misreading this, `base_path` is not a field on the form, should we render that one as form level error, since react-hook-from will not be able to associate `base_path` with any present field?

jerabekjiri: In this case the error is associated with the wrong entered `name`, the problem is that the API error returns parameter as a `base_path` instead, so I think this is exactly case for form level error.

(without this users doesn't really know what field is base_path and has to try again)

In this case, the error is associated with the incorrectly entered `name`. The problem is that the API error returns the parameter as a `base_path` instead. So, I think this really is form level error. Without this information, users don't really know which field is base_path and have to try again.

nixocio (approved):

PR #1562, opened 2024-02-01
fixing flaxy HUB tests

acruzgon:


vishsanghishetty (approved):

cypress/e2e/hub/collections.cy.ts
```diff
@@ -2,16 +2,19 @@ import { randomString } from '../../../framework/utils/random-string';
 import { hubAPI } from '../../support/formatApiPathForHub';
 import { Collections } from './constants';
 
-describe('Collections- List View', () => {
+describe.skip('Collections- List View', () => {
```
nixocio, line 5: @acruzgon is that the intent to skip those tests?

nixocio (approved):
A fix to unblock those tests are coming soon.

cypress/e2e/hub/collections.cy.ts
vishsanghishetty: yes, temporarily skipping until they are less flaky @nixocio

PR #1568, opened 2024-02-02
implemented add inventory source form to sources page list

resoluteCoder:
Jira ticket: https://issues.redhat.com/browse/AAP-8287

Before:
image

After: 
image


prat98 (commented):
I noticed the old UI has a `Template additional resources and host vars at runtime` source option. Was the left out for a particular reason?
Screenshot 2024-02-05 at 11 48 41 AM

resoluteCoder:
@prat98 Yes. In the old UI, clicking that option renders nothing of value. It looks like an unfinished source type.

PR #1569, opened 2024-02-02
Namespaces > test bulk delete namespace

acruzgon:
this PR closed [AAH-2908](https://issues.redhat.com/browse/AAH-2908)

### DIFF
--- a/cypress/e2e/hub/namespaces.cy.ts
+++ b/cypress/e2e/hub/namespaces.cy.ts
@@ -4,6 +4,11 @@ import { Namespaces, MyImports } from './constants';
 const apiPrefix = Cypress.env('HUB_API_PREFIX') as string;
 
 describe('Namespaces', () => {
+  const testSignature: string = randomString(5, undefined, { isLowercase: true });
+  function generateNamespaceName(): string {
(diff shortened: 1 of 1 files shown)


cypress/e2e/hub/namespaces.cy.ts
```diff
@@ -132,4 +132,27 @@ describe('Namespaces', () => {
 
     cy.deleteNamespace(namespaceName);
   });
+
+  it('user can bulk dekete namespaces', () => {
+    let namespaceName: string;
+
+    for (let i = 1; i <= 5; i++) {
+      namespaceName = `namespace_bulk_delete_${i}_${randomString(5, undefined, {
```
nixocio, line 140: @acruzgon This can filter 5 existing namespaces and delete them. Ideally, we should create a unique signature for the created namespaces, filter by that signature, bulk delete all of those, and then filter again to ensure that no namespace with the unique signature remains. This approach is just to be less error-prone regarding which namespaces would be deleted. See the remotes and remote-registries tests for an idea

nixocio (changes requested):

nixocio (approved):
💯

vishsanghishetty (approved):

PR #1577, opened 2024-02-05
Hub tests - deleteNamespace, createNamespace - use galaxykit, merge

himdel:
Rename `getNamespace` to `createNamespace` - dedup with existing createNamespace,
use galaxykit.

Update `deleteNamespace` to use galaxykit and not be fatal when it fails.

### DIFF
--- a/cypress/support/hub-commands.ts
+++ b/cypress/support/hub-commands.ts
@@ -4,7 +4,6 @@ import { Role } from '../../frontend/hub/access/roles/Role';
 import { CollectionVersionSearch } from '../../frontend/hub/collections/Collection';
 import { parsePulpIDFromURL } from '../../frontend/hub/common/api/hub-api-utils';
 import { HubItemsResponse } from '../../frontend/hub/common/useHubView';
-import { HubNamespace } from '../../frontend/hub/namespaces/HubNamespace';
 import './commands';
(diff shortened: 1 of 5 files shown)


ZitaNemeckova (approved):

cypress/support/hub-commands.ts
```diff
@@ -230,30 +229,11 @@ Cypress.Commands.add('cleanupCollections', (namespace: string, repo: string) =>
...
-  });
-});
-
-// Cypress.Commands.add('deleteNamespace', (namespaceName: string) => {
-//   cy.galaxykit('namespace delete', namespaceName);
-// });
-
 Cypress.Commands.add('deleteNamespace', (namespaceName: string) => {
-  cy.requestDelete(hubAPI`/_ui/v1/namespaces/${namespaceName}/`);
+  cy.galaxykit('-i namespace delete', namespaceName);
```
nixocio, line 236: Do we need the `-i`?

nixocio (approved):

cypress/support/hub-commands.ts
himdel: I'm not sure we can always guarantee that a namespace can be deleted:
* rejected collections will break it
* repositories with multiple retain versions will break it
* orphaned collections will break it

So I think we'll have to rely on an async cleanup tasks to make sure namespaces are gone.

PR #1578, opened 2024-02-05
Hub - CI - galaxykit - use unique user per run

himdel:
galaxykit requests a token for the current user,
but one user can only have one token at a time

when multiple galaxykit instances access the same CI server, they can clash by changing the token under each other's feet

creating a separate per-run user, so that each run can only break itself


himdel:
Closing in favor of https://github.com/ansible/ansible-ui/pull/1592
which allows for multiple cypress instances per PR

PR #1593, opened 2024-02-06
Fix: Save instance groups to job template

marshmalien:
Associate selected instance groups in the job template form on submit by fixing the bug that sends an empty array; change the registered input name from `instanceGroups` to `instance_groups`. Additionally, include a component test to ensure the correct instance groups payload is sent to the API.

Other fixes included:
* Fix unknown prop warning `isDisabled` PageFormMultiInput ChipHolder console error by leveraging [styled-components transient props](https://styled-components.com/docs/api#transient-props)
 > Warning: React does not recognize the `isDisabled` prop on a DOM element.

* Fix[ PatternFly Switch error](https://github.com/patternfly/patternfly-react/blob/b1aaf3cb9d48aef7ca1a72977ca386a3f36eecad/packages/react-core/src/components/Switch/Switch.tsx#L59) by passing a value to aria-label
> Switch: Switch requires either a label or an aria-label to be specified


AlexSCorey (approved):
Looks good to me.

AlexSCorey:
@vidyanambiar @akus062381

vishsanghishetty (approved):

akus062381 (approved):

PR #1597, opened 2024-02-06
Update page columns to accept date format as an option

vidyanambiar:


vidyanambiar:
Closing as @jamestalton is working on a solution to have date format picked up and applied globally from the user settings dialog.

PR #1598, opened 2024-02-06
Repository detail add collection

MilanPospisil:
Adds ability to add collection to particular repository at repositories -> detail -> tab collection versions.

 Button Add Collections in toolbar or empty state opens a modal that allows selection of collections. After clicking Confirm button, collections are added as one request to server (endpoint modify).
 
 What is hard in this PR is multiple modals - that required changes in the framework, because collections has repository selector filter, that has browse button, that opens another modal. So ToolbarAsyncBuilder (single and multi) has to be enhanced to not uses setDialog function, but pushDialog and popDialog props that can handle multiple modals.
 
 
image


ZitaNemeckova (approved):
LGTM :+1: Verified that old UI behaves the same way.

ZitaNemeckova:
Clicking Manage columns in the modal closes it.

PR #1600, opened 2024-02-06
Launch a workflow job and navigate to output

AlexSCorey:
The user can launch a workflow job and navigate to the workflow output, which has not yet been implemented.
Addresses https://issues.redhat.com/browse/AAP-19087
Screenshot 2024-02-06 at 2 20 14 PM


resoluteCoder (approved):
lgtm, pulled down and tested locally

PR #1601, opened 2024-02-06
Remove navigate to remote registry during clean up

nixocio:
Remove navigate to remote registry during clean up


skochay330 (approved):
Passes against testing server locally, lgtm

PR #1605, opened 2024-02-07
Repair multitext

MilanPospisil:
Converts multi texts to single texts - checked that  our API cant do multi text. By multi text filter I mean the filter that allows user to do query for example collection name where 'aaa' OR 'bbb'.


ZitaNemeckova:
`MultiText` in `frontend/hub/access/roles/hooks/useRoleFilters.tsx` needs discussion with Vidya. Will be solved in followup PR.

jamestalton (approved):

ZitaNemeckova (approved):

PR #1611, opened 2024-02-07
Bump cypress retries

nixocio:
Bump cypress retries


jamestalton:
Not sure we should bump to 3.
Cypress runs a test and if it fails, retries.
If the retry passes, it marks it as a flaky test.
Cypress already reties the test 2 times, 3 seems like overkill.

nixocio:
> Not sure we should bump to 3. Cypress runs a test and if it fails, retries. If the retry passes, it marks it as a flaky test. Cypress already reties the test 2 times, 3 seems like overkill.

👍 Just plain with the heuristics of the test failures here.

PR #1612, opened 2024-02-07
UI Settings Support User Date Format Preference

jamestalton:
Makes date formatting in the UI be based on user preference.


vidyanambiar (commented):
🎉  Thanks for adding this! It's helpful to be able to set the format globally. 

I noticed an inconsistency in the Template details page, where the created and modified fields show up in different formats. (The created field does not pick up the format from settings)

vidyanambiar (approved):

PR #1615, opened 2024-02-07
implemented notifications

resoluteCoder:
Jira ticket: https://issues.redhat.com/browse/AAP-8174

Before:
image

After:
image


frontend/awx/resources/notifications/hooks/useNotificationColumns.tsx
```diff
@@ -0,0 +1,42 @@
...
+      });
+    },
+    [pageNavigate]
+  );
+  const nameColumn = useNameColumn({ ...options, onClick: nameClick });
+  const typeColumn = useMemo<ITableColumn<NotificationTemplate>>(
+    () => ({
+      header: t('Type'),
+      type: 'text',
+      value: (notificationTemplate: NotificationTemplate) => {
```
matoval, line 29: Types should be capitalized

frontend/awx/resources/notifications/hooks/useNotificationColumns.tsx
```diff
@@ -0,0 +1,42 @@
...
+        return notificationTemplate.notification_type;
+      },
+      card: 'subtitle',
+      list: 'subtitle',
+    }),
+    [t]
+  );
+  const tableColumns = useMemo<ITableColumn<NotificationTemplate>[]>(
```
matoval, line 37: Do we want a header of Options above the start, success, failure toggles?

matoval (commented):
Looks great, I left a few commands about a few tiny details:

The breadcrumb for sources is missing a link for the sources name(sources/:source_id)

frontend/awx/resources/notifications/hooks/useNotificationColumns.tsx
resoluteCoder: I'm not entirely sure. The old UI does have them but I wasn't able to figure out how to use the switch toggle within a column. It seems that the new UI uses the switch toggle component in the action part of the row. I could be wrong of course :sweat_smile:

frontend/awx/resources/notifications/hooks/useNotificationColumns.tsx
resoluteCoder: Not sure what you mean here. However I did fix where I had `Notification Type` where it seems the rest of the UI uses capital then lower `Notification type`

matoval (approved):

PR #1620, opened 2024-02-08
fixed inventory source bug

resoluteCoder:
Fixed an issue where adding an inventory source would not submit unless certain non-required fields were filled out. 
Added a test case to cover this situation


kialam (changes requested):
Nice catch on the required field for project source. I noticed that other sources are missing required field indicator on the form (credentials). There is a strange behavior where navigating between different source forms marks fields as required when they aren't. See attached:

source-toggle

matoval (approved):
LGTM, ran locally and worked as expected

resoluteCoder:
@kialam Looks like everything is updated. Please let me know if I missed anything. Thanks! :smile:

kialam:
@resoluteCoder thanks for fixing that! The `credentials` field needs to marked as required for certain forms: Google Compute Engine, Azure, VMware, RHS6, OpenStack, RHV, RHAAP, and RHI. I also noticed the forms dropdown is missing the "Template additional groups and hostvars at runtime" option, although selecting that option does not give me a form in the old UI and sending it to the API returns this error:
Screenshot 2024-02-14 at 11 51 29 AM

Maybe @AlexSCorey or @marshmalien would know if this is expected behavior or not?

kialam (approved):
This LGTM. If we need that missing dropdown "template additional groups and hostvars" we can create a separate issue to get it added.

PR #1621, opened 2024-02-08
Disable Save Button, Show Toast on Save

AlexSCorey:
This addresses https://issues.redhat.com/browse/AAP-20536 and https://issues.redhat.com/browse/AAP-20537.

The save button on the workflow visualizer will be disabled until the user makes a change.  Additionally, when the user saves the workflow visualizer they will no longer be navigated to the details view automatically.  Instead, we show a toast to indicate to the user that the save was successful, or not.  This allows the user to save their visualizer as they are working on it.


frontend/awx/resources/templates/WorkflowVisualizer/wizard/NodeAddWizard.tsx
```diff
@@ -162,7 +160,13 @@ export function NodeAddWizard() {
     }
     model.nodes?.push(nodeToCreate);
     controller.fromModel(model, true);
+    controller.getNodeById(nodeToCreate.id)?.setState({ modified: true });
+    controller
+      .getEdgeById(`${state.sourceNode?.getId()}-${nodeToCreate.id}`)
+      ?.setState({ modified: true });
```
marshmalien, line 166: I don't think we need to set `modified: true` to the newly created node or the edge. 

The saveVisualizer hook detects new nodes through the node's name (probably not the most foolproof way to detect new nodes, but we can look into that) and the newly created edge's source node should be set to `modified: true` up on line 148, that should be adequate enough for save to detect which nodes and edges to POST.

marshmalien (approved):
Looks great! 

I did notice a bug where we can't see the success toast on save in fullscreen mode, but we can address it in a follow-up.

PR #1624, opened 2024-02-08
[AAH-2896] Test collections delete actions

skochay330:
This PR adds to the collections e2e tests to verify `delete collection from system` and `delete collection from repository` actions

Jira issue: https://issues.redhat.com/browse/AAH-2896

### DIFF
--- a/cypress/e2e/hub/collections.cy.ts
+++ b/cypress/e2e/hub/collections.cy.ts
@@ -44,9 +44,10 @@ describe('Collections- List View', () => {
   });
 
   it('user can upload and delete collection', () => {
-    const namespace = `upload_namespace_${randomString(3, undefined, { isLowercase: true })}`;
+    const namespace = 'hub_e2e_namespace_' + randomString(3, undefined, { isLowercase: true });
     cy.createNamespace(namespace);
-    const collection = randomString(5, undefined, { isLowercase: true }).replace(/\d/g, '');
+    const collection =
+      'hub_e2e_collection_' + randomString(5, undefined, { isLowercase: true }).replace(/\d/g, '');
     cy.galaxykit(`collection upload ${namespace} ${collection} --skip-upload`).then((result) => {
       cy.navigateTo('hub', Collections.url);
       cy.verifyPageTitle(Collections.title);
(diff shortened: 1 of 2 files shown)


cypress/e2e/hub/collections.cy.ts
```diff
@@ -57,49 +58,106 @@ describe('Collections- List View', () => {
...
   });
 
-  it.skip('user can delete a collection using the list toolbar', () => {});
+  it('user can delete a collection using the list toolbar', () => {
+    const namespace = 'hub_e2e_namespace_' + randomString(3, undefined, { isLowercase: true });
+    const collection =
+      'hub_e2e_collection_' + randomString(5, undefined, { isLowercase: true }).replace(/\d/g, '');
+    cy.createNamespace(namespace);
+    cy.uploadCollection(collection, namespace);
+    cy.approveCollection(collection, namespace, '1.0.0');
```
nixocio, line 67: @skochay330  Could we get the version from the uploadCollection? Not sure if it is always `1.0.0`.

cypress/e2e/hub/collections.cy.ts
```diff
@@ -120,30 +178,55 @@ describe('Collections List- Line Item Kebab Menu', () => {
...
   });
 
   it.skip('user can upload a new version to an existing collection', () => {});
 
-  it.skip('user can delete entire collection from system', () => {});
+  it('user can delete entire collection from system', () => {
+    cy.approveCollection(collection, namespace, '1.0.0');
+    cy.get('[data-cy="table-view"]').click();
+    cy.clickTableRow(collection);
+    cy.wait(400);
```
nixocio, line 219: Could we just do an assertion? To avoid wait, the reason why is because cypress will retry if an assertion fails, but the wait is pure heuristics.


cy.get('StringThatShouldDisplayed')

cypress/e2e/hub/collections.cy.ts
```diff
@@ -120,30 +178,55 @@ describe('Collections List- Line Item Kebab Menu', () => {
...
+    cy.selectDetailsPageKebabAction('delete-entire-collection-from-system');
+  });
 
-  it.skip('user can delete entire collection from repository', () => {});
+  it('user can delete entire collection from repository', () => {
+    cy.approveCollection(collection, namespace, '1.0.0');
+    cy.get('[data-cy="table-view"]').click();
+    cy.clickTableRow(collection);
+    cy.wait(400);
```
nixocio, line 227: Same as above.

cypress/e2e/hub/collections.cy.ts
```diff
@@ -57,49 +58,106 @@ describe('Collections- List View', () => {
...
 
   afterEach(() => {
     if (Cypress.currentTest.title === 'user can deprecate a collection') {
-      cy.undeprecateCollection(thisCollectionName, namespace, repository);
+      cy.undeprecateCollection(collection, namespace, repository);
       cy.galaxykit('task wait all');
     }
-    cy.deleteCollection(thisCollectionName, namespace, repository);
+    if (
+      Cypress.currentTest.title !== 'user can delete entire collection from repository' &&
```
nixocio, line 124: Perhaps we should break down this one in different `describe` ones to avoid this kind of logic.

nixocio (approved):
@skochay330 good job. I left minor suggestions/questions. The tests passed locally.

cypress/e2e/hub/collections.cy.ts
skochay330: I put this in because the details screen reloads a few times before stabilizing, so if I put a string that should be displayed it will work but still not be able to find the next assertion until it finishes reloading- unless there is a way to say 'wait until page finishes loading'?

cypress/e2e/hub/collections.cy.ts
skochay330: If a version is not given during time of galaxykit `uploadCollection` command, it defaults to 1.0.0

skochay330: and our current cypress `uploadCollection` command does not accept a `version` parameter so we should be good

nixocio: ok, I was thinking about something
 ```
cy.uploadCollection(collection, namespace).then((result) => {
``` 
result has the version

File: cypress/e2e/hub/collections.cy.ts [hunk 5]
```diff
@@ -18,9 +18,10 @@ describe.skip('Collections- List View', () => {
   });
 
   it('user can upload and delete collection', () => {
-    const namespace = `upload_namespace_${randomString(3, undefined, { isLowercase: true })}`;
+    const namespace = 'hub_e2e_namespace_' + randomString(3, undefined, { isLowercase: true });
     cy.createNamespace(namespace);
-    const collection = randomString(5, undefined, { isLowercase: true }).replace(/\d/g, '');
+    const collection =
+      'hub_e2e_collection_' + randomString(5, undefined, { isLowercase: true }).replace(/\d/g, '');
     cy.galaxykit(`collection upload ${namespace} ${collection} --skip-upload`).then((result) => {
```
MilanPospisil, line 25: There should be same cy.galaxykit('task wait all').

Every time you are doing some galaxykit command that modifies data (namespace create, collection upload...), you need to call task wait all. Because galaxykit returns when task is queued in the server, not when the task is finish.

But galaxykit task wait all returns when the server has no running tasks, so you can be sure its done. 

Also. you dont have to call then after galaxykit. It could be easily as:
cy.galaxykit('collection upoload...');
cy.galaxkit('task wait all');
cy.navigateTo('hub', Collections.url);
....

it is the same and its more readable.

skochay330:
Bulk collection delete is being removed from the toolbar, so toolbar action tests are also being removed.

PR #1638, opened 2024-02-09
Update to use new PageFormSelectOrganization

jamestalton:
Fixes refresh issue which makes select dialogs disappear.
 Made a new AWX component `PageFormSingleSelectAwxResource`
 Which will make updating all the code easier
 
 ```tsx
 <PageFormSelectOrganization<Credential> name="organization" />
 ```
 
 Where `PageFormSelectOrganization` is implemented as simple as
 ```tsx
 export function PageFormSelectOrganization<
  TFieldValues extends FieldValues = FieldValues,
  TFieldName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>,
>(props: { name: TFieldName; isRequired?: boolean }) {
  const organizationColumns = useOrganizationsColumns({ disableLinks: true });
  const organizationFilters = useOrganizationsFilters();
  return (
    <PageFormSingleSelectAwxResource<Organization, TFieldValues, TFieldName>
      name={props.name}
      id="organization"
      label={t('Organization')}
      placeholder={t('Select organization')}
      queryPlaceholder={t('Loading organizations...')}
      queryErrorText={t('Error loading organizations')}
      isRequired={props.isRequired}
      url={awxAPI`/organizations/`}
      tableColumns={organizationColumns}
      toolbarFilters={organizationFilters}
    />
  );
}
```

Note: A follow up PR will be made for `PageFormMultiSelectAwxResource`

### CODE REVIEW

File: framework/PageFramework.css [hunk 1]
```diff
@@ -86,3 +86,7 @@ tr.selected:hover {
 }
 
 */
+
```
kialam, line 89: We can use styled components here. In `SelectSingleDialog.tsx`:

```
const StyledModal = styled(Modal)`
  .pf-v5-c-modal-box__header {
    padding-bottom: 24px;
  }
`;
```

kialam (changes requested):
Let's use styled components instead of adding custom css to global pf classes. 

I am noticing 2 requests to orgs/ whenever the user clicks the "browse" button. We should try to limit the number of queries we're making to resources. I would expect when clicking "browse" from the select dropdown to not query the API since we already have an initial list of orgs.

framework/PageFramework.css
jamestalton: The PF header does not have the right padding and then the content of the modal scrolls, it looks really bad.
The fix is to update it at the global level so if fixes all modals.

jamestalton:
> Let's use styled components instead of adding custom css to global pf classes.
> 
> I am noticing 2 requests to orgs/ whenever the user clicks the "browse" button. We should try to limit the number of queries we're making to resources. I would expect when clicking "browse" from the select dropdown to not query the API since we already have an initial list of orgs.

One request is for the paginated with out a sort and will a page limit of 200.
The second one is use the awx view, has a page limit and a sort.

PR #1653, opened 2024-02-13
Fix Signature keys tests after updated server settings

ZitaNemeckova:
Remove test that assumes there's no signature key on the server. From now on server should be configured to have signature keys. Updating rest of the test to not mock data as it's no longer needed.


himdel (approved):
LGTM, test failures seem unrelated..

PR #1654, opened 2024-02-13
[AAP-17959] Fix Double Header

prat98:
This PR adds a flag to display or hide the header on the schedules component. The projects section and inventories section, both of which use this component in their tabbed views no use this component without the header to follow UI designs more closely.

Small changes have also been made to the application tokens page to remove the header and fix the trailing slash redirect endpoint issue.

### DIFF
--- a/frontend/awx/administration/applications/ApplicationPage/ApplicationPageTokens.cy.tsx
+++ b/frontend/awx/administration/applications/ApplicationPage/ApplicationPageTokens.cy.tsx
@@ -6,7 +6,7 @@ describe('Application Tokens', () => {
       cy.intercept(
         {
           method: 'GET',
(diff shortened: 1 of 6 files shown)


kialam:
@prat98 please sync up with @daphnemaeve on implementation of this. We now have different ways of displaying the header and I'd rather we use one unified approach. Her work on adding the Templates List tabs to the other resources (Projects, EE, etc) uses the approach of abstracting out the Templates list without the header component. The list is then reused in the top level Templates list component that includes the page header. I prefer that approach to adding another prop to the component.

prat98:
@kialam updated things so I am following the same implementation pattern that @daphnemaeve has been using

kialam (approved):
💯

PR #1670, opened 2024-02-14
Cypress core commands

jamestalton:
These commands are the core commands that are used to interact with the UI.
They are used to interact with the UI in a way that is consistent and reliable.
They check that the element is not disabled or hidden before interacting with it.


akus062381 (approved):
Very nice. Love the organization and clarity.

vishsanghishetty (approved):

PR #1672, opened 2024-02-14
Hub E2E Convert Execution Environment Tests to use cypress request

jamestalton:
Adds
- cy.waitOnHubTask
- cy.createHubRemoteRegistry
- cy.deleteHubRemoteRegistry
- cy.createHubExecutionEnvironment
- cy.deleteHubExecutionEnvironment

This uses the cypress connection credentials and the cy.request underneath.
This allows the developer to see what API calls are made and what the request and responses are instead of using galaxykit which hides what is going on.
This also has the benefits of interacting with the API like EDA and AWX tests.


jerabekjiri (approved):
looks great, thanks :+1:

PR #1680, opened 2024-02-15
Handle user has aborted the request error

jamestalton:
We want to abort requests when a user navigates away from a page.
We do not want those abort errors bubbling up to the user.
This suppresses that abort error.
This also makes development and testing cleaner regarding this abort error.

### DIFF
--- a/frontend/common/crud/useGet.tsx
+++ b/frontend/common/crud/useGet.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useMemo } from 'react';
+import { useCallback, useMemo, useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 import useSWR, { SWRConfiguration } from 'swr';
 import { createRequestError } from './RequestError';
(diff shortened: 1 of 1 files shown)


frontend/common/crud/useGet.tsx
```diff
@@ -12,9 +12,28 @@ export function useGet<T>(
...
+      try {
+        return await getRequest(url).then((value) => {
+          setLastValue(value);
+          return value;
+        });
+      } catch (error) {
+        if (error instanceof Error && error.name === 'AbortError') {
+          // We need to return the last value if the request was aborted
+          // This is because SWR internal workings
+          return lastValue as T;
```
nixocio, line 27: @jamestalton Do we need to return a default value different than `undefined`?

frontend/common/crud/useGet.tsx
```diff
@@ -12,9 +12,28 @@ export function useGet<T>(
   swrConfiguration: SWRConfiguration = {}
 ) {
   const getRequest = useGetRequest<T>();
+  const [lastValue, setLastValue] = useState<T | undefined>();
+  const getRequestWithoutAbortError = useCallback(
+    async (url: string) => {
+      try {
+        return await getRequest(url).then((value) => {
```
nixocio, line 19: What about ?
```js
try {
const value = await getRequest(url)
setLastValue(value)
return value
} catch(error)

```

frontend/common/crud/useGet.tsx
jamestalton: I am seeing a case where two swr calls are going on at the same time.
One aborts, the other then gets undefined.
Not sure what to do about it. returning the last value works, but i am worried it might be out of date.
Still pondering this one.

PR #1685, opened 2024-02-15
Use per request AbortController with cleanup

jamestalton:
Sets up an AbortController per request and when the request is complete, removes the abortController.
This keeps the "User aborted request" error from showing up all the time.


nixocio (approved):
Great job

PR #1688, opened 2024-02-16
[AAP-8116] Hosts List

prat98:
This PR updates the hosts list to be in line with the old UI. Notable changes include adding a smart inventory button which is enabled when the hosts list is filtered. Clicking on this button redirects users to the smart inventory add form which will be pre-filled using search params which were passed from the filtered hosts list.


frontend/awx/resources/hosts/hooks/useHostsActions.tsx
```diff
@@ -1,42 +1,89 @@
...
+  const deleteHosts = useDeleteHosts(onDelete);
+
+  const patchRequest = usePatchRequest<{ enabled: boolean }, AwxHost>();
+
+  const params = useParams<{ id: string; inventory_type: string; host_id: string }>();
+
+  const handleToggleHost: (host: AwxHost, enabled: boolean) => Promise<void> = useCallback(
+    async (host, enabled) => {
+      await patchRequest(awxAPI`/hosts/${host.id.toString()}/`, { enabled });
+      await onToggle();
```
nixocio, line 34: @prat98 as we talked lets try to use the `updateItem` to update the value return by the PATCH request. Instead of the `view.refresh`?

See: https://github.com/ansible/ansible-ui/pull/1796

nixocio (approved):
Good job @prat98, thanks for updating the usage of `updateItem`.

PR #1698, opened 2024-02-19
Update Cypress tests guidelines

nixocio:
Update Cypress tests guidelines

### DIFF
--- a/cypress/CYPRESS.md
+++ b/cypress/CYPRESS.md
@@ -8,15 +8,105 @@ SEE: [README](../README.md) for instructions on setting up and running tests
 
 To get total coverage, run both e2e and component tests.
 
-```
+```bash
 npm run cypress:run
 ```
 
 Open the coverage report
 
-```
+```bash
 npm run coverage
 ```
+
+## End-to-End Tests Guidelines
+
+End-to-End tests for our project are located in the `cypress/e2e` directory. The following guidelines are designed to ensure that our E2E tests are robust, maintainable, and effective.
+
+### General Principles
+
+1. **Test Isolation and Independence**:
+
+   - Each test should be independent and capable of running on its own without relying on the state created by previous tests.
+   - If during deployment certain resources were created, those should not be deleted as part of the test runs.
+   - Avoid tests that require execution in a specific order.
+
+2. **Environment**:
+
+   - Tests should be executable in any environment without depending on the specific state of that environment.
+   - Design tests to be environment-agnostic, ensuring they can run in development and production environments.
+
+3. **Resource Management**:
+   - Each test should create all necessary resources for its execution and clean them up after completion.
+   - This includes data creation, user logins, and any other setup or teardown tasks.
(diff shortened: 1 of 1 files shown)


cypress/CYPRESS.md
```diff
@@ -8,16 +8,106 @@ SEE: [README](../README.md) for instructions on setting up and running tests
...
+       cy.contains('.profile-list', 'John Doe');
+
+       // Asserting on elements that are expected to update
+       cy.get('.profile-status').should('contain', 'Profiles Loaded');
+     });
+   });
+   ```
+
+3. **Test Data**:
+   - Utilize dynamic data creation methods or fixtures to avoid dependency on hard-coded data.
```
akus062381, line 82: fixtures should ideally only be used in component tests. E2E tests should communicate with the live API.

akus062381 (changes requested):

akus062381 (approved):

cypress/CYPRESS.md
```diff
@@ -8,16 +8,106 @@ SEE: [README](../README.md) for instructions on setting up and running tests
...
+
+   - Prefer using data attributes (like `data-cy`) for selecting elements.
+   - Avoid using CSS classes or IDs for element selection, as they can change frequently and are not unique.
+
+2. **Asynchronous Operations**:
+
+   - Properly handle the asynchronous nature of web applications, using Cypress's automatic waiting features or custom waits when necessary.
+   - Avoid to use `cy.wait` as much as possible - mainly with hardcoded wait values, prefer to use `cy.get` with assertions.
+
+   **Example**:
```
elyezer, line 55: I really like that an example is being given, we could probably think about adding more examples related to the other items on the next iteration.

cypress/CYPRESS.md
```diff
@@ -8,16 +8,106 @@ SEE: [README](../README.md) for instructions on setting up and running tests
...
+
+3. **Performance Considerations**:
+   - Keep tests efficient to minimize impact on the development and deployment process.
+
+### Documentation and Collaboration
+
+1. **Documentation of Custom Commands and Utilities**:
+
+   - Document any Cypress [custom commands](./support/core-commands.ts) or utilities developed for the project.
+   - Include usage examples and explanations of the purpose of each command.
```
elyezer, line 103: Maybe link to a good example, or give an example here?

cypress/CYPRESS.md
```diff
@@ -8,16 +8,106 @@ SEE: [README](../README.md) for instructions on setting up and running tests
...
+
+2. **Team Collaboration**:
```
elyezer, line 105: Should we share a Slack channel for further conversations or related topics here?

cypress/CYPRESS.md
nixocio: @elyezer great suggestion, as we move forward we can add more details here.

cypress/CYPRESS.md
nixocio: @elyezer we are lacking this one. At least we do not have a dedicated markdown rendering this information.

PR #1701, opened 2024-02-19
Hub -> Test add and delete execution environment

akus062381:
This PR does the following:

- adds a test to create and delete a Hub execution environment
- fixes failing tests on collections.cy

### DIFF
--- a/cypress/e2e/hub/execution-environments.cy.ts
+++ b/cypress/e2e/hub/execution-environments.cy.ts
@@ -87,7 +87,6 @@ describe('Execution Environment Details tab', () => {
 
   before(() => {
     cy.hubLogin();
-
     cy.galaxykit('registry create', registry, 'https://registry.hub.docker.com/');
     cy.galaxykit('container create', container, 'library/alpine', registry);
   });
--- a/cypress/e2e/hub/execution-environments.cy.ts
+++ b/cypress/e2e/hub/execution-environments.cy.ts
@@ -4,6 +4,7 @@ import { randomString } from '../../../framework/utils/random-string';
 
 describe('Execution Environments', () => {
   let remoteRegistryName = '';
+
   before(() => {
     cy.hubLogin();
   });
@@ -11,6 +12,10 @@ describe('Execution Environments', () => {
   beforeEach(() => {
     remoteRegistryName = `remote_registry_${randomString(3, undefined, { isLowercase: true })}`;
     cy.createRemoteRegistry(remoteRegistryName);
+  });
+
+  afterEach(() => {
+    cy.deleteRemoteRegistry(remoteRegistryName);
   });
 
   it('can render the execution environments page', () => {
--- a/cypress/e2e/hub/execution-environments.cy.ts
+++ b/cypress/e2e/hub/execution-environments.cy.ts
@@ -22,17 +22,16 @@ describe('Execution Environments', () => {
 
   it('should open a new tab and verify correct docs url', () => {
     cy.navigateTo('hub', ExecutionEnvironments.url);
-
     cy.window().then((win) => {
       cy.stub(win, 'open').as('docsTab');
     });
-
     cy.get('[data-cy="push-container-images"]').click();
-
     cy.get('@docsTab').should(
       'be.calledWith',
       'https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/'
     );
+  });
+
   it('can add and delete a new execution environment', () => {
     const remoteRegistryName = generateRemoteRegistryName();
     cy.createRemoteRegistry(remoteRegistryName).then((remoteRegistry: IRemoteRegistry) => {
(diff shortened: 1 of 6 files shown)


cypress/e2e/hub/execution-environments.cy.ts
```diff
@@ -1,14 +1,60 @@
...
+    cy.intercept('DELETE', hubAPI`/v3/plugin/execution-environments/repositories/${eeName}/`).as(
+      'deleted'
+    );
+    cy.get('[data-ouia-component-id="Permanently delete execution environments"]').within(() => {
+      cy.get('[data-ouia-component-id="confirm"]').click();
+      cy.get('[data-ouia-component-id="submit"]').click();
+      cy.clickButton('Close');
+    });
+    cy.wait('@deleted').then((deleted) => {
+      expect(deleted.response?.statusCode).to.eql(202);
```
nixocio, line 56: Verifying that a 202 was returned does not mean that the EE was deleted, but rather that a task was created. Ideally, the status of the action should be verified. For the user, the status of the HTTP code here does not matter. If the API changes and sends back a 204, as it does in other places, the feature will still behave the same way for the use

cypress/e2e/hub/execution-environments.cy.ts
```diff
@@ -1,14 +1,60 @@
...
+    const upstreamName = `upstream_name_${randomString(3, undefined, { isLowercase: true })}`;
     cy.navigateTo('hub', ExecutionEnvironments.url);
     cy.verifyPageTitle(ExecutionEnvironments.title);
+    cy.getByDataCy('add-execution-environment').click();
+    cy.getByDataCy('name').type(eeName);
+    cy.getByDataCy('upstream-name').type(upstreamName);
+    cy.contains('[data-ouia-component-id="menu-select"]', 'Select registry')
+      .click()
+      .then(() => {
+        cy.contains('button[type="button"]', `${remoteRegistryName}`).click();
```
jerabekjiri, line 33: Fails on this line in my local environment, probably because I have a lot of registries in the system.

cypress/e2e/hub/execution-environments.cy.ts
```diff
@@ -1,14 +1,60 @@
...
+  let remoteRegistryName = '';
   before(() => {
     cy.hubLogin();
   });
 
-  it('it should render the execution environments page', () => {
+  beforeEach(() => {
+    remoteRegistryName = `remote_registry_${randomString(3, undefined, { isLowercase: true })}`;
+
+    cy.createRemoteRegistry(remoteRegistryName);
```
jerabekjiri, line 14: If the test is creating remotes and EE, should we do cleanup as well?

cypress/e2e/hub/execution-environments.cy.ts
```diff
@@ -1,14 +1,63 @@
...
+        .then(() => {
+          cy.get(`[data-cy="delete-environment"]`).click();
+        });
+    });
+    cy.get('[data-ouia-component-id="Permanently delete execution environments"]').within(() => {
+      cy.get('[data-ouia-component-id="confirm"]').click();
+      cy.get('[data-ouia-component-id="submit"]').click();
+      cy.clickButton('Close');
+    });
+    cy.contains('h2', 'No results found').should('be.visible');
```
nixocio, line 57: If there are other EES on the system this may fail, right?

cypress/e2e/hub/remote-registries.cy.ts
```diff
@@ -16,34 +14,23 @@ describe('Remote Registry', () => {
     cy.hubLogin();
   });
 
-  after(() => {
```
nixocio, line 19: This change was added to make sure of clean up being successfully done. Leveraging the test signature to delete the resources created by a specific test run. Lets not remove this.

nixocio: Also, this change does not seem to be related to the EE change. Please, add separate PRs to different concerns.

jerabekjiri:
nice fix :+1:  unfortunately, it's now failing locally :grimacing: 
Screenshot from 2024-02-21 11-24-54

I think we should add cy cmd that lets you select the registry, this would always open the modal and search for the value, regardless of how many registries are present in the system, wdyt?

cypress/e2e/hub/execution-environments.cy.ts
```diff
@@ -1,30 +1,102 @@
...
+      cy.getByDataCy('add-execution-environment').click();
+      cy.wait('@registries')
+        .its('response.body.data.length')
+        .then((count) => {
+          cy.getByDataCy('name').type(eeName);
+          cy.getByDataCy('upstream-name').type(upstreamName);
+          cy.contains('[data-ouia-component-id="menu-select"]', 'Select registry')
+            .click()
+            .then(() => {
+              if (count < 11) {
```
nixocio, line 55: It would nice to add a comment explaining this logic here.

nixocio (approved):

cypress/e2e/hub/execution-environments.cy.ts
akus062381: I took your suggestion, thank you

PR #1711, opened 2024-02-20
[AAH-2897] Delete version- collections test

skochay330:
Jira issue: https://issues.redhat.com/browse/AAH-2897 

Tests passing locally:
Screenshot 2024-02-20 at 4 39 57 PM

### DIFF
--- a/cypress/e2e/hub/collections.cy.ts
+++ b/cypress/e2e/hub/collections.cy.ts
@@ -278,7 +278,7 @@ describe('Collections Details View', () => {
     cy.hubLogin();
     cy.createNamespace(namespace);
     cy.galaxykit('task wait all');
-    cy.uploadCollection(collection, namespace);
+    cy.uploadCollection(collection, namespace, '1.0.0');
     cy.galaxykit('task wait all');
     cy.navigateTo('hub', Collections.url);
   });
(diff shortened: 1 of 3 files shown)


cypress/e2e/hub/collections.cy.ts
```diff
@@ -222,9 +222,23 @@ describe('Collections Details View', () => {
...
   it.skip('user can deprecate a collection', () => {});
 
-  it.skip('user can delete version from system', () => {});
+  it('user can delete version from system', () => {
+    cy.approveCollection(collection, namespace, '1.0.0');
+    cy.galaxykit('task wait all');
+    cy.get('[data-cy="table-view"]').click();
+    cy.clickLink(collection);
+    cy.wait(400);
+    cy.selectDetailsPageKebabAction('delete-version-from-system');
```
MilanPospisil, line 231: If you are doing some action in the test, its good to check if the action was successful - thus version is really missing in the system - otherwise we are only testing if the operation is not failing, but not if its doing what its expected. I saw this missing in many tests and it will need fine tuning.

And for testing version deletion, you have to upload and approve the collection with the same name and same namespace, but different version.

cy.galaxykit('collection upload namespace name version');

Then select one of those versions, delete it and check if the version is really missing.

cypress/e2e/hub/collections.cy.ts
```diff
@@ -222,9 +222,23 @@ describe('Collections Details View', () => {
 
   it.skip('user can deprecate a collection', () => {});
 
-  it.skip('user can delete version from system', () => {});
+  it('user can delete version from system', () => {
+    cy.approveCollection(collection, namespace, '1.0.0');
+    cy.galaxykit('task wait all');
+    cy.get('[data-cy="table-view"]').click();
+    cy.clickLink(collection);
```
MilanPospisil, line 229: I would suggest to use cy.filterTableBySingleText(collection) before you call cy.clickLink(collection) that will filter for this particular rows, we have now single texts filter (that not allows for filter by multiple text keys) and this broke some of the tests.

cypress/e2e/hub/collections.cy.ts
```diff
@@ -222,9 +222,23 @@ describe('Collections Details View', () => {
...
+    cy.wait(400);
```
nixocio, line 230: Lets avoid hardcoded values with `cy.wait`, it might lead to failures.

See: https://filiphric.com/waiting-in-cypress-and-how-to-avoid-it

skochay330: @MilanPospisil and I discussed this- unfortunately, due to some framework glitch that causes reloading we have to keep that in for now.

skochay330: Nevermind, I found a way around it by waiting for the correct URL!

PR #1714, opened 2024-02-21
Add repo sync test

ZitaNemeckova:
Test case:
- create a remote (repository without remote cannot be synced. The button is disabled)
- create a repository with remote
- click `Sync`
- verify that alert with correct message is sent
- delete created resources

Related task: https://issues.redhat.com/browse/AAH-2915

### DIFF
--- a/cypress/e2e/hub/repositories.cy.ts
+++ b/cypress/e2e/hub/repositories.cy.ts
@@ -3,16 +3,21 @@ import { randomString } from '../../../framework/utils/random-string';
 import { pulpAPI } from '../../support/formatApiPathForHub';
 
 describe('Repositories', () => {
+  let repositoryName: string;
+  let remoteName: string;
   before(() => {
     cy.hubLogin();
+  });
+  afterEach(() => {
+    cy.galaxykit('-i repository delete ' + repositoryName);
+    cy.galaxykit('-i remote delete ' + remoteName);
   });
   it('should render the repositories page', () => {
     cy.navigateTo('hub', Repositories.url);
     cy.verifyPageTitle(Repositories.title);
   });
   it('should copy CLI to clipboard', () => {
-    const repositoryName =
-      'repositories_repository_' + randomString(6, undefined, { isLowercase: true });
+    repositoryName = 'repositories_repository_' + randomString(6, undefined, { isLowercase: true });
     cy.navigateTo('hub', Repositories.url);
     cy.verifyPageTitle(Repositories.title);
     cy.get('[data-cy="create-repository"]').should('be.visible').click();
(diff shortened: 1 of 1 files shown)


cypress/e2e/hub/repositories.cy.ts
```diff
@@ -23,6 +23,28 @@ describe('Repositories', () => {
...
+    cy.verifyPageTitle(Repositories.title);
+    cy.get('[data-cy="create-repository"]').should('be.visible').click();
+    cy.get('[data-cy="name"]').type(repositoryName);
+    cy.get('#remote').click();
+    cy.get('button').contains(remoteName).click();
+    cy.get('[data-cy="Submit"]').click();
+    cy.get('[data-cy="actions-dropdown"]').click();
+    cy.get('[data-cy="sync-repository"]').click();
+    cy.get('button').contains('Sync').click();
+    cy.get('[data-cy="alert-toaster"]')
```
nixocio, line 42: I think there is a `cy.hasAlert` that would use here.

cypress/e2e/hub/repositories.cy.ts
```diff
@@ -23,6 +23,28 @@ describe('Repositories', () => {
     cy.get('[data-cy="alert-toaster"]').should('be.visible');
     cy.galaxykit('-i repository delete ' + repositoryName);
   });
+  it('should sync repository', () => {
+    const remoteName =
+      'repositories_repository_' + randomString(6, undefined, { isLowercase: true });
+    cy.createRemote(remoteName);
```
nixocio, line 29: @ZitaNemeckova we could do something as you did in other tests...

```suggestion
    cy.createRemote(remoteName).then((remote) => {
    
  ... rest of the test
  
  ```
    
    ;
```

cypress/e2e/hub/repositories.cy.ts
```diff
@@ -23,6 +23,28 @@ describe('Repositories', () => {
...
+    cy.get('button').contains(remoteName).click();
+    cy.get('[data-cy="Submit"]').click();
+    cy.get('[data-cy="actions-dropdown"]').click();
+    cy.get('[data-cy="sync-repository"]').click();
+    cy.get('button').contains('Sync').click();
+    cy.get('[data-cy="alert-toaster"]')
+      .should('be.visible')
+      .should('contain', `Sync started for repository "${repositoryName}".`);
+    cy.galaxykit('-i repository delete ' + repositoryName);
+    cy.galaxykit('-i remote delete ' + remoteName);
```
nixocio, line 46: Good clean up.

nixocio (approved):
@ZitaNemeckova minor suggestions/comments. Good job.

cypress/e2e/hub/repositories.cy.ts
```diff
@@ -23,6 +23,28 @@ describe('Repositories', () => {
     cy.get('[data-cy="alert-toaster"]').should('be.visible');
     cy.galaxykit('-i repository delete ' + repositoryName);
   });
+  it('should sync repository', () => {
+    const remoteName =
```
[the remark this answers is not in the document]
ZitaNemeckova: @akus062381 Based on the name in the URL this screenshot has to be from 'Repositories - revert to repository version'.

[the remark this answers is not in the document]
ZitaNemeckova: I'm not able to reproduce but I have some idea why it's happening. I'll push a fix tomorrow.

akus062381:
I spent a while working on this test to figure out why I am seeing a flaky failure:

The test 'it should copy CLI to clipboard' is flaky about 20% of the time. For a reason unknown to me, sometimes clicking the copy button results in the following toast alert:
image

The toast alert persists in the UI, so when the next test runs, it is still sitting there, and the toast alert for the next test shows up, like so:
image

The test is looking to close the second alert, but it cannot, because it finds 2 "x" buttons (Close buttons) and doesn't know which one to click
So..... the problem is that the copy button doesn't always work.

akus062381 (approved):

PR #1720, opened 2024-02-21
[Workflow] Cleanup UnifiedJobType enum as configuration

marshmalien:
Using the UnifiedJobType enum for both a value and a type was leading to less extensible code. The enum was becoming tightly coupled to different parts of the codebase, making it challenging to add new functionality without impacting existing code.

This PR separates these concerns into UnifiedJobType for type checking and a constant UNIFIED_JOB_TYPE for values.


AlexSCorey (approved):
I like this new approach.  Well done!

PR #1721, opened 2024-02-21
Cypress update the singleSelect and MultiSelect command to support as…

jamestalton:
…ync loading items

### DIFF
--- a/cypress/support/core-commands.ts
+++ b/cypress/support/core-commands.ts
@@ -35,15 +35,27 @@ Cypress.Commands.add('typeByDataCy', (dataCy: string, text: string) => {
   cy.typeBy(`[data-cy="${dataCy}"]`, text);
 });
 
+Cypress.Commands.add('selectLoadAll', () => {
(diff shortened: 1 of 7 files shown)


akus062381 (approved):

cypress/support/core-commands.ts
```diff
@@ -35,15 +35,27 @@ Cypress.Commands.add('typeByDataCy', (dataCy: string, text: string) => {
...
 });
 
+Cypress.Commands.add('selectLoadAll', () => {
+  cy.get('#loading').should('not.exist');
+  cy.get('button').then((buttons) => {
+    for (let i = 0; i <= buttons.length; i++) {
+      const button = buttons[i];
+      if (button?.innerText === 'Load more') {
+        button.click();
+        cy.selectLoadAll();
```
nixocio, line 45: Could this cause an infinite loop?

jamestalton: Load more shows as long as there are more items to load.
Once all items are loaded, it disappears.

PR #1723, opened 2024-02-22
PageFormDataEditor Enhancements

jamestalton:
PageFormDataEditor
- Support loading/saving json/yaml/object formats specified by `format=object`
  - Some forms were trying to handle this during load and save, but having the component do it is cleaner.
  - This is needed by the generated AWX Settings inputs.
- Single instance the Monaco workers
  - Pages such as the AWX settings have a lot of DataEditor inputs and it was loading a ton of workers, killing the browser tab.
- Fixed the wrapping of labels, additional info and data editor settings for all form inputs.

This does change the defaults so that all DataEditors show all supported features, with the ability to turn those off if really needed.

Note
> This adds an initial version of AWX Settings as it was the driving force behind updating the PageFormDataEditor. Follow on PR will finish that work up and add tests.

Screenshot 2024-02-22 at 10 37 37 AM

Screenshot 2024-02-22 at 10 37 52 AM


framework/PageMasonry.tsx
```diff
@@ -0,0 +1,107 @@
...
+      Math.min(
+        props.maxColumns ?? 12,
+        Math.max(Math.floor(entry.contentRect.width / props.minSize), 1)
+      )
+    );
+  });
+  const [span, setSpan] = useState<gridSpans>(12);
+
+  const [sizes, setSizes] = useState<Record<number, number>>({});
+  useLayoutEffect(() => {
```
nixocio, line 28: What about?
```tsx
const spanMap = {
  1: 12,
  2: 6,
  3: 4,
  4: 3,
  5: 2,
  6: 2,
};

useLayoutEffect(() => {
  const newSpan = spanMap[columns] || 1;
  setSpan(newSpan);
}, [columns]);

```

framework/components/DropZone.tsx
```diff
@@ -0,0 +1,65 @@
+import { AlertProps } from '@patternfly/react-core';
+import { ReactNode, useCallback } from 'react';
+import { useDropzone } from 'react-dropzone';
+import { useTranslation } from 'react-i18next';
+import { usePageAlertToaster } from '..';
+
+export function DropZone(props: {
```
nixocio, line 7: Nice break down in smaller components.

framework/components/icons/ExpandIcon.tsx
```diff
@@ -0,0 +1,23 @@
+import { Icon } from '@patternfly/react-core';
```
nixocio, line 1: We might need to wrap all the others icons in this component.

framework/utils/codeEditorUtils.ts
```diff
@@ -1,52 +1,12 @@
...
+    return '';
   }
   const value = JSON.parse(jsonString) as object;
   if (Object.entries(value).length === 0) {
-    return '---\n';
+    return '';
   }
   return jsyaml.dump(value);
 }
-
```
nixocio, line 24: Those functions might be nice to keep around...since we deal with JSON/YAML conversion all the time.

framework/utils/download-file.ts
```diff
@@ -1,9 +1,9 @@
-export function downloadTextFile(name: string, content: string) {
+export function downloadTextFile(name: string, content: string, extension: string = 'txt') {
```
nixocio, line 1: We could get rid of duplicated code here...

What about?

```js

export function downloadFile(name: string, content: string, extension: string) {
  const file = new Blob(content.split('/n'), { type: 'text/plain' });

  const element = document.createElement('a');
  element.href = URL.createObjectURL(file);
  element.download = name + '.' + extension;

  document.body.appendChild(element);
  element.click();
  document.body.removeChild(element);
}

export function downloadTextFile(name: string, content: string) {
  downloadFile(name, content, 'txt');
}

export function downloadCvsFile(name: string, content: string) {
  downloadFile(name, content, 'csv');
}

```

nixocio (approved):
@jamestalton lots of changes. I tried to make sure I was able to follow them. Overall they look good, minor suggestions.

PR #1707, opened 2024-02-20
Workflow visualizer Create and Edit phase

AlexSCorey:
This PR addresses the CREATE phase of the workflow visualizer epic.  That epic can be found here, and the issues related to it are noted with [Create] at the beginning of the issue title.  

@akus062381 @vishsanghishetty

PR #1725, opened 2024-02-22
Verify that promptValues is not undefined before setting original key

marshmalien:
Fix undefined property within workflow node edit 

https://github.com/ansible/ansible-ui/pull/1707#issuecomment-1957937634

image


AlexSCorey (approved):
This looks good.

PR #1728, opened 2024-02-22
Adds tests for workflow vis

AlexSCorey:
This addresses testing gaps found in this pr.  https://github.com/ansible/ansible-ui/pull/1707#issuecomment-1958506987 Ideally, we would merge this work into the workflow-visualizer-edit branch to address the testing gaps.  Then merge the workflow-visualizer-edit branch into main.

### DIFF
--- a/cypress/e2e/awx/resources/workflowVisualizer.cy.ts
+++ b/cypress/e2e/awx/resources/workflowVisualizer.cy.ts
@@ -9,51 +9,57 @@ import { WorkflowNode } from '../../../../frontend/awx/interfaces/WorkflowNode';
 
 describe('Workflow Job templates visualizer', () => {
   let organization: Organization;
-  let inventory: Inventory;
   let project: Project;
+  let inventory: Inventory;
   let inventorySource: InventorySource;
+  let jobTemplate: JobTemplate;
+  let workflowJobTemplate: WorkflowJobTemplate;
 
-  function cleanup() {
-    cy.clickPageAction('delete-template');
-    cy.get('#confirm').click();
-    cy.clickButton(/^Delete template/);
-    cy.verifyPageTitle('Template');
-  }
-  before(() => {
+  before(function () {
+    organization = this.globalOrganization as Organization;
+    project = this.globalProject as Project;
     cy.awxLogin();
-  });
 
-  beforeEach(() => {
-    cy.createAwxOrganization().then((o) => {
-      organization = o;
-      cy.createAwxInventory({ organization: organization.id }).then((i) => {
+    cy.createAwxInventory({ organization: organization.id })
+      .then((i) => {
         inventory = i;
+      })
+      .then(() => {
+        cy.createAwxInventorySource(inventory, project).then((invSrc) => {
+          inventorySource = invSrc;
+        });
+        cy.createAwxJobTemplate({
+          organization: organization.id,
+          project: project.id,
+          inventory: inventory.id,
+        }).then((jt) => (jobTemplate = jt));
       });
-    });
   });
-
-  afterEach(() => {
-    cy.deleteAwxInventory(inventory, { failOnStatusCode: false });
-    cy.deleteAwxOrganization(organization, { failOnStatusCode: false });
-  });
-
-  it('should render a workflow visualizer view with multiple nodes present', () => {
+  beforeEach(function () {
     cy.createAwxWorkflowJobTemplate({
       organization: organization.id,
       inventory: inventory.id,
-    }).then((workflowJobTemplate) => {
-      cy.renderWorkflowVisualizerNodesFromFixtureFile(
-        `${workflowJobTemplate.name}`,
-        'wf_vis_testing_A.json'
-      );
-      cy.get('[class*="66-node-label"]')
-        .should('exist')
-        .should('contain', 'Cleanup Activity Stream');
-      cy.get('[class*="43-node-label"]').should('exist').should('contain', 'bar');
-      cy.get('[class*="42-node-label"]').should('exist').should('contain', '1');
-      cy.get('[class*="41-node-label"]').should('exist').should('contain', 'Demo Project');
-      cy.deleteAwxWorkflowJobTemplate(workflowJobTemplate);
-    });
+    }).then((wfjt) => (workflowJobTemplate = wfjt));
+  });
+  afterEach(() => {
+    cy.deleteAwxWorkflowJobTemplate(workflowJobTemplate, { failOnStatusCode: false });
+  });
+  after(function () {
+    cy.deleteAwxInventory(inventory, { failOnStatusCode: false });
+    cy.deleteAwxInventorySource(inventorySource, { failOnStatusCode: false });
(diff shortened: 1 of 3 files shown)


cypress/e2e/awx/resources/workflowVisualizer.cy.ts
```diff
@@ -166,54 +141,274 @@ describe('Workflow Job templates visualizer', () => {
...
+          cy.createAwxWorkflowVisualizerManagementNode(workflowJobTemplate, 1)
+            .then((managementNode) => {
+              cy.createWorkflowJTSuccessNodeLink(projectNode, inventorySourceNode);
+              cy.createWorkflowJTAlwaysNodeLink(inventorySourceNode, managementNode);
+            })
+            .then(() => {
+              cy.visit(`/templates/workflow_job_template/${workflowJobTemplate?.id}/visualizer`);
+              cy.get(`g[data-id=${projectNode.id}] .pf-topology__node__action-icon`).click({
+                force: true,
+              });
```
akus062381, line 182: Can you add a baseline assertion here to show that the initial value of the node is type project? that provides a comparison for later when the node is edited.

cypress/e2e/awx/resources/workflowVisualizer.cy.ts
```diff
@@ -166,54 +141,274 @@ describe('Workflow Job templates visualizer', () => {
...
+    cy.clickButton('Finish');
+    cy.get('g[data-id="1-unsavedNode"]').should('have.text', 'Test NodeProject Node');
+    cy.clickButton('Save');
+    cy.get('[data-cy="alert-toaster"]').should('be.visible');
+    cy.get('button[data-cy="workflow-visualizer-toolbar-close"]').click();
+
+    cy.get('[data-cy="page-title"]').should('have.text', `${workflowJobTemplate.name}`);
+  });
+  it('Adds a new node specifically linked to an already existing node.', function () {
+    let projectNode: WorkflowNode;
```
akus062381, line 237: There is a test on line 114 of this file: 'Adds a new node linked to an existing node with always status, and save the visualizer.' 
These two tests are very similar. Perhaps we can combine them and kill two birds with one stone? :-)

cypress/e2e/awx/resources/workflowVisualizer.cy.ts
```diff
@@ -166,54 +141,274 @@ describe('Workflow Job templates visualizer', () => {
...
-        .should('contain', 'Workflow Visualizer')
-        .should('be.visible');
       cy.get('[data-cy="wf-vzr-name"]')
         .should('contain', `${workflowJobTemplate.name}`)
         .should('be.visible');
-      //remove-node, add-node-and-link, add-link, add-link, edit-node
       cy.removeAllNodesFromVisualizerToolbar();
       cy.contains('button', 'Save').click();
       cy.deleteAwxWorkflowJobTemplate(workflowJobTemplate);
     });
```
akus062381, line 167: Can we add an assertion after line 165 (215) that asserts that the nodes have been deleted? And since the test indicates that the user will delete the template, can we perform the deletion in the UI instead of through an HTTP request? either that or remove the piece of the test description that indicates that the test will be deleted?

cypress/e2e/awx/resources/workflowVisualizer.cy.ts
```diff
@@ -166,54 +141,274 @@ describe('Workflow Job templates visualizer', () => {
...
+        cy.get('[data-cy="alert-toaster"]').should(
+          'have.text',
+          'Success alert:Successfully saved workflow visualizer'
+        );
+        cy.get('button[data-cy="workflow-visualizer-toolbar-close"]').click();
+        cy.get('[data-cy="page-title"]').should('have.text', `${workflowJobTemplate.name}`);
+      });
+  });
+  it.skip('Can manually delete all nodes, save the visualizer, then add new nodes, and successfully save again.', function () {
+    let projectNode: WorkflowNode;
```
akus062381, line 274: Any reason why this test is skipped?

cypress/e2e/awx/resources/workflowVisualizer.cy.ts
```diff
@@ -166,54 +141,274 @@ describe('Workflow Job templates visualizer', () => {
...
+          'have.text',
+          'Success alert:Successfully saved workflow visualizer'
+        );
+        cy.get('button[data-cy="workflow-visualizer-toolbar-close"]').click();
+        cy.get('[data-cy="page-title"]').should('have.text', `${workflowJobTemplate.name}`);
+      });
+    });
+  });
+  it.skip('Can remove all existing nodes on a visualizer using the button in the toolbar kebab, save the visualizer, then add 2 new nodes and save the visualizer again', function () {
+    cy.createAwxWorkflowVisualizerProjectNode(workflowJobTemplate, project).then((projectNode) => {
```
akus062381, line 372: Just wondering why this test is skipped?

akus062381 (changes requested):

cypress/e2e/awx/resources/workflowVisualizer.cy.ts
```diff
@@ -166,54 +141,274 @@ describe('Workflow Job templates visualizer', () => {
         cy.get(`g[data-id=${approvalNode.id}-3-unsavedNode]`).should('have.text', 'Run always');
         cy.clickButton('Save');
         cy.get('[data-cy="alert-toaster"]').should('be.visible');
-        cy.get('[data-cy="wf-vzr-name"]').should('have.text', `${workflowJobTemplate.name}`);
```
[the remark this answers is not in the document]
AlexSCorey: If you are talking about [this line](https://github.com/ansible/ansible-ui/pull/1728/files#diff-429f33cf321187f2a244661cd073b48fc5c2561229c14079277ec8e9fbd4440cL159), then no.  It seems to me that the component that is used should be able to handle that case....and the custom cypress command should be able to handle it also.  This particular line should not need to care about how many resources there are available to choose from.

cypress/e2e/awx/resources/workflowVisualizer.cy.ts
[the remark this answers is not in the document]
AlexSCorey: We have other assertions in this file that ensure the workflow template is deleted.  In this case, I really only care about whether I can remove the nodes, and deleting the template is more of a clean up step here.

cypress/e2e/awx/resources/workflowVisualizer.cy.ts
AlexSCorey: It fails currently, but should pass before we merge into main

cypress/e2e/awx/resources/workflowVisualizer.cy.ts
AlexSCorey: It fails currently, but should pass before we merge into main

akus062381 (approved):
thanks for your hard work!

vishsanghishetty:
awesome work 🚀  @AlexSCorey thanks for handling the tests, appreciate it!

PR #1735, opened 2024-02-26
AWX  Hide empty job and template expanded row columns

jamestalton:
Columns that do not have a value are not shown in the expanded row.

Screenshot 2024-02-26 at 2 04 51 PM


resoluteCoder (approved):
lgtm, pulled and tested locally

vidyanambiar (commented):
Execution environment shows up as empty
image

PR #1742, opened 2024-02-27
Cypress HUB API commands

jamestalton:
Screenshot 2024-02-27 at 11 59 31 AM

### DIFF
--- a/cypress/support/hub-commands.ts
+++ b/cypress/support/hub-commands.ts
@@ -1,17 +1,111 @@
 /* eslint-disable @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access */
+import { SetRequired } from 'type-fest';
 import { randomString } from '../../framework/utils/random-string';
 import { Role } from '../../frontend/hub/access/roles/Role';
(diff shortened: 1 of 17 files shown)


cypress/support/hub-commands.ts
```diff
@@ -1,18 +1,112 @@
...
+Cypress.Commands.add('hubRequest', (options: HubRequestOptions) => {
+  cy.getCookie('csrftoken').then((cookie) =>
+    cy
+      .request({
+        ...options,
+        headers: { 'X-CSRFToken': cookie?.value, Referer: Cypress.config().baseUrl },
+      })
+      .then((response) => {
+        switch (response.status) {
+          case 202: // Accepted
```
nixocio, line 36: @jamestalton a couple of questions here:

Do we want to return the task in a few cases?  Perhaps, we could add a option to return the response without waiting for the task to be completed. Mainly thinking about the tasks tests.

We could return the response not matter what the status code?

akus062381 (approved):
I ran the spec files multiple times and everything is looking good!

PR #1743, opened 2024-02-27
Reduce unnecessary renders in workflow save hook

marshmalien:
The workflow useSave hook was triggering unnecessary re-renders. I encapsulated the hook and its associated hooks within a useCallback to optimize and minimize render cycles.


AlexSCorey (approved):
Just found 1 issue with a console warning.  Marly and I spoke offline about it.

PR #1747, opened 2024-02-28
HUB Approvals E2E

jamestalton:


akus062381 (changes requested):
I would like to avoid having one big test that asserts lots of different functionality in it. Let's keep the tests separate please. Thanks!

jamestalton:
> I would like to avoid having one big test that asserts lots of different functionality in it. Let's keep the tests separate please. Thanks!

This creates one collection, which is the slow part of the tests.

The previous tests would

- Verify Approve
- Verify Reject
- Verify Approve then reject (Which is testing the approve and rejects again)
- Verify Reject then approve(Which is testing the approve and rejects again)

This instead checks all that by approve, then reject, then approve.
Not really any different than what the original tests tested, just tests it more efficiently.

akus062381:
@jamestalton I understand that your goal is speed and optimization of the tests. I am just trying to think about how we can best verify the UI functionality the way a user is likely to use it, while also optimizing for speed? 

What if we have two separate tests:
1. can upload a new collection, approve,  view import logs
2. can upload a new collection, reject, and view import logs

I feel like this is a bit more in line with how a user would be likely to use this feature...? 
Thanks for your consideration!

akus062381 (approved):
This looks good, thanks for all your hard work on it!

PR #1749, opened 2024-02-28
HUB E2E API Cleanup Old E2E Resources

jamestalton:


nixocio (commented):
@jamestalton I really appreciate your efforts to improve the tests in general.  

I did think about this one, and my thoughts:

Ideally clean ups like that would run before the tests runs. Because this one might change the order of the run once more tests are added, and this file modified. We can not rely on this one.

We should strive to clean up the resources created by the tests at the end of the test run.

jamestalton:
> We should strive to clean up the resources created by the tests at the end of the test run.

I agree, but there are cases, especially when we are writing tests that resources get created and not cleaned up.
This is just general maintenance cleanup for those cases.

nixocio (approved):
@jamestalton we talked about moving those commands to a setup before the tests run, but it will require more changes, I am approving this one as it is, then we can revisit this one when necessary.

PR #1754, opened 2024-02-28
Dedupe old workflow nodes after fetching new nodes

marshmalien:
Fix addresses duplicate orphaned nodes after save

Screenshot 2024-02-28 at 10 37 53 AM


AlexSCorey (approved):
Marly and I spoke offline about this.  This code looks good, though we should probably file a bug with PF as well.

PR #1756, opened 2024-02-29
[AAH-3103] Add useful links

nixocio:
Refactor Namespace details page

Modify Namespace form

Add PageDetailKeyValue component

Details Page

image

Create

image

### DIFF
--- a/framework/PageDetails/PageDetailKeyValueList.tsx
+++ b/framework/PageDetails/PageDetailKeyValueList.tsx
@@ -1,0 +1,65 @@
+import {
+  DescriptionList,
+  DescriptionListDescription,
+  DescriptionListGroup,
+  DescriptionListTerm,
+  Text,
+  TextContent,
+} from '@patternfly/react-core';
+
+type KeyValue = {
+  key: string;
+  value: string;
+};
+
+type KeyValueProps = {
+  keyValue: KeyValue[];
+  title: string;
+  keyColumn: string;
+  valueColumn: string;
+  renderValue?: (item: KeyValue) => JSX.Element;
+};
+
+export function PageDetailKeyValueList(props: KeyValueProps) {
+  const { keyValue, title, keyColumn, valueColumn, renderValue } = props;
+
+  const defaultRenderValue = (item: KeyValue) => <div data-cy={`item-value`}>{item.value}</div>;
+
+  const renderFunction = renderValue || defaultRenderValue;
+
+  return (
+    <>
+      <DescriptionListGroup>
+        <TextContent>
+          <Text
+            data-cy={'key-value-list-title'}
+            style={{ fontSize: 'medium-text', fontWeight: 'bold' }}
+          >
+            {title}
+          </Text>
+        </TextContent>
+        <DescriptionList isHorizontal>
+          <DescriptionListGroup>
+            <DescriptionListTerm>{keyColumn}</DescriptionListTerm>
+            <DescriptionListTerm>{valueColumn}</DescriptionListTerm>
+          </DescriptionListGroup>
+          {keyValue.map((item, index) => (
+            <DescriptionListGroup key={index}>
+              <DescriptionListTerm
+                style={{ fontWeight: 'normal', whiteSpace: 'pre-wrap', overflow: 'hidden' }}
+                data-cy={`item-key-${index}`}
+              >
+                {item.key}
+              </DescriptionListTerm>
+              <DescriptionListDescription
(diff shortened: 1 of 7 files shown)


frontend/hub/namespaces/PageInputDynamicFields.tsx
```diff
@@ -0,0 +1,67 @@
+import { Button, Flex, FlexItem, TextInput } from '@patternfly/react-core';
+import { PlusCircleIcon, TrashIcon } from '@patternfly/react-icons';
+import { useFieldArray, useFormContext, Controller } from 'react-hook-form';
+import { useTranslation } from 'react-i18next';
+import { PageFormSection } from '../../../framework/PageForm/Utils/PageFormSection';
+
+export function PageInputDynamicFields() {
```
nixocio, line 7: This name will be updated. Just to have an idea what direction this work was going.

ShaiahWren (commented):
Nice work @nixocio!
A couple surprises I encountered. Do you see these?
1. Surprising red when pasting a LINK. Why is that? It has no message and no network error. Must be a ui issue.
Screenshot 2024-03-04 at 10 53 35 AM

2. On full screen list view, the link is clipped off. The formatting is also surprising. We can fix by putting it in a new section.I will look for a solution today. :)

Screenshot 2024-03-04 at 10 56 01 AM

framework/PageDetails/PageDetailKeyValueList.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <DescriptionListTerm data-cy={'key-value-list-title'}>{title}</DescriptionListTerm>
+      <DescriptionList isHorizontal style={{ paddingLeft: '20px' }}>
+        <DescriptionListGroup>
+          <DescriptionListTerm>{keyColumn}</DescriptionListTerm>
+          <DescriptionListDescription>{valueColumn}</DescriptionListDescription>
+        </DescriptionListGroup>
+        {keyValue.map((item, index) => (
+          <DescriptionListGroup key={index}>
+            <DescriptionListDescription
+              style={{ whiteSpace: 'nowrap', textOverflow: 'ellipsis' }}
```
jamestalton, line 39: Would we want this to use ellipsis vs wrapping?

framework/PageDetails/PageDetailKeyValueList.tsx
```diff
@@ -0,0 +1,54 @@
...
+              data-cy={`item-key-${index}`}
+            >
+              {item.key}
+            </DescriptionListDescription>
+            <DescriptionListDescription
+              style={{ whiteSpace: 'nowrap', overflow: 'hidden', textOverflow: 'ellipsis' }}
```
jamestalton, line 45: Would we want this to use ellipsis vs wrapping?

framework/PageDetails/PageDetailKeyValueList.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const { keyValue, title, keyColumn, valueColumn, renderValue } = props;
+
+  const defaultRenderValue = (item: KeyValue) => <div data-cy={`item-value`}>{item.value}</div>;
+
+  const renderFunction = renderValue || defaultRenderValue;
+
+  return (
+    <>
+      <DescriptionListTerm data-cy={'key-value-list-title'}>{title}</DescriptionListTerm>
+      <DescriptionList isHorizontal style={{ paddingLeft: '20px' }}>
```
jamestalton, line 31: padding left here is not optimal at small screens. might be better to have a section with heading and no left padding.

jamestalton (approved):

nixocio:
> Nice work @nixocio! A couple surprises I encountered. Do you see these?
> 
>     1. Surprising red when pasting a LINK. Why is that? It has no message and no network error. Must be a ui issue.
> 
> 
> Screenshot 2024-03-04 at 10 53 35 AM
> 
>     2. On full screen list view, the link is clipped off. The formatting is also surprising. We can fix by putting it in a new section.I will look for a solution today. :)
> 
> 
> Screenshot 2024-03-04 at 10 56 01 AM

@ShaiahWren thanks for your review, great findings, please take another look.

ShaiahWren:
Nice work @nixocio , I'm still seeing some ux inconsistencies with the other items on the detail page. I think the "Useful links" probably shouldn't be an h3 because that indicates a tertiary view for the screen reader and I'm not sure that's appropriate here. If the intent is to get a smaller text, we should use font-size=14px to match the other headers (name, desc, company, etc). 
Screenshot 2024-03-06 at 9 49 51 AM

PR #1757, opened 2024-02-29
Advanced search filtering

kialam:
This PR adds a hook to handle dynamically rendering the list page toolbar filter based on the OPTIONS response from the Controller API. It can also handle rendering any keys with pre-filled values.
```js
interface DynamicToolbarFiltersProps {
  /** API endpoint for the options that generated the filters */
  optionsPath: string;

  /** A list of keys to order the filters toolbar */
  preSortedKeys?: string[];

  /** A list of keys to pre-populate dropdown values. Note: knownAwxFilterKeys adds some keys that require quering specific endpoints */
  preFilledValueKeys?: string[];

  /** Additional filters in addition to the dynamic filters */
  additionalFilters?: IToolbarFilter[];
}
```
Sample usage:
```js
export function useProjectsFilters() {
  const createdByToolbarFilter = useCreatedByToolbarFilter();
  const modifiedByToolbarFilter = useModifiedByToolbarFilter();
  const toolbarFilters = useDynamicToolbarFilters<Project>({
    optionsPath: 'projects',
    preSortedKeys: ['name', 'description', 'id', 'scm_type', 'created-by', 'modified-by'],
    preFilledValueKeys: ['name', 'id'],
    additionalFilters: [createdByToolbarFilter, modifiedByToolbarFilter],
  });
  return toolbarFilters;
}
```

### DIFF
--- a/cypress/e2e/awx/views/jobs.cy.ts
+++ b/cypress/e2e/awx/views/jobs.cy.ts
@@ -50,16 +50,16 @@ describe('jobs', () => {
     cy.verifyPageTitle('Jobs');
     const jobId = jobList.id ? jobList.id.toString() : '';
     const jobName = jobList.name ? jobList.name : '';
-    cy.filterTableByTypeAndText('ID', jobId);
+    cy.filterTableByMultiSelect('id', [jobId]);
     cy.contains(jobName);
-    cy.clickButton(/^Clear all filters$/);
+    cy.clearAllFilters();
   });
 
   it('relaunches job and navigates to job output', () => {
     cy.navigateTo('awx', 'jobs');
     const jobId = jobList.id ? jobList.id.toString() : '';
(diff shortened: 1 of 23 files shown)


akus062381:
I saw the following behavior:

I changed the filter to 'unified job template'
I saw the following:
image

So in the text box, I added part of a name of one of the job templates
The API returned the following:
image


Maybe we can add something to let the user know that 'unified job template' is expecting an ID?

akus062381:
I noticed the following behavior:

When I filtered the results, and then clicked on the sort button to reveal the separate sorting filter, and when I sorted the list a few times, I noticed that the "clear all filters" buttons disappeared:
image

I also noticed that, after those buttons disappeared, the kebab menu showed blank:
image

jamestalton (approved):

PR #1758, opened 2024-02-29
HUB Repositories E2E Tests

jamestalton:
Removes calls to `cy.galaxykit('task wait all')`.
Reduced the number of collections the tests were creating which also speeds up the tests.

Repositories tests now take < 2 min to run
Screenshot 2024-02-29 at 12 49 54 PM
vs the 4 min it used to take
Screenshot 2024-02-29 at 12 49 34 PM


cypress/e2e/hub/repositories.cy.ts
```diff
@@ -1,299 +1,224 @@
...
-      cy.galaxykit(`collection upload ${namespace} ${collection}`);
-      cy.galaxykit('task wait all');
+  it('should copy CLI to clipboard', () => {
+    cy.createHubRepository().then((repository) => {
+      cy.filterTableBySingleText(repository.name, true);
+      cy.clickTableRow(repository.name, false);
+      cy.verifyPageTitle(repository.name);
+      cy.clickPageAction('copy-cli-configuration');
+      cy.get('[data-cy="alert-toaster"]').should('be.visible');
+      cy.deleteHubRepository(repository);
```
ZitaNemeckova, line 69: Sarah ran into issues when running test locally that is deleting repository via API when we are at that repo's detail page.

ZitaNemeckova:
Adding Milan as he wrote half of the tests. I'd like to have a look before it gets merged.

framework/utils/random-string.ts
```diff
@@ -21,3 +21,7 @@ export function randomString(
   }
   return text;
 }
+
+export function randomLowercaseString(length: number): string {
```
ZitaNemeckova, line 25: We will lose all naming conventions we agreed on. It would be nice to leave the naming to Hub. It gives us more info when a test goes wrong. Plus we have objects that cannot be named just whatever.

cypress/e2e/hub/repositories.cy.ts
```diff
@@ -1,299 +1,224 @@
...
-    collections.forEach((collection) => {
-      cy.galaxykit(`collection upload ${namespace} ${collection}`);
-      cy.galaxykit('task wait all');
+  it('should copy CLI to clipboard', () => {
+    cy.createHubRepository().then((repository) => {
+      cy.filterTableBySingleText(repository.name, true);
+      cy.clickTableRow(repository.name, false);
+      cy.verifyPageTitle(repository.name);
+      cy.clickPageAction('copy-cli-configuration');
+      cy.get('[data-cy="alert-toaster"]').should('be.visible');
```
ZitaNemeckova, line 68: Missing closing the alert.

ZitaNemeckova: The alert should be closed as it can stay in the UI even when a page changes so it can negatively influence following tests by covering toolbar kebab.

cypress/e2e/hub/repositories.cy.ts
ZitaNemeckova: Delete should be happening in the afterEach so there's no failure based on UI trying to reload a detail page of delete resource.

jamestalton: There is a test that creates and deletes a repository so putting the delete in an afterEach would fail in that case

framework/utils/random-string.ts
jamestalton: It would be nice to use consistent e2e test resource naming across the platform.
Any resources starting with e2e_, we should be able to assume they were created with E2E testing.

jamestalton:
Updated tests to use the newer more robust functions.
Added before and after sections for creating and cleaning up resources.

cypress/e2e/hub/repositories.cy.ts
jamestalton: fixed

framework/utils/random-string.ts
ZitaNemeckova: Please do not throw away naming we had there in favour of an absolutely random string. Hub agreed to do naming in specific way that helps us to find out where the object came from to the `describe`. We want something like `hub_e2e_[test_name]_[describe_name]_[random string]`. We can switch `hub` and `e2e` anytime there's a consensus on naming on the platform level that is documented. We also need to start the name of some objects with specific prefix like roles need to start with `galaxy.`. Framework should allow for flexible naming instead of cutting individual projects from it.

jamestalton: > Framework should allow for flexible naming instead of cutting individual projects from it.
If you mean project as HUB, AWX, or EDA, we are a Platform UI and a Platform UI Team.

Naming resources with really long names makes the tests run slower as the name has to be typed in by cypress.

Trying to put the full test name and describe name in the resource seems like it could be problematic. In addition if a test fails, you can see the name right in the failure.

Having hub_ on the beginning of the name is kind of redundant. All hub resources are hub resources.

Having a common pattern for naming resources makes it consistent across testing in the platform. Makes it so that you can goto any resource view and search for "starts with" e2e. This also allows us to have scripts to cleanup e2e resources that are left over from testing. We should be cleaning up resources in tests, but if the test fails, it does not cleanup.

The current hub tests are not consistent with naming of resources. Having a function to generate that name makes it more consistent. `randomE2Ename()`. If we want to enhance that function to support a postfix on the end of that name, i'm all for it.

The support cypress functions are designed to be fully flexible allowing overriding any properties of the resource.
```
createHubRemoteRegistry() // uses a default generated name
```
or
```
createHubRemoteRegistry({
  name: "some_name_I_want_to_use"
})
```

This gives us common patterns for working with the custom cypress commands for working with resources. This allows developers to easily work on tests across the whole platform.

nixocio (approved):

PR #1761, opened 2024-03-01
[AAP-20541] Remove extra header from the Role details page. Update scrolling.

lgalis:
Remove extra header from the Role details page.
 Update scrolling.

Screenshot from 2024-03-01 15-15-08


tiyiprh (approved):
LGTM! TY

PR #1777, opened 2024-03-05
PageFormTextInput DateTime Local support

jamestalton:
Screenshot 2024-03-05 at 2 55 37 PM

Screenshot 2024-03-05 at 2 55 54 PM


framework/PageForm/Inputs/PageFormTextInput.tsx
```diff
@@ -233,8 +233,23 @@ export function PageFormTextInput<
             : error?.message
           : undefined;
 
+        let massagedValue: string = value;
```
nixocio, line 236: what about parsedValue?

nixocio (approved):

PR #1778, opened 2024-03-05
Workflow Job Ouput 

AlexSCorey:
This addresses:
https://issues.redhat.com/browse/AAP-19089
https://issues.redhat.com/browse/AAP-19090
https://issues.redhat.com/browse/AAP-21460
https://issues.redhat.com/browse/AAP-21520

It adds status, and elapsed time decorators to the nodes in the job output as the status of the node is update via websockets.  

websockets

### DIFF
--- a/cypress/e2e/awx/resources/workflowVisualizer.cy.ts
+++ b/cypress/e2e/awx/resources/workflowVisualizer.cy.ts
@@ -7,7 +7,7 @@ import { Project } from '../../../../frontend/awx/interfaces/Project';
 import { WorkflowJobTemplate } from '../../../../frontend/awx/interfaces/WorkflowJobTemplate';
 import { WorkflowNode } from '../../../../frontend/awx/interfaces/WorkflowNode';
 
-describe('Workflow Job templates visualizer', () => {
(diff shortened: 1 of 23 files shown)


AlexSCorey:
@akus062381 @vishsanghishetty

akus062381:
Testing currently, and I am running a wfjt, the decorate for Debug 50 isn't showing
image

akus062381:
Question- while the different nodes are in the process of running, should we see the time counting forward on each individual node? or should we see "elapsed time" as shown here (with the elapsed time only being displayed once the workflow is complete?)
image

akus062381:
I saw that while a wfjt is in the process of running, switching to the Details view of the job and looking at Duration- the time first shows as '0 seconds' before switching over to the actual progressed time

akus062381:
I noticed that sometimes, all nodes will show that they are finished running, but the job time counter continues ticking, and the status of the job shows Running until I clicked refresh on the browser

image

akus062381 (changes requested):
I made some comments on the PR about little bugs I was seeing.

marshmalien (approved):

PR #1790, opened 2024-03-07
Copy Collection to repositories fix

MilanPospisil:
Fixing broken copy collection to repositories.

This PR also adds refresh table after modal is closed.

### DIFF
--- a/frontend/hub/collections/hooks/useCopyToRepository.tsx
+++ b/frontend/hub/collections/hooks/useCopyToRepository.tsx
@@ -18,7 +18,7 @@ import { PageTable } from './../../../../framework/PageTable/PageTable';
 import { useGetRequest } from './../../../common/crud/useGet';
 import { TFunction } from 'i18next';
 
-export function useCopyToRepository() {
+export function useCopyToRepository(refresh: (collections: CollectionVersionSearch[]) => void) {
   const [_, setDialog] = usePageDialog();
   const onClose = useCallback(() => setDialog(undefined), [setDialog]);
   const context = useHubContext();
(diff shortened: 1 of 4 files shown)


frontend/hub/collections/hooks/useCopyToRepository.tsx
```diff
@@ -233,7 +236,7 @@ export async function copyToRepositoryAction(
   }
 
   const pipeline = collection.repository?.pulp_labels?.pipeline;
-  if (!(pipeline === 'staging' || pipeline === 'rejected')) {
+  if (operation === 'approve' && !(pipeline === 'staging' || pipeline === 'rejected')) {
```
MilanPospisil, line 239: This caused the problems, the approvals have this condition to allow only approvals from staging or rejected, but it was failing in copy (which is allowed in UI only from approved repos).

akus062381:
Thanks for adding this!
I am testing your PR now, and see the following behavior:

1. Upload a collection
2. About 10 repositories already exist (if not add about 10 repositories)
3. Copy the collection into all 10 repositories
4. See the new list of collections in all the repositories
5. Try to delete all those collections, including the one you originally uploaded
6. See the following in the UI:
image

akus062381 (changes requested):
Can we find a way to handle bulk deletion of multiple copied collections?

MilanPospisil:
> Can we find a way to handle bulk deletion of multiple copied collections?

Lets solve it in separate PR, this is unrelated problem and I know why it happends.

akus062381 (approved):

PR #1794, opened 2024-03-07
AWX E2E Organization Tests Update

jamestalton:
Organization tests keep randomly hanging forever.
Could not reproduce.
Updated the tests to use the newer more robust commands.
Hoping that this fixes the random hang.


akus062381 (approved):
Ran the tests 4 times, all passing! thanks!

PR #1796, opened 2024-03-07
AwxView updateItem support

jamestalton:


nixocio (approved):
This will allow to update certain item without have to refresh the whole cache.

Thank you @jamestalton

PR #1808, opened 2024-03-08
PageTable support defaultExpandedRows

jamestalton:


vidyanambiar (commented):
- The expanded sections appear to be empty. I tested this in the Jobs and Templates lists.
- I manually set `defaultExpandedRows` in the `PageTable` component under `JobsList`. The rows are shown to be expanded but the expanded state is empty. 
Also, can we add the "expand/collapse all" button?

image

PR #1813, opened 2024-03-11
AWX hosts add edit form

MilanPospisil:
Adds create and edit form for host list.

It uses existing forms in inventories - hosts. It only modifies them so:

it can work with new URL, when inventory id params is missing, the form knows it is accessed from hosts list. It will then either display inventory (in edit mode), or it will offer inventory dropdown (in create mode). 

image

After clicking Browse:
image

image

### DIFF
--- a/frontend/awx/resources/inventories/inventoryHostsPage/InventoryHostForm.tsx
+++ b/frontend/awx/resources/inventories/inventoryHostsPage/InventoryHostForm.tsx
@@ -242,7 +242,7 @@ export function EditHost() {
     name: host.name,
     description: host.description,
     variables: host.variables,
-    inventory: { name: host.summary_fields?.inventory.name },
+    inventory: { name: host.summary_fields?.inventory?.name },
   };
 
   let breadcrumbs: ICatalogBreadcrumb[] = [];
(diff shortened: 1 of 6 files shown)


frontend/awx/resources/inventories/inventoryHostsPage/InventoryHostForm.tsx
```diff
@@ -133,31 +158,73 @@ export function CreateHost() {
...
+    group_id: string;
+  }>();
+
+  let id = params.id;
+  let host_id = params.host_id;
+
+  let inventory_host = false;
+
+  if (!host_id && mode === 'edit') {
+    // this means the form was opened from host list, which passes host_id as id, so we have to rename it
```
prat98, line 187: We need to remove comments before merging

prat98 (commented):
Changes look good! Just a few small things to update that I found

frontend/awx/resources/inventories/inventoryHostsPage/InventoryHostForm.tsx
MilanPospisil: Ok, But I left this for purpose.

prat98 (approved):

PR #1816, opened 2024-03-11
implemented relaunch/cancel btn for wf job header

resoluteCoder:
Jira:
- https://issues.redhat.com/browse/AAP-21531
- https://issues.redhat.com/browse/AAP-21533


frontend/awx/views/jobs/hooks/useJobHeaderActions.tsx
```diff
@@ -0,0 +1,51 @@
...
+    };
+
+    const actions: IPageAction<UnifiedJob>[] = [
+      {
+        type: PageActionType.Button,
+        selection: PageActionSelection.Single,
+        variant: ButtonVariant.secondary,
+        isPinned: true,
+        icon: RocketIcon,
+        label: t(`Relaunch job`),
```
jerabekjiri, line 32: Should we let the user relaunch a job if it's still running/pending? It can lead to creating a lot of the same pending jobs.

Screenshot from 2024-03-12 14-49-41

frontend/awx/views/jobs/hooks/useJobHeaderActions.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { useCancelJobs } from './useCancelJobs';
+import { useRelaunchJob } from './useRelaunchJob';
+
+export function useJobHeaderActions(onComplete: (jobs: UnifiedJob[]) => void) {
+  const { t } = useTranslation();
+  const cancelJobs = useCancelJobs(onComplete);
+  const relaunchJob = useRelaunchJob();
+
+  return useMemo<IPageAction<UnifiedJob>[]>(() => {
+    const cannotCancelJob = (job: UnifiedJob) => {
```
jerabekjiri, line 17: This method is already defined in https://github.com/ansible/ansible-ui/blob/main/frontend/awx/views/jobs/hooks/useJobRowActions.tsx#L29-L35, would it be a good idea to move it to some utils dir?

jerabekjiri (approved):
Other than my two questions, LGTM. 👍

frontend/awx/views/jobs/hooks/useJobHeaderActions.tsx
resoluteCoder: @jerabekjiri It looks like the current (through row actions) and the old UI have the ability to relaunch when pending/running. 

I'd imagine the behavior would want to be transferred over to the new UI.

frontend/awx/views/jobs/hooks/useJobHeaderActions.tsx
resoluteCoder: @jerabekjiri  I went ahead and updated the code to have the function be exported in the header actions and imported into the row actions.

PR #1817, opened 2024-03-11
Add advanced search to Credentials Page

daphnemaeve:


frontend/awx/access/credentials/hooks/useCredentialsFilters.tsx
```diff
@@ -1,25 +1,12 @@
...
-      createdByToolbarFilter,
-      modifiedByToolbarFilter,
-    ],
-    [nameToolbarFilter, descriptionToolbarFilter, createdByToolbarFilter, modifiedByToolbarFilter]
-  );
-  return toolbarFilters;
+  const toolBarFilters = useDynamicToolbarFilters<Credential>({
+    optionsPath: 'credentials',
+    preSortedKeys: ['name'],
+    preFilledValueKeys: ['name', 'credential_type'],
```
kialam, line 8: Let's add `credential_type` to `knownAwxFilterKeys` like so:

```
credential_type: {
    resourceType: 'credential_type',
    params: { order_by: '-created' },
    labelKey: 'name',
    valueKey: 'id',
  }, 
```

That way the dropdown values will be a list of credential type names instead of their IDs.

frontend/awx/access/credentials/hooks/useCredentialsFilters.tsx
```diff
@@ -1,25 +1,12 @@
...
+  });
```
kialam, line 9: We'll also want to include the createdByToolbarFilter and modifiedByToolbarFilter. I don't think those queries are returned by the API in the OPTIONS response.

PR #1819, opened 2024-03-11
Instances List Updates

prat98:
This PR addresses a few updates suggested by UXD for instances list. Which includes the following:

**Instances Table View:**

- remove Last Health Check from being shown as the default column on the instance table view and have it just live in the details view of the instance
- Add Used Capacity as a column to the instances table and have the bar and percentage fit to scale
- On the top toolbar the "Remove instance" and "Run Health Check" should live in a kebab menu, both should only become enabled when at least one of the instances is checked/selected in the table

**Instance Details View:**

- Enable/disable toggle that is currently in the details section should be in the top header to the left of the primary "Edit instance" button
- "Remove instance" should be in the kebab menu
- "Last Health Check" should be added to the details view, before the Created and Modified fields

In addition to this the following has also been added: 

the status column has been updated to display the instances `node_state` along with a tooltip indicating the last time an health check was run on an instance

A new capacity adjustment column which uses a new `InstanceForksSlider` component to render an interactive slider to adjust instance forks from the list page

An improved UX for the health check flow...

1. Running a health check from the toolbar: 
- A health check button is present in the kebab menu of the instance list toolbar which is only enabled if all selected instances are a execution nodes that currently do not have a running/pending health checks on them. Upon clicking the health check button a classic AWX `bulkconfirmationdialog` is shown to the user. After confirming the instances to run health checks on, the user is notified if the request was sent successfully and the status column of all chosen instances will update to indicate to the user a health check is currently running on them. In addition, the toolbar health check button is disabled if any instance is selected with a currently pending/running health check or a hop or control node is selected. 

3. Running a health check using the row actions: 
 - A health check button is present in the row actions of all instances. The health check button is only enabled if the instance is an execution node that currently does not have a running/pending health check running against it. Upon clicking the row action health check, a temp toast is shown to the user indicating a health check has been triggered against the instance, the status column is updated to `running` to indicate to the user a health check is running against the instance. In addition, through the duration of the health check, the row action button for health check is disabled with a tooltip reading `Health check pending` to ensure only one health check is attempted on an instance at a time. 

4. Running a health check using the instance page action:
-  A health check button is present in the top header of the tabbed instance page behind a kebab button. It is only visible for execution nodes and is only enabled when there is no pending/running health check on the instance. Clicking on this button triggers a health check on the instance and a temp toast displays to the user indicating a health check has been triggered against the instance, and the status field is updated to `running` to indicate to the user a health check is running against the instance. In addition, through the duration of the health check, the health check button is disabled with a tooltip reading `Health check pending` to ensure only one health check is attempted on an instance at a time. 

**Known Issues:**
There is an issue with displaying expandable columns. When setting the table property to `ColumnTableOption.expanded` the column does not appear at all.

**Future improvements:**
Use websockets to update status of instance once health check is completed. Not sure if it's possible to do that for instances right now. I see websockets being used in other parts of the UI for example `Projects` and `Inventories`.

### DIFF
--- a/framework/components/Unavailable.tsx
+++ b/framework/components/Unavailable.tsx
@@ -1,0 +1,14 @@
+import { ReactNode } from 'react';
+import { getPatternflyColor } from './pfcolors';
+
+export function Unavailable(props: { children: ReactNode }) {
+  return (
+    <span
+      style={{
+        color: getPatternflyColor('red'),
+      }}
+    >
+      {props.children}
+    </span>
+  );
+}
--- a/frontend/awx/administration/instances/InstanceDetails.tsx
+++ b/frontend/awx/administration/instances/InstanceDetails.tsx
@@ -112,7 +112,7 @@ export function InstanceDetailsTab(props: {
           ))}
         </PageDetail>
       )}
-      {instance.related?.install_bundle && (
+      {!instance.managed && instance.related?.install_bundle && (
         <PageDetail label={t`Download bundle`} data-cy="download-bundle">
           <Button
             size="sm"
--- a/frontend/awx/administration/instances/InstanceDetails.tsx
+++ b/frontend/awx/administration/instances/InstanceDetails.tsx
@@ -97,15 +97,7 @@ export function InstanceDetailsTab(props: {
         </Tooltip>
       </PageDetail>
       <PageDetail label={t('Status')} data-cy="node-status">
-        <StatusCell
-          status={
-            !instance.enabled
-              ? 'disabled'
-              : instance.health_check_pending
-                ? 'running'
-                : instance.node_state
-          }
-        />
+        <StatusCell status={instance.health_check_pending ? 'running' : instance.node_state} />
       </PageDetail>
       {instanceGroups && instanceGroups.results.length > 0 && (
         <PageDetail label={t(`Instance groups`)} data-cy="instance-groups">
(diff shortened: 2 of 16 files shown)


frontend/awx/administration/instances/InstanceDetails.tsx
```diff
@@ -67,134 +66,119 @@ export function InstanceDetailsTab(props: {
...
+            <Label color="blue" style={{ marginRight: '10px' }} key={instance.id}>
+              <Link to={getPageUrl(AwxRoute.InstanceGroupDetails, { params: { id: instance.id } })}>
+                {/* eslint-disable-next-line i18next/no-literal-string */}
+                {instance.name}
+              </Link>
+            </Label>
+          ))}
+        </PageDetail>
+      )}
+      {instance.related?.install_bundle && (
```
acruzgon, line 115: Could you please add the tighten validation for Download bundle? like [this](https://github.com/ansible/awx/commit/40150a2be840aa75537165237ee398a38e01e96a#diff-7e68ea2512bd483497e12ce9e359fec5c3c1819aceabc3155be368330dba4217R212)

acruzgon (changes requested):

frontend/awx/administration/instances/InstanceDetails.tsx
prat98: done

acruzgon (approved):
lgtm - great job!

framework/components/Unavailable.tsx
```diff
@@ -0,0 +1,14 @@
+import { ReactNode } from 'react';
+import { getPatternflyColor } from './pfcolors';
+
+export function Unavailable(props: { children: ReactNode }) {
```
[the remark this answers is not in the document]
prat98: Thanks :)

PR #1820, opened 2024-03-12
Host details

MilanPospisil:
image


MilanPospisil:
> Everything looks good just a few small changes. Need to add a description field and an inventory field help text like the old UI has in the image below
> 
> Screenshot 2024-03-12 at 7 38 35 AM

Ok, I did not saw this in the issue, so I did not know if we want it here :-)

prat98 (approved):

PR #1824, opened 2024-03-12
[AAP-21458] cancel job action in job output toolbar

jerabekjiri:
Issue: AAP-21458

Screenshot from 2024-03-12 18-15-27

I'm wondering if we still need this since https://github.com/ansible/ansible-ui/pull/1816 ([AAP-21533](https://issues.redhat.com/browse/AAP-21533)) adds `Cancel job` button to header? 
CC @marshmalien @resoluteCoder


resoluteCoder:
That's a good question. I would think it should be removed. The header buttons will be the primary focus. 

Just my thoughts.

marshmalien (commented):
Good catch! I agree with you both and think it should be removed.

jerabekjiri:
Replaced by https://github.com/ansible/ansible-ui/pull/1816

PR #1827, opened 2024-03-13
Create, Edit and Delete host

MilanPospisil:
Enables Create and Delete host under inventories to work also under standalone hosts. Also add edit test.


RedOrNot:
Amazing work improving the code!!!!
approved!

PR #1828, opened 2024-03-13
Adds Schedules Occurrences Step

AlexSCorey:
This addresses AAP-21599 

It it adds the step, but does not have any logic to save an occurrence, or persist a schedule.


matoval (approved):
LGTM, ran locally and works as expected

PR #1830, opened 2024-03-13
Add advanced search to credentials page

daphnemaeve:
Addresses https://issues.redhat.com/browse/AAP-21056

### DIFF
--- a/frontend/awx/access/credentials/hooks/useCredentialsFilters.tsx
+++ b/frontend/awx/access/credentials/hooks/useCredentialsFilters.tsx
@@ -10,7 +10,7 @@ export function useCredentialsFilters() {
   const toolBarFilters = useDynamicToolbarFilters({
     optionsPath: 'credentials',
     preSortedKeys: ['name', 'credential_type', 'created-by', 'modified-by'],
-    preFilledValueKeys: ['name', 'credential_type', 'id'],
+    preFilledValueKeys: ['name', 'id'],
     additionalFilters: [createdByToolbarFilter, modifiedByToolbarFilter],
   });
 
--- a/frontend/awx/access/credentials/hooks/useCredentialsFilters.tsx
+++ b/frontend/awx/access/credentials/hooks/useCredentialsFilters.tsx
@@ -1,25 +1,12 @@
-import { useMemo } from 'react';
-import { IToolbarFilter } from '../../../../../framework';
-import {
-  useCreatedByToolbarFilter,
-  useDescriptionToolbarFilter,
-  useModifiedByToolbarFilter,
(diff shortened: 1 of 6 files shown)


frontend/awx/access/credentials/hooks/useCredentialsFilters.tsx
```diff
@@ -1,25 +1,18 @@
...
-      descriptionToolbarFilter,
-      createdByToolbarFilter,
-      modifiedByToolbarFilter,
-    ],
-    [nameToolbarFilter, descriptionToolbarFilter, createdByToolbarFilter, modifiedByToolbarFilter]
-  );
-  return toolbarFilters;
+  const toolBarFilters = useDynamicToolbarFilters({
+    optionsPath: 'credentials',
+    preSortedKeys: ['name'],
```
skochay330, line 12: I would put credential type towards the top of the filter list since it is prefilled. I would do the same with created by and modified by since those were additionally added. So the order would be name, credential type, created by, modified by- just my two cents

daphnemaeve: Done! Also put ID in the preFilledValueKeys.

frontend/awx/access/credentials/hooks/useCredentialsFilters.tsx
```diff
@@ -1,25 +1,18 @@
...
-      createdByToolbarFilter,
-      modifiedByToolbarFilter,
-    ],
-    [nameToolbarFilter, descriptionToolbarFilter, createdByToolbarFilter, modifiedByToolbarFilter]
-  );
-  return toolbarFilters;
+  const toolBarFilters = useDynamicToolbarFilters({
+    optionsPath: 'credentials',
+    preSortedKeys: ['name', 'credential_type', 'created-by', 'modified-by'],
+    preFilledValueKeys: ['name', 'credential_type', 'id'],
```
kialam, line 13: Adding `credential_type` to `knownAwxFilterKeys` means we pre-fetch the values - we don't need to add it here.

skochay330 (approved):
Tested in UI, LGTM!

frontend/awx/access/credentials/hooks/useCredentialsFilters.tsx
daphnemaeve: I put it there so that credential type would be one of the first filters in the dropdown as Salma suggested. Is there another way you'd suggest doing that?

kialam: It doesn't need to be in `preFilledValueKeys`.

PR #1833, opened 2024-03-13
Ensure workflow save functionality does not cancel client-side API requests

marshmalien:
Bug: The workflow visualizer's save functionality cancels API requests on the client-side, particularly when multiple POST requests are made to the same `usePostRequest` instance.

By default, `usePostRequest` initializes a ref to handle the `AbortController` instance. It then checks if there is another instance of `AbortController`, and if it exists, it calls its abort() method. If we don't pass in our own custom signal, then the `AbortController` instance's signal will be used in the fetch request. 

A bug arises when `usePostRequest` is used more than once, leading to request cancellations. These multiple instances reference the same instance of the `AbortController`, resulting in unintended aborts. 

```
if (abortControllerRef.current.abortController) {
  abortControllerRef.current.abortController.abort();
}
```

Bug screenshot
Screenshot 2024-03-13 at 4 03 49 PM

Bug fix working:
add_wf_nodes


nixocio (approved):
Great job @marshmalien

PR #1838, opened 2024-03-14
Update facts component to also render on host page

prat98:
Small update to enable to the facts component to render correctly on both the inventories hosts facts and hosts facts pages.


MilanPospisil (approved):
LGTM :+1:

PR #1843, opened 2024-03-14
Async select search query support

jamestalton:
Big change. :( 

Enhances the asynchronous select components.
- Fixes displaying the current selections even if not loaded in the current paginated options.
- Updates the load more interfaces to be able to load incrementally not using page.
  - i.e. get the next 10 items after the last item received vs page 2
  - this fixes issues of duplicates and missed items between pages
- Added support for search with a debounce that then updates the paginated query to use that search for results.

Reviewed with UXD.

Screenshot 2024-03-14 at 4 43 35 PM

Screenshot 2024-03-14 at 4 44 15 PM


nixocio (approved):
I got in a call with @jamestalton there are some future improvements that would be added in follow up PRs. I locally tested those changes, and they seem to be working as expected.

PR #1844, opened 2024-03-15
Fix Create Host Inventories Load More

MilanPospisil:
Load more in inventories dropdown now works correctly.

image

### DIFF
--- a/frontend/awx/resources/inventories/inventoryHostsPage/InventoryHostForm.tsx
+++ b/frontend/awx/resources/inventories/inventoryHostsPage/InventoryHostForm.tsx
@@ -305,10 +305,12 @@ export function EditHost() {
 function HostInputs(props: { edit_mode?: boolean; inventory_host?: boolean }) {
   const { t } = useTranslation();
 
-  const queryOptions = useCallback(async () => {
+  const page_size = 10;
+  const queryOptions = useCallback(async (page: number) => {
     const response = await requestGet<AwxItemsResponse<Inventory>>(
-      awxAPI`/inventories/?order_by=name`
+      awxAPI`/inventories/?order_by=name&page=${page.toString()}&page_size=${page_size.toString()}`
     );
+
     return Promise.resolve({
       total: response.count,
       options:


prat98 (commented):
These changes work well! But using `PageFormSingleSelectAwxResource` component instead of `pageFormAsyncSingleSelect` might be a solution more consistent with the rest of the UI. This component wraps `pageFormAsyncSingleSelect` and has the queryOptions built in so they don't have to be set manually https://github.com/ansible/ansible-ui/blob/main/frontend/awx/common/PageFormSingleSelectAwxResource.tsx#L36. `PageFormOrganizationSelect` is currently using this component as well. Something like the code below should work
```typescript
<PageFormSingleSelectAwxResource<Inventory>
  name="inventory.id"
  id="inventory"
  label={t('Inventory')}
  placeholder={t('Select Inventory')}
  queryPlaceholder={t('Loading inventory...')}
  queryErrorText={t('Error loading inventory')}
  isRequired={true}
  url={awxAPI`/inventories/`}
  tableColumns={inventoryColumns}
  toolbarFilters={inventoryFilters}
/>
``` 
along with setting the columns and filters above like so

```typescript
const inventoryColumns = useInventoriesColumns({ disableLinks: true });
const inventoryFilters = useInventoriesFilters();
```

MilanPospisil:
> These changes work well! But using `PageFormSingleSelectAwxResource` component instead of `pageFormAsyncSingleSelect` might be a solution more consistent with the rest of the UI. This component wraps `pageFormAsyncSingleSelect` and has the queryOptions built in so they don't have to be set manually https://github.com/ansible/ansible-ui/blob/main/frontend/awx/common/PageFormSingleSelectAwxResource.tsx#L36. `PageFormOrganizationSelect` is currently using this component as well. Something like the code below should work
> 
> ```ts
> <PageFormSingleSelectAwxResource<Inventory>
>   name="inventory.id"
>   id="inventory"
>   label={t('Inventory')}
>   placeholder={t('Select Inventory')}
>   queryPlaceholder={t('Loading inventory...')}
>   queryErrorText={t('Error loading inventory')}
>   isRequired={true}
>   url={awxAPI`/inventories/`}
>   tableColumns={inventoryColumns}
>   toolbarFilters={inventoryFilters}
> />
> ```
> 
> along with setting the columns and filters above like so
> 
> ```ts
> const inventoryColumns = useInventoriesColumns({ disableLinks: true });
> const inventoryFilters = useInventoriesFilters();
> ```

Ok, thanks, I will try. I had something very similar in our Hub too, however I was then focused to finish the hub project rather then do things in framework. I will use that :-)

PR #1857, opened 2024-03-19
Constructed inventory hosts

MilanPospisil:
image

Constructed inventories are implemented as extension of Inventory hosts component, they are similar to smart inventories. And component tests is again implemented in hosts.cy.tsx.


frontend/awx/resources/hosts/hooks/useHostsFilters.tsx
```diff
@@ -17,11 +17,24 @@ export function useHostsFilters() {
...
 
-    if (params.inventory_type === 'inventory' || !params.inventory_type) {
-      filters.push(createdByToolbarFilter);
+    if (!params.inventory_type || params.inventory_type === 'inventory') {
+      filters = [
+        nameToolbarFilter,
+        descriptionToolbarFilter,
+        createdByToolbarFilter,
+        modifiedByToolbarFilter,
+      ];
```
AlexSCorey, line 28:
```suggestion
    if (!params.inventory_type || params.inventory_type === 'inventory') {
      filters = filters.splice(1, 0, descriptionToolbarFilter)
```

I think that'll work to make this whole conditional a bit shorter.

frontend/awx/resources/hosts/hooks/useHostsFilters.tsx
```diff
@@ -17,11 +17,24 @@ export function useHostsFilters() {
...
     }
+
+    if (
+      params.inventory_type === 'smart_inventory' ||
+      params.inventory_type === 'constructed_inventory'
+    ) {
+      filters = [nameToolbarFilter, createdByToolbarFilter, modifiedByToolbarFilter];
+    }
+
```
AlexSCorey, line 37:
```suggestion
```

frontend/awx/resources/inventories/InventoryPage/InventoryHosts.tsx
```diff
@@ -35,6 +35,23 @@ export function InventoryHosts() {
...
+  if (params.inventory_type === 'inventory') {
+    emptyStateTitle = canCreateHost
+      ? t('There are currently no hosts added to this inventory.')
+      : t('You do not have permission to create a host.');
+
+    emptyStateDescription = canCreateHost
+      ? t('Please create a host by using the button below.')
+      : t('Please contact your organization administrator if there is an issue with your access.');
+  } else {
+    emptyStateTitle = t('No Hosts Found');
```
AlexSCorey, line 51:
```suggestion
    emptyStateTitle = t('No hosts found');
```

frontend/awx/resources/inventories/InventoryPage/InventoryHosts.tsx
```diff
@@ -35,6 +35,23 @@ export function InventoryHosts() {
...
+    emptyStateDescription = t('Please add Hosts to populate this list');
+  }
+
```
AlexSCorey, line 54: We need to decide whether we are capitalizing Hosts/Host here.  Currently we are doing a bit of both.

frontend/awx/resources/hosts/Hosts.cy.tsx
```diff
@@ -9,15 +9,32 @@ describe('Hosts.cy.ts', () => {
...
-  const types = [hosts, inventory_hosts, smart_inventory_hosts];
+  const types = [hosts, inventory_hosts, smart_inventory_hosts, constructed_inventory_hosts];
 
   types.forEach((type) => {
     const typeDesc = ` (${type})`;
     const component = type === hosts ? <Hosts /> : <InventoryHosts />;
     const path = type === hosts ? '/hosts' : '/inventories/:inventory_type/:id/hosts';
 
-    const inventory_type = type === smart_inventory_hosts ? 'smart_inventory' : 'inventory';
+    let inventory_type = '';
```
AlexSCorey, line 21: I understand we are trying to reuse code that would otherwise be duplicated for each type of inventory.  However, I tend to think that in this case duplicating the code....or making it less complex is a good thing.  It helps us quickly identify any test failures that may get caught in the future.

AlexSCorey (changes requested):
This looks great.  Most of my comments are super nit picky.  Let me know if you have questions.

frontend/awx/resources/inventories/InventoryPage/InventoryHosts.tsx
MilanPospisil: I found more of those differences, some of them are corrected in my different PRs. I wanted to file an issue, lets communicate this in channel.

frontend/awx/resources/hosts/Hosts.cy.tsx
MilanPospisil: I do see advantage of your approach, but having so much duplicated code will be bad too. What about this compromise - create functions for duplicated its and call the issue of those its and get rid of the foreach? So it will be more linear but with reused functions.

PR #1859, opened 2024-03-19
Constructed inventory groups

MilanPospisil:


prat98:
Changes look good! Just one small issue I found with the empty list state. User's should not be able to navigate to the create group form for constructed inventories. We should update the empty state with the appropriate description and remove the button to add a group if the inventory is a constructed inventory.

 
Screenshot 2024-03-28 at 11 36 30 AM

prat98 (approved):
LGTM! 🚀

PR #1860, opened 2024-03-19
Workflow Visualizer Remaining Tests

akus062381:
1. Fixes the skipped tests and reorganized the spec file to get it working in an optimized way.
2. Adds a test that removes the link between 2 nodes.
3. Adds a test for choosing prompt on launch settings on a node, launching the WFJT and asserting information on the output screen.

### DIFF
--- a/cypress/e2e/awx/resources/workflowVisualizerCRUD.cy.ts
+++ b/cypress/e2e/awx/resources/workflowVisualizerCRUD.cy.ts
@@ -148,7 +148,7 @@ describe('Workflow Visualizer', () => {
       cy.contains('Workflow Visualizer').should('be.visible');
       cy.get(`g[data-id="${projectNode.id}"]`)
         .should('be.visible')
-        .and('contain', `${projectNode.summary_fields.unified_job_template.name}`);
+        .and('contain', `${projectNode?.summary_fields?.unified_job_template?.name}`);
       cy.get(`g[data-id="${approvalNode.id}"] [class*="node-label"]`).should('be.visible').click();
       cy.getByDataCy('workflow-topology-sidebar').should('be.visible');
       cy.contains('h2', 'Node details').should('be.visible');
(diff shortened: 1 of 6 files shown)


cypress/e2e/awx/resources/workflowVisualizerCRUD.cy.ts
```diff
@@ -0,0 +1,459 @@
...
+        `${workflowJobTemplate.name}`,
+        'wf_vis_testing_A.json'
+      );
+      cy.get('[class*="66-node-label"]')
+        .should('exist')
+        .should('contain', 'Cleanup Activity Stream');
+      cy.get('[class*="43-node-label"]').should('exist').should('contain', 'bar');
+      cy.get('[class*="42-node-label"]').should('exist').should('contain', '1');
+      cy.get('[class*="41-node-label"]').should('exist').should('contain', 'Demo Project');
+      cy.deleteAwxWorkflowJobTemplate(workflowJobTemplate);
```
nixocio, line 71: There are resources being delete as part of the `afterEach`, and `after`. This one seem to be deleted as part of them. Either we create the workflow as part of the test itself, and delete at the end. Or we create as part of a `before`, and delete as part of `after`.  This is a general comment. We might address in a different PR.

cypress/e2e/awx/resources/workflowVisualizerCRUD.cy.ts
```diff
@@ -0,0 +1,459 @@
...
+      cy.get('[data-cy="workflow-visualizer"]').should('be.visible');
+      cy.get('h4.pf-v5-c-empty-state__title-text').should(
+        'have.text',
+        'There are currently no nodes in this workflow'
+      );
+      cy.get('div.pf-v5-c-empty-state__actions').within(() => {
+        cy.get('[data-cy="add-node-button"]').should('be.visible');
+      });
+      cy.get('button[data-cy="workflow-visualizer-toolbar-close"]').click();
+      cy.verifyPageTitle(`${jtName}`);
```
nixocio, line 92: Should we assert that certain details are shown as part of the details page? At least the ones added to create the workflow job template?

nixocio (approved):

PR #1863, opened 2024-03-19
Adds exceptions step to schedules wizard

AlexSCorey:
This addresses AAP-21600.  It adds the exception to to the wizard.  Adds a few tests, does a bit of refactoring.  It does not have any save functionality.


matoval (approved):
LGTM! Ran locally and worked as expected

PR #1868, opened 2024-03-20
Add terraform state inventory source

hakbailey:
A new Terraform state inventory source was added to AWX in [PR 14840](https://github.com/ansible/awx/pull/14840). This PR updates the UI to include the new source type. It also adds Execution Environment selection to the inventory source add/edit forms, as an EE selection is required for the Terraform source -- users must provide and select an EE containing a Terraform binary, as we cannot ship one with the Terraform binary for licensing reasons.

In addition, this PR fixes a related bug: when trying to edit an inventory source that is not sourced from a project, the API returns a 400: Bad request error with detail "Cannot set source_path if not SCM type." due to a default value in the form for the source_path field.

### DIFF
--- a/cypress/e2e/awx/resources/inventorySource.cy.ts
+++ b/cypress/e2e/awx/resources/inventorySource.cy.ts
@@ -107,6 +107,12 @@ describe('Inventory source page', () => {
     cy.verifyPageTitle('amazon ec2 source');
     cy.clickButton('Edit inventory source');
     cy.verifyPageTitle('Edit source');
+    cy.getBy('[data-cy="name"]').clear().type('updated amazon ec2 source');
+    cy.getBy('[data-cy="overwrite_vars"]').check();
+    cy.getBy('[data-cy="Submit"]').click();
+    cy.verifyPageTitle('updated amazon ec2 source');
+    cy.clickButton('Edit inventory source');
+    cy.verifyPageTitle('Edit source');
     cy.getBy('[data-cy="name"]').clear().type('new project');
     cy.selectDropdownOptionByResourceName('source_control_type', 'Sourced from a Project');
     cy.getBy('[data-cy="overwrite_vars"]').check();
--- a/cypress/e2e/awx/resources/inventorySource.cy.ts
+++ b/cypress/e2e/awx/resources/inventorySource.cy.ts
@@ -78,17 +78,17 @@ describe('Inventory source page', () => {
     cy.selectDropdownOptionByResourceName('project', project.name);
     cy.selectDropdownOptionByResourceName('inventory', 'Dockerfile');
     cy.getBy('[data-cy="execution-environment-select-form-group"]').within(() => {
-      cy.getBy('.pf-v5-c-input-group > .pf-v5-c-button').click();
+      cy.getBy('[data-ouia-component-id="lookup-execution_environment-button"]').click();
     });
-    cy.getBy('[data-ouia-component-type="PF5/ModalContent"]').within(() => {
+    cy.getBy('[data-ouia-component-id="Select an execution environment-dialog"]').within(() => {
       cy.searchAndDisplayResource(executionEnvironmentName);
       cy.getBy('[data-cy="checkbox-column-cell"] > label').click();
       cy.contains('button', 'Confirm').click();
     });
     cy.getBy('[data-cy="credential-select-form-group"]').within(() => {
-      cy.getBy('.pf-v5-c-input-group > .pf-v5-c-button').click();
+      cy.getBy('[data-ouia-component-id="lookup-credential-button"]').click();
     });
-    cy.getBy('[data-ouia-component-type="PF5/ModalContent"]').within(() => {
+    cy.getBy('[data-ouia-component-id="Select credential"]').within(() => {
       cy.searchAndDisplayResource(credentialName);
       cy.getBy('[data-cy="checkbox-column-cell"] > label').click();
       cy.getBy('#submit').click();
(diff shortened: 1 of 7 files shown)


cypress/e2e/awx/resources/inventorySource.cy.ts
```diff
@@ -70,7 +77,17 @@ describe('Inventory source page', () => {
...
+    });
+    cy.getBy('[data-ouia-component-type="PF5/ModalContent"]').within(() => {
+      cy.searchAndDisplayResource(executionEnvironmentName);
+      cy.getBy('[data-cy="checkbox-column-cell"] > label').click();
+      cy.contains('button', 'Confirm').click();
+    });
+    cy.getBy('[data-cy="credential-select-form-group"]').within(() => {
+      cy.getBy('.pf-v5-c-input-group > .pf-v5-c-button').click();
+    });
     cy.getBy('[data-ouia-component-type="PF5/ModalContent"]').within(() => {
```
AlexSCorey, line 91: I think this is not part of your work, but we should try to find a different selector here too.

cypress/e2e/awx/resources/inventorySource.cy.ts
```diff
@@ -70,7 +77,17 @@ describe('Inventory source page', () => {
     cy.selectDropdownOptionByResourceName('source_control_type', 'Sourced from a Project');
     cy.selectDropdownOptionByResourceName('project', project.name);
     cy.selectDropdownOptionByResourceName('inventory', 'Dockerfile');
-    cy.getBy('.pf-v5-c-input-group > .pf-v5-c-button').click();
+    cy.getBy('[data-cy="execution-environment-select-form-group"]').within(() => {
+      cy.getBy('.pf-v5-c-input-group > .pf-v5-c-button').click();
+    });
+    cy.getBy('[data-ouia-component-type="PF5/ModalContent"]').within(() => {
```
AlexSCorey, line 83: Lets try to find a different locator.  We avoid using these locators, because they come from patternfly, and they are subject to change, which would break out tests.

cypress/e2e/awx/resources/inventorySource.cy.ts
```diff
@@ -70,7 +77,17 @@ describe('Inventory source page', () => {
...
+      cy.searchAndDisplayResource(executionEnvironmentName);
+      cy.getBy('[data-cy="checkbox-column-cell"] > label').click();
```
AlexSCorey, line 85: Our app has a convention in place to use `cy.get`.  I realize this file already has uses of `cy.getBy` though I'm not real sure that is any better.  If we want to be sepecific and get a selector we can use `cy.getBySel` instead.  Otherwise I'd advoated for us to follow existing patterns and just use `cy.get`

Here are the [docs](https://docs.cypress.io/guides/references/best-practices#Real-World-Example)

frontend/awx/interfaces/InventorySource.ts
```diff
@@ -107,7 +107,8 @@ export interface InventorySourceForm {
   inventory?: number;
   source_path: { name: string | undefined | null };
   source_script?: string;
-  execution_environment?: SummaryFieldsExecutionEnvironment;
+  execution_environment?: string;
+  execution_environmentIdPath?: number;
```
AlexSCorey, line 111: What's the different use cases for these 2 lines? Do we need them both?

AlexSCorey (changes requested):
This is a really awesome catch in bringing in this functionality.  Great start on the code.  I have left a few comments, and some questions.  Let me know if I can help in anyway!

Screenshot 2024-03-20 at 10 45 51 AM

Are only showing certain types of credentials?

cypress/e2e/awx/resources/inventorySource.cy.ts
hakbailey: That makes sense, I can poke around and see if there's a better option.

cypress/e2e/awx/resources/inventorySource.cy.ts
hakbailey: Got it, I was trying to follow precedent in this file but can go through and update all of these.

frontend/awx/interfaces/InventorySource.ts
hakbailey: Yes it seems like it. We need the EE name string to be displayed in the search box but we need the EE ID number to be set when selecting an EE and submitting the data to the API. I'm not sure if this is the best way to do it, but it worked for the credential field so I just did the same here :)

cypress/e2e/awx/resources/inventorySource.cy.ts
hakbailey: Oh hmmm, looking at the [custom commands](https://github.com/ansible/ansible-ui/blob/d9113fdc0bb0e89f999f141588ece6d85115684f/cypress/COMMANDS.md?plain=1#L5) it seems like we might actually want the "makes sure the element is enabled and visible" features of `cy.getBy` for at least some of the fields in this file? A few fields have `hidden` logic based on the source type and the submit buttons are all disabled until the required data is entered.

cypress/e2e/awx/resources/inventorySource.cy.ts
hakbailey: It looks like we are setting `data-ouia-component-id` explicitly so I updated these to use that instead.

AlexSCorey:
@hakbailey We have some failing tests.  I haven't gone through to figure out the reason they are failing, but a rebase may help resolve _some_  of them.  Once we have passing tests I can better review this pr.  Thanks for sticking with us on this one!

hakbailey:
> @hakbailey We have some failing tests. I haven't gone through to figure out the reason they are failing, but a rebase may help resolve _some_ of them. Once we have passing tests I can better review this pr. Thanks for sticking with us on this one!

Ah, I made this PR from a fork instead of a branch in the repo, should I close this and do the latter instead? My repo doesn't have the secrets needed for CI tests to run.

PR #1881, opened 2024-03-21
Contructed inventory add edit form

MilanPospisil:
Create and Edit constructed inventory. Enhance the InventoryForm, which is used for all 3 types - inventory, smart and constructed.

Also:
- created Multi selector for AWX resources
- added QueryParams and LabelHelp to both single and multi AWX resources
- disabled Inventory form create and edit submit on enter, so labels can be added by typing and pressing enter
(without that, the user will type and possibly press enter and he dont want the form to be submitted).
- added error handling for edit form data loading and loading spinner     
- component and e2e tests will come in follow ups PRs

image

### DIFF
--- a/frontend/awx/common/PageFormMultiSelectAwxResource.tsx
+++ b/frontend/awx/common/PageFormMultiSelectAwxResource.tsx
@@ -34,23 +34,10 @@ export function PageFormMultiSelectAwxResource<
 }) {
   const id = useID(props);
 
-  function constructQueryParams() {
-    let queryString = '&';
-    if (props.queryParams) {
-      Object.keys(props.queryParams).forEach((key) => {
-        queryString += key + '=' + props.queryParams?.[key];
-      });
-    } else {
-      return '';
-    }
-
-    return queryString;
-  }
-
   const queryOptions = useCallback<PageAsyncSelectOptionsFn<PathValue<FormData, Name>>>(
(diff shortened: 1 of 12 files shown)


frontend/awx/common/PageFormMultiSelectAwxResource.tsx
```diff
@@ -0,0 +1,172 @@
...
+      title={props.title}
+      onSelect={props.onSelect}
+      toolbarFilters={props.toolbarFilters ?? []}
+      tableColumns={props.tableColumns}
+      view={view}
+    />
+  );
+}
+
+export function constructQueryParams(queryParams: QueryParams) {
```
prat98, line 161: Do we want to put this function in a separate file maybe under utils?

MilanPospisil: I will look if there is any usage of this functions in fw.

prat98:
Found one more thing. Need to change `PageFormCodeEditor` label from `Variables` to `Source vars` and add a help text for that section of the form. In addition, this input should be a required field and needs to have validation on it to ensure a `plugin` parameter is provided in the source vars.
Screenshot 2024-03-28 at 11 31 20 AM

prat98 (approved):
LGTM! 🚀

PR #1883, opened 2024-03-21
PageFormExecutionEnvironmentSelect Update

jamestalton:
Updated `PageFormExecutionEnvironmentSelect` to use the `PageFormSingleSelectAwxResource` component.

Our select dropdowns current have a bug where if an update happens, the browse dialog will go away. PageFormSingleSelectAwxResource fixes that bug and also updates the dropdown to use the newer PageSingleSelect.


framework/PageInputs/PageSingleSelect.tsx
```diff
@@ -12,6 +12,7 @@ import {
   SelectList,
```
MilanPospisil, line 12: Forgotten changes for PageMultiSelect?

MilanPospisil:
Are we going to do those extensions also in multi versions?

framework/PageInputs/PageSingleSelect.tsx
```diff
@@ -162,42 +162,49 @@ export function PageSingleSelect<
 
   const selectedLabel = useMemo(() => {
     let selectedLabel: ReactNode = selectedOption?.label;
-    if (!selectedLabel && value !== undefined && value !== '') {
+    if (!selectedLabel && value !== undefined && value !== '' && value !== null) {
```
MilanPospisil, line 165: what about simplification:
!selectedLabel && value

jamestalton: If value was 0 that would fail.

jamestalton:
> Are we going to do those extensions also in multi versions?

Yes

framework/PageInputs/PageSingleSelect.tsx
MilanPospisil: Oh, so it is not only string, makes sense.

akus062381 (changes requested):
Can we give a user the ability to remove a selected form field option for a form field that is not marked as required?

akus062381:
When I go to the Execution Environment form field and select Browse, I see the following:

image

akus062381:
When I click into the Edit form for the Job Template, it shows the execution environment's ID instead of its name. Can we change this so the name is displayed?
image

PR #1888, opened 2024-03-21
[AAP-19886] Add organizations

lgalis:
Screenshot from 2024-03-29 13-29-18
Screenshot from 2024-03-26 17-25-39
Screenshot from 2024-03-26 17-25-25
Screenshot from 2024-03-26 17-25-20

### DIFF
--- a/cypress/fixtures/edaOrganizations.json
+++ b/cypress/fixtures/edaOrganizations.json
@@ -1,0 +1,79 @@
+{
+  "count": 12,
+  "next": "/api/eda/v1/organizations/?page=2&page_size=10",
+  "previous": null,
+  "page_size": 10,
+  "page": 1,
+  "results": [
+    {
+      "id": 1,
+      "name": "Default",
+      "description": "The default organization",
+      "created": "2024-03-28T15:00:08.079759Z",
+      "modified": "2024-03-28T15:00:08.079771Z"
+    },
+    {
+      "id": 2,
+      "name": "Organization 2",
+      "description": "This is Organization 2",
+      "created": "2024-03-28T15:00:08.079759Z",
+      "modified": "2024-03-28T15:00:08.079771Z"
+    },
+    {
+      "id": 3,
+      "name": "Organization 3",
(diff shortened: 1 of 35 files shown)


lgalis:
Thanks @vidyanambiar
# 1 At the moment, the org is not a mandatory field in the APIs. I will  mark it as such - but expect the same marking in the APIs. @Dostonbek1 
# 2 - If we mark the org as mandatory - then I agree - the Default will be selected by default
# 3 This is an API issue  - I will open a Jira card for that or @Dostonbek1  can fix it in the feature/rbac branch.
Regarding 
#5 - Resources that returned organization_id ( including Decision Env. ) were midfied in this PR () to return an organization ref, including id and name. Unless you use the PR above in your backend, you will not see the org in the details.

vidyanambiar:
> Resources that returned organization_id ( including Decision Env. ) were modified in this PR (https://github.com/ansible/eda-server/pull/783 ) to return an organization ref, including id and name. Unless you use the PR above in your testing, you will not see the org in the details.
Re: Decision environment I do see the organization in the details UI, its just not shown as selected in the org field in the edit form.

lgalis:
@AlanCoding and @vidyanambiar - I agree  - the Default should be selected by deafault - I will make that change. But will not mark the field as mandatory until after the backend [PR](https://github.com/ansible/eda-server/pull/783) is merged.

vidyanambiar:
@lgalis Thanks. To confirm - once `Default` is pre-filled, the only options available to the user on the `Organization` field would be to change to a different organization or keep `Default`. We should not provide an option to unset this field in Create/Edit modes (since the resource always ends up getting associated with `Default` if an org isn't provided). The effect of this would be similar to making the field required.

lgalis:
@vidyanambiar - Default org added as a default selection. DE org on the edit page fixed.

vidyanambiar (approved):
Minor: Could we show the indicate the selected organization in the modal that opens up on clicking the `Browse` button in the orgs dropdown?

PR #1891, opened 2024-03-25
[Schedule form] Schedule Form Survey Step

acruzgon:
This PR is adding Survey step for Create Schedule Wizard.
Screen:
image


matoval (approved):
LGTM! Ran locally and worked as expected

PR #1897, opened 2024-03-26
Bug fix:  Show deleted icon when resource has been deleted

marshmalien:
Screenshot 2024-03-26 at 3 18 16 PM

### DIFF
--- a/frontend/awx/interfaces/WorkflowNode.ts
+++ b/frontend/awx/interfaces/WorkflowNode.ts
@@ -31,7 +31,7 @@ export interface WorkflowNode {
       status: string;
       type: string;
     };
-    unified_job_template: {
+    unified_job_template?: {
       id: number;
       name: string;
       description: string;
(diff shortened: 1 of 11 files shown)


jerabekjiri:
Nice fix :+1: 
Screenshot from 2024-03-27 11-45-44

jerabekjiri:
So I can add a new node to the `Deleted` node and then saving throws this error. Maybe we could disable or hide 'Add step and link' for deleted nodes?

Screenshot from 2024-03-27 11-50-39

marshmalien:
> So I can add a new node to the Deleted node and then saving throws this error. Maybe we could disable or hide Add step and link option for deleted nodes?

Good find. Let me look into that error message. The old UI allows us to add nodes to deleted nodes, so I'm going to try get that working.

jerabekjiri (approved):
LGTM :+1:

PR #1899, opened 2024-03-27
AAP 21773 - prompts step

matoval:
Jira: https://issues.redhat.com/browse/AAP-21773

Use NodePromptsStep for schedule wizard


AlexSCorey (changes requested):
This is a great way to reuse code.  Awesome!  I would like to see some tests for each resource type to ensure the prompt step shows up only when required.  IE. Prompt step should show up only for job templates, or workflow job templates that have prompted on launch needs.  Projects, System Jobs, and Inventory Sources should never see the prompt step.

AlexSCorey (approved):
Looks good to me.  Well done!

PR #1900, opened 2024-03-27
Constructed inventory detail

MilanPospisil:
Added Last Job status and fixed some error handling and page loading.

Also added label helps to detail.

Added sync inventory operation for constructed inventory detail only (and only exclusively for details tab). This operation also needs to load InventorySource - it is loaded both in Page to ensure correct permissions for action are loaded and also in detail itself, when it is listening for websocket info and refreshes InventorySource data.

Add ability to cancle the running sync.


prat98 (commented):
Changes look good! We need to add help text to a few of the fields which I've highlighted in the image below. In addition, there is a sync button which updates the constructed inventory with changes in the composite inventories, this needs to be added to the page and row actions of constructed inventories. This can be done in a separate PR but just noting it down here as well.

Need to add help text here on New UI
Screenshot 2024-03-28 at 11 49 10 AM

The sync button should exist in the constructed inventory page actions here
Screenshot 2024-03-26 at 1 39 26 PM

prat98 (approved):
This looks great!

PR #1906, opened 2024-03-28
[AAP-22186] Fix Timeout field size on resize 

jerabekjiri:
Issue: [AAP-22186](https://issues.redhat.com/browse/AAP-22186)

chrome-capture-2024-3-28

### DIFF
--- a/frontend/awx/resources/templates/WorkflowVisualizer/wizard/NodeTypeStep.tsx
+++ b/frontend/awx/resources/templates/WorkflowVisualizer/wizard/NodeTypeStep.tsx
@@ -4,16 +4,13 @@ import {
   InputGroupItem,
   InputGroupText,
   TextInput,
+  Grid,
+  GridItem,
 } from '@patternfly/react-core';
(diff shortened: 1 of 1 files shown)


akus062381 (changes requested):
So actually when I saw the resizing issue was when I had the sidebar open to add the Approval Node, and with that sidebar open, I also opened the inspector on the right side of the screen. I can attach a video.

akus062381:
Video of the sizing issue:
https://www.awesomescreenshot.com/video/26249278?key=8627d15bbcf0b9ae52a6eea23e3332a3

jerabekjiri:
@akus062381 hmm so I'm not sure if I'm able to make this pixel perfect with current PF framework media breakpoints. So Is this really a problem? Since the user can expand the resizable panel in case of some weird resolution. I thought that's why we use the resizable drawer panel component in a first place? :)

akus062381:
@jerabekjiri if that is the case, then let's move forward as-is! Thanks for explaining!

akus062381 (approved):

PR #1908, opened 2024-03-28
Support job deletion through job header

marshmalien:
**Summary**
Add job delete support to job header and component tests. 

When a job is running, we should enable the cancel button and disable the delete button. 
When a job is finished, we should disable the cancel button and enable the delete button. 

Screenshot 2024-03-28 at 2 02 28 PM


resoluteCoder (approved):
lgtm

PR #1911, opened 2024-03-29
Adds functionality to the discard button

AlexSCorey:
Addresses AAP-21982
Adds functionality to discard a rule.  Should function as a normal cancel button except should not navigate away from the wizard, or from the wizard step.

Note** If no rules exist and a user clicks discard rule, the form should just reset but should not disappear.

### DIFF
--- a/frontend/awx/views/schedules/components/RuleForm.tsx
+++ b/frontend/awx/views/schedules/components/RuleForm.tsx
@@ -1,7 +1,7 @@
 import { useFormContext } from 'react-hook-form';
 import { PageFormSection } from '../../../../../framework/PageForm/Utils/PageFormSection';
 import { PageFormSelect, PageFormTextInput } from '../../../../../framework';
-import { RuleFields, RuleListItemType, ScheduleFormWizard } from '../types';
+import { RuleFields, RuleListItemType, RuleType, ScheduleFormWizard } from '../types';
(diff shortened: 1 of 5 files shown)


frontend/awx/views/schedules/components/RuleForm.tsx
```diff
@@ -259,7 +259,10 @@ export function RuleForm(props: {
           variant="secondary"
           isDanger
           onClick={() => {
-            reset({ keepDefaultValues: true });
+            const { rules = [], exceptions = [] } = getValues() as RuleFields;
+            const ruleType = props.title === t('Define occurrences') ? 'rules' : 'exceptions';
```
nixocio, line 263: If there is a typo in the title? 
What about adding a ruleType prop?

```
export enum RuleType {
  Rules = 'rules',
  Exceptions = 'exceptions',
}
```

```
 ruleType: RuleType; 

```

matoval (changes requested):
The kebab of the rules list is hidden behind the scroll bar:
[Screencast from 2024-04-10 11-12-07.webm](https://github.com/ansible/ansible-ui/assets/53712446/4036b4db-47d0-45f3-ba7b-d6a6e988218d)

matoval (approved):
Looks great!

PR #1927, opened 2024-04-02
Controller => Inventories tests- add Structure and Assertion Notes

akus062381:
Purpose of this PR:

1) Move Inventories Groups tests off the Inventories Spec file and over to the Inventories Group spec file
2) Add commented notes to indicate where assertions need to be added
3) Add test stubs for missing test cases so an engineer can come back later and write the missing tests

### DIFF
--- a/cypress/e2e/awx/resources/inventories.cy.ts
+++ b/cypress/e2e/awx/resources/inventories.cy.ts
@@ -6,7 +6,7 @@ import { Label } from '../../../../frontend/awx/interfaces/Label';
 import { Organization } from '../../../../frontend/awx/interfaces/Organization';
 import { User } from '../../../../frontend/awx/interfaces/User';
 
-describe('inventories', () => {
+describe('Inventories Tests', () => {
   let organization: Organization;
   let inventory: Inventory;
   let instanceGroup: InstanceGroup;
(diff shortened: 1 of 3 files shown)


matoval (approved):
LGTM

cypress/e2e/awx/resources/inventories.cy.ts
```diff
@@ -44,200 +44,175 @@ describe('inventories', () => {
...
+      cy.clickLink(/^Create inventory$/);
+      //Assert that user is on the form view to create an inventory
+      cy.get('[data-cy="name"]').type(inventoryName);
+      cy.singleSelectByDataCy('organization', organization.name);
+      cy.get('[data-cy="prevent_instance_group_fallback"]').click();
+      cy.clickButton(/^Create inventory$/);
+      //Add an interception call for the newly created inventory, which will allow for the deletion at the end of the test
+      cy.verifyPageTitle(inventoryName);
+      cy.hasDetail(/^Organization$/, organization.name);
+      cy.hasDetail(/^Enabled options$/, 'Prevent instance group fallback');
```
nixocio, line 63: There are several other fields that we could verify here as well, this might avoid future regressions of removing a field or modifying any logic.

image

cypress/e2e/awx/resources/inventories.cy.ts
```diff
@@ -44,200 +44,175 @@ describe('inventories', () => {
...
+      cy.clickPageAction('delete-inventory');
+      cy.get('#confirm').click();
+      cy.clickButton(/^Delete inventory/);
```
nixocio, line 66: Nice clean up.

nixocio (commented):
@akus062381 overall good job. A couple of suggestions.

PR #1934, opened 2024-04-03
EDA Project wizard submissions

keithjgrant:
Adds form submission to EDA Project add teams and add users wizards

Screenshot 2024-04-03 at 10 23 01 AM


vidyanambiar (approved):
Works well 👍 

Since will need a very similar submit action for the remaining EDA resource types (credential, decision env, rulebook activation), I wonder if it would help to create a common function/hook for submit `useSubmitEDARolesForResource`. Something that takes in the postRequest options and onClose callbacks and reads teams and users from the wizard data. WDYT?

PR #1935, opened 2024-04-03
[AAP-19501] Constructed Inventory Host Details

prat98:
This PR updates the inventory host details section tabs and fields. Smart and constructed inventories display an extra `enabled` field on the details page. In addition, the tabs in the inventory hosts section are different. Only the details tab is shown for smart and constructed inventories; all other tabs are hidden and only shown for regular inventories.


MilanPospisil (approved):
LGTM :+1:

PR #1936, opened 2024-04-03
Creates shared directory for workflow visualizer wizard and schedules wizard

AlexSCorey:
This addresses AAP-2228 (partially).  It creates the shared directory and implements shared components for step 1 of the wizards.  The remaining steps will follow on.

There is no change in functionality


cypress/e2e/awx/resources/workflowJobTemplates.cy.ts
```diff
@@ -147,56 +145,36 @@ describe('Workflow Job templates form', () => {
           this.globalProject as Project
         ).then((projectNode) => {
           cy.createAwxWorkflowVisualizerJobTemplateNode(workflowJobTemplate, jobTemplate).then(
-            (jobTemplateNode) => {
```
AlexSCorey, line 150: These lines were removed because they felt like unnecessary complexity in this test....adding nodes and links via the api to a workflow visualizer, for the purpose of ensuring the job is launch and renders the output properly.  Removing these lines simplifies and speeds up the test.

frontend/awx/common/SharedWizard/ResourceSelectionStep.tsx
```diff
@@ -0,0 +1,256 @@
+import { Dispatch, SetStateAction, useEffect } from 'react';
```
AlexSCorey, line 1: File formerly known as NodeTypeStep that lived in WorkflowVisualizer/wizard.

frontend/awx/resources/templates/WorkflowVisualizer/wizard/NodeTypeStep.tsx
```diff
@@ -1,644 +0,0 @@
...
-    nodeType,
-    setValue,
-    setVisibleSteps,
-    setWizardData,
-    stepData,
-  ]);
-
-  return (
-    <>
-      {pathname.split('/')[1] === 'schedules' ? (
```
matoval, line 246: These checks are missing from this refactor and has removed conditional code that renders schedule specific inputs for the first step of the schedule wizard.

matoval (changes requested):
These changes have removed conditional code that renders schedule specific inputs for the first step of the schedule wizard.

frontend/awx/resources/templates/WorkflowVisualizer/wizard/NodeTypeStep.tsx
AlexSCorey: This file has been moved, and broken down into smaller files.  It, and its parts now live in /common/SharedWizard 
Here is what that same view looks like on my branch for me.  
Screenshot 2024-04-04 at 7 03 01 PM

matoval: When I open the schedule wizard from the template schedules page not rendering correctly:
image

PR #1939, opened 2024-04-03
[AAP-19507] Constructed Inventory Group Hosts

prat98:
This PR contains updates to the inventory -> group -> hosts section of the new UI which now accounts for the constructed inventory type. 

The `GroupHosts` list component now dynamically hides toolbar and row actions depending on the inventory type (either constructed or regular inventory in this case as smart inventories will not be able to route to this page). 

In addition this, small changes include updating the `Run Command` button so it is enabled regardless of a host being selected on the page and as long as the user has the correct permissions to run a command against an inventory. 

Another small change includes removing the page actions from the `GroupPage` component to be in parity with the old UI.

***Updating component tests will move out of draft once that is done***

### DIFF
--- a/frontend/awx/resources/inventories/hooks/useInventoriesGroupsHostsToolbarActions.tsx
+++ b/frontend/awx/resources/inventories/hooks/useInventoriesGroupsHostsToolbarActions.tsx
@@ -59,13 +59,14 @@ export function useInventoriesGroupsHostsToolbarActions(view: IAwxView<AwxHost>)
     [setDialog, params.group_id, view, alertToaster, t]
   );
 
-  return useMemo<IPageAction<AwxHost>[]>(
-    () => [
+  return useMemo<IPageAction<AwxHost>[]>(() => {
+    const toolbarActions: IPageAction<AwxHost>[] = [
       {
         type: PageActionType.Dropdown,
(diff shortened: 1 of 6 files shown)


frontend/awx/resources/inventories/hooks/useInventoriesGroupsHostsToolbarActions.tsx
```diff
@@ -59,13 +59,14 @@ export function useInventoriesGroupsHostsToolbarActions(view: IAwxView<AwxHost>)
...
-  return useMemo<IPageAction<AwxHost>[]>(
-    () => [
+  return useMemo<IPageAction<AwxHost>[]>(() => {
+    const toolbarActions: IPageAction<AwxHost>[] = [
       {
         type: PageActionType.Dropdown,
         selection: PageActionSelection.None,
         variant: ButtonVariant.primary,
         isPinned: true,
+        isHidden: () => params.inventory_type === 'constructed_inventory',
```
MilanPospisil, line 69: Thats good, I did not noticed that there is isHidden property, useful :-)

MilanPospisil (approved):
LGTM :+1 If the tests passes.

PR #1941, opened 2024-04-04
Access roles wizard from Project Team Access page

vidyanambiar:
This PR adjusts route names to link the EDA Project Team Access UI with the Add Roles wizard for adding teams to a project.

cc: @keithjgrant @ZitaNemeckova

### DIFF
--- a/frontend/eda/projects/components/EdaProjectAddTeams.tsx
+++ b/frontend/eda/projects/components/EdaProjectAddTeams.tsx
@@ -116,7 +116,7 @@ export function EdaProjectAddTeams() {
           resolve();
         },
         onClose: () => {
-          pageNavigate(EdaRoute.ProjectDetails, {
+          pageNavigate(EdaRoute.ProjectTeamAccess, {
             params: { id: project.id.toString() },
           });
         },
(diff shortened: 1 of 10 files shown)


frontend/eda/projects/components/EdaProjectAddTeams.tsx
```diff
@@ -136,12 +136,19 @@ export function EdaProjectAddTeams() {
...
         ]}
       />
-      <PageWizard<WizardFormValues> steps={steps} onSubmit={onSubmit} disableGrid />
+      <PageWizard<WizardFormValues>
+        steps={steps}
+        onSubmit={onSubmit}
+        disableGrid
+        onCancel={() => {
+          pageNavigate(EdaRoute.ProjectTeamAccess, { params: { id: project?.id } });
+        }}
```
nixocio, line 150: @vidyanambiar I believe there is a prop errorAdapter specific per project.

nixocio: @vidyanambiar clarified the use case here. It does not seem to be required.

PR #1943, opened 2024-04-04
[AAP-21055] Add dynamic filters to execution environments page

daphnemaeve:
This PR updates the execution environments page to use dynamic filters and fixes affected tests.

Jira issue: https://issues.redhat.com/browse/AAP-21055

Before:
Screenshot 2024-04-17 at 1 59 24 PM

After:
Screenshot 2024-04-17 at 1 58 44 PM

### DIFF
--- a/cypress/support/table-commands.ts
+++ b/cypress/support/table-commands.ts
@@ -39,6 +39,7 @@ Cypress.Commands.add('selectTableFilter', (dataCy: string) => {
   // We need to use document() to get the body of the page
   // and then find the filter-select within that body
   // This fixes the issue where this command would fail inside a within() block
+  cy.document().its('body').find('#filter-search').type(dataCy.replaceAll('-', ' '));
   cy.document()
     .its('body')
     .find('#filter-select')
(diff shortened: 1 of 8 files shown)


cypress/support/table-commands.ts
```diff
@@ -39,6 +39,7 @@ Cypress.Commands.add('selectTableFilter', (dataCy: string) => {
   // We need to use document() to get the body of the page
   // and then find the filter-select within that body
   // This fixes the issue where this command would fail inside a within() block
+  cy.document().its('body').find('#filter-search').type(dataCy.replaceAll('-', ' '));
```
kialam, line 42: Remove - this isn't being called anywhere in the tests.

kialam (changes requested):

cypress/support/table-commands.ts
skochay330: The `selectTableFilter` command is called in the `filterTableByTextFilter` command which is used here https://github.com/ansible/ansible-ui/pull/1943/files#diff-70b1ba24eeb1e57585230c8e46e8acf6752d0316bc172f916da27340543eff49R52. This change was necessary in order for the filter to be visible in cypress.

kialam (approved):

PR #1950, opened 2024-04-05
[AAP-8292] Smart Inventory Details

prat98:
This PR updates the inventory details section to account for smart inventories. The work involved hiding fields that were specific to constructed inventories, showing the smart host filter field for smart inventories and showing common fields for all types of inventories. In addition the inventory details component tests have been updated to account for the constructed and smart inventory views.

### DIFF
--- a/frontend/awx/resources/inventories/InventoryPage/InventoryDetails.cy.tsx
+++ b/frontend/awx/resources/inventories/InventoryPage/InventoryDetails.cy.tsx
@@ -3,21 +3,84 @@ import { Inventory } from '../../../interfaces/Inventory';
 import { InventoryDetailsInner as InventoryDetails } from './InventoryDetails';
 
 describe('InventoryDetails', () => {
-  it('Component renders and displays inventory', () => {
-    cy.fixture('inventory').then((inventory: Inventory) => {
-      cy.mount(<InventoryDetails inventory={inventory} />);
-      cy.get('#name').should('have.text', 'test inventory');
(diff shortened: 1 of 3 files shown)


frontend/awx/resources/inventories/InventoryPage/InventoryDetails.cy.tsx
```diff
@@ -3,21 +3,84 @@ import { Inventory } from '../../../interfaces/Inventory';
...
+          }
+        })
+        .then((inventory: Inventory) => {
+          cy.mount(<InventoryDetails inventory={inventory} />);
+          cy.get('#name').should('have.text', 'test inventory');
+          cy.get('#organization').should('have.text', 'Default');
+          cy.get('#total-hosts').should('have.text', '1');
+          cy.get('#labels').should('have.text', 'test label');
+          cy.get(
+            '#created > .pf-v5-c-description-list__text > .date-time > .pf-v5-c-button'
```
MilanPospisil, line 37: Please dont use those selectors. First reasons are the updates of patternfly library, that sometimes enforces us to change things in the framework to match new classes (you can see for example v5 string which is for version 5), thus we will have to rewrite the tests too. Try to use data-cy, possibly aria label or element type (button, tr...).

prat98: Good catch updated

MilanPospisil (approved):
LGTM :+1:

PR #1955, opened 2024-04-08
[AAP-17646] Include Workflow Approval List view tests

fullsushidev:
relates to AAP-17646

### DIFF
--- a/cypress/e2e/awx/administration/workflowApproval.cy.ts
+++ b/cypress/e2e/awx/administration/workflowApproval.cy.ts
@@ -375,10 +375,6 @@ function bulkActAssertAndDeleteWorkflowApproval(selectorDataCy: 'approve' | 'den
         cy.clickKebabAction('actions-dropdown', 'delete');
       });
       cy.actionsWFApprovalConfirmModal('delete');
-      cy.wait('@deleteWFApprov')
-        .its('response')
-        .then((response) => {
-          expect(response?.statusCode).to.eql(204);
-        });
+      cy.waitTimes('@deleteWFApprov', numOfBulkActions, 204);
     });
 }
--- a/cypress/e2e/awx/administration/workflowApproval.cy.ts
+++ b/cypress/e2e/awx/administration/workflowApproval.cy.ts
@@ -11,8 +11,8 @@ import { WorkflowNode } from '../../../../frontend/awx/interfaces/WorkflowNode';
 import { WorkflowApproval } from '../../../../frontend/awx/interfaces/WorkflowApproval';
 
 const wFApprovURL = '/administration/workflow-approvals/';
-const wFApprovName = 'E2E WFJT Approval Request ' + randomString(4);
-const numOfBulkActions = 5;
+const wFApprovName = 'E2E WFJT Approval Request ' + randomString(4); // custom name to make it easier to filter for this specific test spec
+const numOfBulkActions = 5; // times to execute the jobs
 
 describe('Workflow Approvals', () => {
   let organization: Organization;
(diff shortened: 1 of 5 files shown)


cypress/e2e/awx/administration/workflowApproval.cy.ts
```diff
@@ -7,18 +8,255 @@ import { User } from '../../../../frontend/awx/interfaces/User';
...
   selectorDataCy: 'approve' | 'deny' | 'cancel',
-  wfaID: string
+  wFApprovID: string
 ) {
   let actionStatusCode: number;
   let statusText: string;
   let actionURL: string;
 
-  cy.visit(wfaURL);
+  cy.visit(wFApprovURL);
```
vishsanghishetty, line 259: @fullsushidev  any reason for `cy.visit` and not `navigate`

cypress/e2e/awx/administration/workflowApproval.cy.ts
```diff
@@ -62,174 +300,81 @@ function actAssertAndDeleteWorkflowApproval(
...
-  let userWFApprove: User;
-  let userWFDeny: User;
-  let userWFCancel: User;
-  let inventory: Inventory;
-  let inventorySource: InventorySource;
-  let jobTemplate: JobTemplate;
-  let workflowJobTemplate: WorkflowJobTemplate;
-  let wfjobTemplateNode: WorkflowNode;
-  let approvalWFNode: WorkflowNode;
+  cy.visit(wFApprovURL);
```
vishsanghishetty, line 320: same as step 259

cypress/e2e/awx/administration/workflowApproval.cy.ts
vishsanghishetty: `cy.navigateTo('awx', 'workflow-approvals')` should do the same as `cy.visit` @fullsushidev

fullsushidev: It was to make it easier since this piece of code will repeat a lot in this spec, although there's only 2 uses now.

fullsushidev:
> I have tried 3 times to run the whole spec file and it keeps crashing the GUI :-(

@akus062381 I usually change `numTestsKeptInMemory` to 2 (instead of 20) to avoid having this issue locally.
(but I've been thinking about possible changes to enhance the memory usage by this spec.)

akus062381 (approved):
Per our conversation, you will add some commented lines giving explanation of the custom functions. Ok to merge after!

vishsanghishetty:
@fullsushidev I see that you resolved the comment so putting it out here.

using `cy.visit` reloads the whole page which is slow
I suggested to use what I wrote above, and yes, I see that the functions you wrote are reusable `actAssertAndDeleteWorkflowApproval` `bulkActAssertAndDeleteWorkflowApproval`

just add this `cy.navigateTo('awx', 'workflow-approvals')` in place of `cy.visit` in the above 2 functions.

vishsanghishetty (approved):

fullsushidev:
Hey @vishsanghishetty
I answered you inside the resolved comment.
Adding more context here: in this case, the intention is indeed to reload, since cypress is way faster than this UI. I tried as much as I could to delay cypress by including assertions of API statuses and UI elements state, also with `reload()` and `visit()`.
I'll add a comment to the code to revisit this approach later on, and see what else can be improved to decrease memory usage and time spent.
Thank you

cypress/e2e/awx/administration/workflowApproval.cy.ts
jamestalton: Hard coding the urls is going to make it hard for us to get the upstream tests working against the downstream platform when we get to that point as the urls change in the platform. using cy.navigate will handle this properly for us.

fullsushidev:
Force pushed to fix rebase conflicts. PR set to auto-merge after pipeline pass.

PR #1956, opened 2024-04-08
constructed inventory  related groups

MilanPospisil:
Modify GroupRelatedGroups component to support constructed inventory - thus not display any actions besides run command.

image


prat98 (approved):
Changes look good! 🚀

PR #1965, opened 2024-04-08
Support for substeps in PageWizard

vidyanambiar:
image


frontend/awx/views/schedules/wizard/ScheduleAddWizard.cy.tsx
```diff
@@ -112,6 +112,7 @@ describe('ScheduleAddWizard', () => {
           'Schedule name is required.'
         );
       });
+      cy.pause();
```
nixocio, line 115:
```suggestion

```

nixocio (changes requested):

framework/PageWizard/types.ts
```diff
@@ -15,6 +15,13 @@ export interface PageWizardStep {
   validate?: (formData: object, wizardData: object) => Promise<void> | void;
 }
 
+/** Type used to define parent steps. */
+export interface PageWizardParentStep extends PageWizardBasicStep {
```
jamestalton, line 19: export interface PageWizardParentStep extends Omit<PageWizardBasicStep, "inputs" | "validate"> {

frontend/awx/views/schedules/wizard/ScheduleAddWizard.cy.tsx
vidyanambiar: Thanks!!

framework/PageWizard/types.ts
vidyanambiar: Good catch! Updated 👍

framework/PageWizard/PageWizardNavigation.tsx
```diff
@@ -47,6 +69,50 @@ export function PageWizardNavigation() {
                   </span>
                 )}
               </button>
+              {
+                /** Nav for substeps */
+                isPageWizardParentStep(step) ? (
+                  <ol className="pf-v5-c-wizard__nav-list">
```
keithjgrant, line 75: I'd consider rendering this bit in a sub component just to help un-nest a bunch of the conditionals and keep this section more purely JSX; you can just define it below in the same file.

framework/PageWizard/PageWizardBody.tsx
```diff
@@ -40,54 +38,55 @@ export function PageWizardBody<T>({
...
       if (isLastStep) {
         return onSubmit(wizardData as T);
       }
 
-      const activeStepIndex = filteredSteps.findIndex((step) => step.id === activeStep?.id);
-      const nextStep = filteredSteps[activeStepIndex + 1];
+      const activeStepIndex = visibleStepsFlattened.findIndex((step) => step.id === activeStep?.id);
+      // If the next step is a parent step, mark its first substep as the next active step
+      const nextStep = isPageWizardParentStep(visibleStepsFlattened[activeStepIndex + 1])
+        ? visibleStepsFlattened[activeStepIndex + 2]
```
keithjgrant, line 58: I think this may throw an error if substeps is an empty array. (Is there a way to ensure the substeps array is at least length one in the type definition?)

keithjgrant (approved):
Looks great, just a couple thoughts

framework/PageWizard/PageWizardNavigation.tsx
vidyanambiar: Thanks! I added a separate component called `SupstepsNav` 👍

framework/PageWizard/PageWizardBody.tsx
vidyanambiar: Great catch! I updated the type for `substeps` to require at least one item and verified that TS throws an error on an empty array for `substeps`.

PR #1969, opened 2024-04-09
Fix Hosts detail inventory link

MilanPospisil:
Inventory link in host detail did not filled the :inventory_type props, so it then failed to load the hosts and groups in tabs.


prat98 (approved):
Good catch this looks good to me 🚀

PR #1970, opened 2024-04-09
implemented question order

resoluteCoder:
Jira: https://issues.redhat.com/browse/AAP-22464

Note: I believe the current API for survey spec does not support a hiding feature. So I did not implement the checkbox in the view.

Interestingly enough, the dashboard view works with local storage and does not have an on submit function.


frontend/awx/resources/templates/hooks/useManageSurveyQuestions.tsx
```diff
@@ -0,0 +1,54 @@
...
+    ],
+    [t]
+  );
+
+  const { data, refresh } = useGet<Survey>(endpoint);
+  const { openManageItems: openManageQuestionOrder } = useManageItems({
+    id: `survey-question-template-${params.id}`,
+    title: 'Manage question order',
+    description: t(
+      'Hide or show the panels you want to see on the overview page by selecting or unselecting, respectively. The panels are ordered from top to bottom on the list. Use the draggable icon :: to re-order your view.'
```
marshmalien, line 30:
```suggestion
      'To reorder the survey questions drag and drop them in the desired location.'
```

marshmalien (approved):
Looks good! I added some small suggestions.

frontend/awx/resources/templates/hooks/useManageSurveyQuestions.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const { t } = useTranslation();
+  const params = useParams<{ id: string }>();
+  const postRequest = usePostRequest();
+  const endpoint = awxAPI`/job_templates/${params.id?.toString() ?? ''}/survey_spec/`;
+  const columns = useMemo(
+    () => [
+      {
+        header: t('Survey questions'),
+        cell: (spec: Spec) => spec.question_name,
+      },
```
marshmalien, line 20: Could we add a second column, "Defaults"  or "Default Answer(s)", to keep feature parity with the old ui?

PR #1971, opened 2024-04-09
Add User Access tab to credentials, dec env, project and rulebook actv

ZitaNemeckova:
- Create a common component `Access` and pass params based on User/Team component.
- Add User Access tab to credentials, dev. env., project and rulebook activation

### DIFF
--- a/frontend/eda/access/common/TeamAccess.tsx
+++ b/frontend/eda/access/common/TeamAccess.tsx
@@ -15,7 +15,7 @@ export function TeamAccess(props: { id: string; type: string; addRolesRoute?: st
           label: t('Team name'),
         },
       }}
-      toolbarFiltersValues={{ label: '', query: '' }}
+      toolbarFiltersValues={{ label: t('Team name'), query: 'team__name' }}
       url={edaAPI`/role_team_assignments/`}
       id={id}
       content_type_model={type}
--- a/frontend/eda/access/common/TeamAccess.tsx
+++ b/frontend/eda/access/common/TeamAccess.tsx
@@ -20,7 +20,7 @@ export function TeamAccess(props: { id: string; type: string; addRolesRoute?: st
       id={id}
       content_type_model={type}
       addRolesRoute={addRolesRoute}
-      translatedType={t('team')}
+      type={'team'}
     />
   );
 }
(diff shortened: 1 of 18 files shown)


frontend/eda/access/common/TeamAccess.tsx
```diff
@@ -1,176 +1,26 @@
...
+          sort: 'team__name',
+          label: t('Team name'),
+        },
+      }}
+      toolbarFiltersValues={{ label: '', query: '' }}
+      url={edaAPI`/role_team_assignments/`}
+      id={id}
+      content_type_model={type}
+      addRolesRoute={addRolesRoute}
+      translatedType={t('team')}
```
himdel, line 23: these params look great! :+1:  (well, hub namespaces will need a custom onSave but that's a problem for later :))

but... `translatedType` not so much .. plurals won't work, and neither will any type of declinations

What about passing in `'team' | 'user'` (not translated) and having the full strings and

```diff
-        label: t('Add {{type}}', { type: props.translatedType }),
+        label: translatedType === 'user' ? t`Add user` : t`Add team`,
```

etc

ZitaNemeckova: Fixed

frontend/eda/access/common/TeamAccess.tsx
```diff
@@ -1,176 +1,26 @@
...
-      defaultSubtitle={t('Team access')}
+    <Access
+      tableColumnFunctions={{
+        name: {
+          function: (teamAccess: TeamAssignment) => teamAccess.summary_fields.team.name,
+          sort: 'team__name',
+          label: t('Team name'),
+        },
+      }}
+      toolbarFiltersValues={{ label: '', query: '' }}
```
vidyanambiar, line 18: The `Team Access` filters do not appear to be working. We probably need to provide initial values here? 
`toolbarFiltersValues={{ label: t('Team name'), query: 'team__name' }}`

vidyanambiar (commented):
I added a comment about filters. Other than that this looks great to me 👍

frontend/eda/access/common/TeamAccess.tsx
ZitaNemeckova: Fixed. Thanks!

vidyanambiar (approved):

PR #1973, opened 2024-04-09
[AAP-8279] Add initial work for credential sub forms

nixocio:
Add initial work for credential sub forms

image

There are several edge cases to be covered and remaining work related to this feature. I will break down in smaller PRs to ease review, and implementation.


marshmalien (approved):
Looks good! Small thing I noticed is that choice field should probably show default text instead of "undefined". 

Screenshot 2024-04-09 at 12 35 39 PM

PR #1982, opened 2024-04-09
Add dynamic filters to users page

daphnemaeve:


frontend/awx/access/users/Users.cy.tsx
```diff
@@ -0,0 +1,75 @@
...
+              type: ToolbarFilterType.MultiText,
+              required: true,
+              label: 'Name',
+              max_length: 512,
+              help_text: 'Name of this user.',
+              filterable: true,
+            },
+          },
+        },
+      },
```
kialam, line 56: It is unclear to me what this stub does/affects user permissions. Why are we importing `ToolbarFilterType` here?

kialam (changes requested):

frontend/awx/access/users/Users.cy.tsx
daphnemaeve: That stub is used to ensure the user is able to create users and that the create user button is enabled. I can play around with figuring out if importing ToolbarFilterType is necessary anymore

kialam (approved):
Should be good to go after we remove `type` from the stub. 👍

PR #1983, opened 2024-04-09
Fix schedules workflow test

matoval:
Fix schedules workflow test by raising limit to 500


frontend/awx/resources/templates/components/PageFormWorkflowJobTemplateSelect.tsx
```diff
@@ -47,7 +47,7 @@ export function PageFormWorkflowJobTemplateSelect<
       loadingPlaceholder={t('Loading job templates...')}
       loadingErrorText={t('Error loading job templates')}
       isRequired={props.isRequired}
-      limit={200}
+      limit={500}
```
jamestalton, line 50: API only supports 200. Cannot change this to 500.

jamestalton (changes requested):

PR #1994, opened 2024-04-10
implemented survey step to template on launch

resoluteCoder:
Jira: https://issues.redhat.com/browse/AAP-15011


jerabekjiri:
I'm getting this on creating new `Job template` schedule 
```
ErrorBoundary.tsx:24 Uncaught error: TypeError: element.choices?.map is not a function
    at eval (ScheduleSurveyStep.tsx:22:24)
    at Array.map (<anonymous>)
    at ScheduleSurveyStep (ScheduleSurveyStep.tsx:20:21)
    at renderWithHooks (react-dom.development.js:16305:1)
    at updateFunctionComponent (react-dom.development.js:19588:1)
    at beginWork (react-dom.development.js:21601:1)
    at beginWork$1 (react-dom.development.js:27426:1)
    at performUnitOfWork (react-dom.development.js:26557:1)
    at workLoopSync (react-dom.development.js:26466:1)
```

jerabekjiri:
Finish Prompt on launch with multiselect option I get `Danger alert:Failure to launch Bad Request`
```
Request URL:
https://localhost:4101/api/v2/job_templates/7/launch/
Request Method:
POST
Status Code:
400 Bad Request
{"labels":[],"extra_vars":"\nvar_name: undefined"}
```

marshmalien (changes requested):
We should merge the survey step data into extra_vars on the review step (TemplateLaunchReviewStep) similar to how you're accomplishing it on save. 

Something like this:
```
let extraVarDetails = extra_vars || '{}';
if (survey) {
  const jsonObj: { [key: string]: string } = {};

  if (extra_vars) {
    const lines = extra_vars.split('\n');
    lines.forEach((line) => {
      const [key, value] = line.split(':').map((part) => part.trim());
      jsonObj[key] = value;
    });
  }

  const mergedData: { [key: string]: string } = {
    ...jsonObj,
    ...survey,
  };

  extraVarDetails = jsonToYaml(JSON.stringify(mergedData));
}

...code...

<PageDetailCodeEditor label={t('Extra vars')} value={extraVarDetails} />

```

We should also probably mask the password fields before showing them in the review step. 
```
export function maskPasswords(vars, passwordKeys) {
  const updated = { ...vars };
  passwordKeys.forEach((key) => {
    if (typeof updated[key] !== 'undefined') {
      updated[key] = '········';
    }
  });
  return updated;
}
```

@resoluteCoder and I chatted and agreed to move ScheduleSurveyStep to a common location and remove "Schedule" from its name since it will be shared across multiple awx views.

resoluteCoder:
@jerabekjiri Those errors should be fixed with this rendition of the code. As far as the UX issue that seems to an issue with all those kinds of inputs where `enter` selects `next` by default. We could bring that up with UX or create a separate ticket. 

@marshmalien I implemented the suggestion as well as moving the file to a more common location.

frontend/awx/common/SurveyStep.tsx
```diff
@@ -1,21 +1,20 @@
...
-export function ScheduleSurveyStep() {
+export function SurveyStep({ templateId }: { templateId?: string }) {
   const { t } = useTranslation();
   const { wizardData } = usePageWizard();
   const { node_resource } = wizardData as WizardFormValues;
-  const { data: survey_spec } = useGet<Survey>(
-    awxAPI`/job_templates/${node_resource ? node_resource.id.toString() : ''}/survey_spec/`
-  );
+  const id = node_resource ? node_resource.id.toString() : templateId ? templateId : '';
+  const { data: survey_spec } = useGet<Survey>(awxAPI`/job_templates/${id}/survey_spec/`);
```
jerabekjiri, line 17: This should support `workflow_job_templates` as well

PR #2001, opened 2024-04-10
E2E InventoryHostForm disable editing of variables as the monaco edit…

jamestalton:
…or is flaky


kialam (commented):
The assertion fails on the Description field and not the variables field.

PR #2006, opened 2024-04-11
Skip inventoryHostForm edit flaky test cases.

kialam:


jamestalton:
Should not need to skip this test after the fix in 
https://github.com/ansible/ansible-ui/pull/2001

PR #2007, opened 2024-04-11
Page navigation routes fix routes being available in modals

jamestalton:


prat98 (approved):
Tested changes with other branch https://github.com/ansible/ansible-ui/tree/prat-mgmt-job-retain-data-modal and modal reroute works after job launch now. LGTM! 🚀

PR #2009, opened 2024-04-11
Fix flaky tests: AWX Org and AWX JT

akus062381:


cypress/support/awx-commands.ts
```diff
@@ -483,7 +480,7 @@ Cypress.Commands.add('selectDetailsPageKebabAction', (dataCy: string) => {
   cy.get('[data-cy="actions-dropdown"]')
     .click()
     .then(() => {
-      cy.getByDataCy(dataCy).click();
+      cy.get(`[data-cy="${dataCy}"]`).click();
```
jamestalton, line 483: Sometimes actions in the kebab are disabled and we need to use the getBy to wait until enabled before clicking

jamestalton (changes requested):

jamestalton (approved):

PR #2010, opened 2024-04-11
Brings in correct breadcrumbs for schedules

AlexSCorey:
This addresses AAP-22719.  

It also cleans up a bunch of code related to Rules Tabs, and old implementations of rules on the form that are no longer needed.


acruzgon (approved):
ran locally and worked as expected!

PR #2014, opened 2024-04-12
notifications detail

MilanPospisil:


akus062381:
I am wondering if we should add form field validation? For example when a user types an email address?

MilanPospisil:
> I am wondering if we should add form field validation? For example when a user types an email address?

There is. But its activated after form submit, then it starts complain if email is wrong.

akus062381 (approved):
Looks good!

PR #2033, opened 2024-04-15
Removes schedule references from NodeTypeStep

AlexSCorey:
This PR creates a new wizard step component for schedules and removes all references to schedules from NodeTypeStep


marshmalien (approved):
I pulled down and manually tested the workflow node wizard and prompt on launch. LGTM!

PR #2034, opened 2024-04-15
Timezone selector has filter feature now

AlexSCorey:
This addresses 22618.


matoval (approved):
LGTM, run locally and worked as expect.

ScheduleEditWizard.cy.tsx might need to be updated like ScheduleAddWizard.cy.tsx

PR #2050, opened 2024-04-16
Notifications add edit form 2

MilanPospisil:
Added new form options in notification type dropdown:
Webhook, Mattermost, Rocketchat, IRC

Additionaly - InnerForm component was moved to its own file to split the big file.
Password handling - server returns string $ encrypted $, which is then hold in form context and the variable must be deleted, so its not sent back to the server (in edit mode), thus changing the password.

TODO (will be done in next PR):
Customize messages form part
Form layout from left to right, not from top to bottom in single column


RedOrNot (approved):
LGTM

PR #2051, opened 2024-04-16
Repair host add - inventory selector restriction to only normal inventory

MilanPospisil:
…tories


prat98 (approved):
Good catch! Changes look good to me! 🚀

PR #2067, opened 2024-04-16
[AAP-22269] Initial implementation of user tokens tab

tznamena:
This PR adds empty user tokens tab to user page but only if the user id on URL matches currently logged in user. If the URL is modified to point to another user, user will be navigated back to Users page.
Its an initial implementation that does not render any tokens yet, that will come in subsequent PRs.

### DIFF
--- a/frontend/awx/access/users/UserPage/UserTokens.tsx
+++ b/frontend/awx/access/users/UserPage/UserTokens.tsx
@@ -14,7 +14,8 @@ export function UserTokens() {
   useEffect(() => {
     if (activeAwxUser?.id === undefined || activeAwxUser?.id.toString() !== params.id) {
       setShowTokens(false);
-      pageNavigate(AwxRoute.Users);
+      // redirect to user details for the active/logged-in user
+      pageNavigate(AwxRoute.UserDetails, { params: { id: activeAwxUser?.id } });
     }
   }, [activeAwxUser?.id, params.id, setShowTokens, pageNavigate]);
 
--- a/frontend/awx/access/users/UserPage/UserTokens.tsx
+++ b/frontend/awx/access/users/UserPage/UserTokens.tsx
@@ -1,26 +1,23 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
-import { useEffect } from 'react';
+import { useEffect, useState } from 'react';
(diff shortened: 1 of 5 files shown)


frontend/awx/access/users/UserPage/UserTokens.tsx
```diff
@@ -0,0 +1,23 @@
...
+export function UserTokens() {
+  const params = useParams<{ id: string }>();
+  const [showTokens, setShowTokens] = useState(true);
+  const { activeAwxUser } = useAwxActiveUser();
+  const pageNavigate = usePageNavigate();
+
+  useEffect(() => {
+    if (activeAwxUser?.id === undefined || activeAwxUser?.id.toString() !== params.id) {
+      setShowTokens(false);
+      pageNavigate(AwxRoute.Users);
```
vidyanambiar, line 17: It might be better to remain in the context of the already selected user. Can we navigate instead to `AwxRoute.UserDetails` passing id as a param?

vidyanambiar (commented):
Looks good!

frontend/awx/access/users/UserPage/UserTokens.tsx
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
+import { useEffect, useState } from 'react';
+import { useParams } from 'react-router-dom';
+import { usePageNavigate, PageNotImplemented } from '../../../../../framework';
+import { AwxRoute } from '../../../main/AwxRoutes';
+import { useAwxActiveUser } from '../../../common/useAwxActiveUser';
+
+export function UserTokens() {
+  const params = useParams<{ id: string }>();
+  const [showTokens, setShowTokens] = useState(true);
```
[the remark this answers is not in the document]
tznamena: Not needed for normal function however it makes component testing easier and less complex. When the user Id from the URL does not match active user (logged in user) the component will not render anything and that's easy to verify in test. Verifying that user gets redirected would make the component test quite complex.

frontend/awx/access/users/UserPage/UserTokens.tsx
tznamena: Good point, I will make that change.

vidyanambiar (approved):

PR #2076, opened 2024-04-17
AAP-19317-inventory-hosts-tabs

RedOrNot:
https://issues.redhat.com/browse/AAP-19317

### DIFF
--- a/cypress/e2e/awx/resources/hosts.cy.ts
+++ b/cypress/e2e/awx/resources/hosts.cy.ts
@@ -29,18 +29,18 @@ describe('host and inventory host', () => {
   });
 
   //tests
-  it('can create, edit and delete a inventory host', () => {
+  it.skip('can create, edit and delete a inventory host', () => {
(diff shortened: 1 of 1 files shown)


cypress/e2e/awx/resources/hosts.cy.ts
```diff
@@ -40,7 +40,14 @@ describe('host and inventory host', () => {
     createAndEditAndDeleteHost(false, inventory);
   });
 
-  it('can work with groups tab', () => {
+  it('can create, edit, assosiat and disassosiate groups at inventory -> hosts -> groups tab', () => {
```
fullsushidev, line 43: it's missing the `skip`
and let's wait for the discussion about how we should organize those repetitive tests since we already have the `inventoryHost.cy.ts` file

fullsushidev (approved):
I'm approving this PR so we can refactor and move the tests between the correct files later on.

PR #2082, opened 2024-04-17
Convert tabbed Groups list to use advanced search.

kialam:
The Groups tabbed view exists in several areas:

- Inventory -> Groups
- Inventory -> Groups -> Related Groups
- Inventory -> Hosts -> Groups

The various filter hooks for these lists have been refactored to use the `useDynamicFilters` hook and we pass a `url` prop to correctly filter down the results.

### DIFF
--- a/frontend/awx/resources/inventories/inventoryHostsPage/InventoryHostGroupsModal.cy.tsx
+++ b/frontend/awx/resources/inventories/inventoryHostsPage/InventoryHostGroupsModal.cy.tsx
@@ -22,7 +22,7 @@ describe('Inventory Host Groups List', () => {
         {
           fixture: 'groups.json',
         }
-      );
+      ).as('getGroupResults');
       cy.intercept(
         { method: 'OPTIONS', url: '/api/v2/groups' },
         { fixture: 'groups_options.json' }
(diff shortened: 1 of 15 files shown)


frontend/awx/resources/inventories/inventoryHostsPage/InventoryHostGroupsModal.cy.tsx
```diff
@@ -30,67 +30,97 @@ describe('Inventory Host Groups List', () => {
...
+        {
+          fixture: 'mock_options.json',
+        }
+      ).as('getFilterOptions');
       cy.mount(<InventoryHostGroupsAddModal {...props} />, {
         path: '/inventories/:inventory_type/:id/hosts/:host_id/*',
         initialEntries: ['/inventories/inventory/1/hosts/1/groups'],
       });
-      cy.get('tbody').find('tr').should('have.length', 10);
+      cy.get('tbody').find('tr').should('have.length', 4);
```
fullsushidev, line 47: The amount of lines comes from the fixture? Might be worth to change it to a const somewhere? WDYT?

fullsushidev (approved):

PR #2111, opened 2024-04-18
Wizard step flow refactor

keithjgrant:
Refactors wizard so that `flattenedSteps` and `visibleSteps` are derived directly from the `steps` array when needed rather than storing them to state variables. This resolves some navigation errors that prevented dynamic steps from being added back in as the form values changed.


AlexSCorey (approved):
Keith and I met offline and he addressed all of my concerns.  I've pulled the code down and tested it and all seems to work as excepted.

PR #2112, opened 2024-04-18
[AAP-8156] Instance Group Instances Tab List

prat98:
This PR is for AAP-8156 which is for creating the instances list inside of the instance group instances tab. The work in this PR includes:

Refactoring the `InstanceForksSlider` component to be more generic and consistent. No more conditional rendering inside of the actual component. If a forkslider needs to be conditionally rendered, it can be done so outside of the component. This makes the code more readable

Added instance errors to the `InstanceDetails` page which is conditionally rendered based on the existence of an instance error

Made a number of reusable pageAction hooks such as `useRunHealthCheckToolbarAction`, `useToggleInstanceRowAction`, `useRunHealthCheckRowAction` and `useEditInstanceRowAction`. Which are now used by `useInstanceRowActions` and `useIGinstanceRowActions` hooks, which are the main hooks responsible for generating the row and toolbar actions for both the instances page and instance group -> instances tab page.

Made a reusable `InstancesList` component which is shared by both instances and  instance group -> instances tab page

Added  instance group -> instances -> details section with routing so clicking on an instance on the  instance group -> instances tab page will direct you to the correct page with the correct route.


frontend/awx/administration/instances/hooks/useInstancesColumns.tsx
```diff
@@ -94,6 +113,16 @@ export function useInstancesColumns(options?: { disableSort?: boolean; disableLi
           ),
         list: 'secondary',
       },
+      {
+        header: t('Running jobs'),
```
MilanPospisil, line 117: It seems that Expanded option is missing in PageTable and I cant see Running and Total jobs, there must be another  switch to turn it on, possibly in PageTable?

frontend/awx/administration/instances/hooks/useInstanceRowActions.tsx
```diff
@@ -34,79 +43,100 @@ export function useInstanceRowActions(onComplete: (instances: Instance[]) => voi
...
+      label: t('Enabled'),
+      labelOff: t('Disabled'),
+      showPinnedLabel: false,
+      isHidden: (instance) => instance.node_type === 'hop',
+      isDisabled: (instance) =>
+        !userAccess ||
+        !isK8s ||
+        (instance?.node_type !== 'execution' && instance?.node_type !== 'hop')
+          ? t(
+              'You do not have permission to edit instances. Please contact your organization administrator if there is an issue with your access.'
```
MilanPospisil, line 64: There must be some difference between old app privileges check and new app checks, because Im logged as the same user in old and new app (my own user using local dev server with all privilleges) and old app allows me to change switch, but new has it disabled because of permission.

frontend/awx/administration/instances/hooks/useInstanceToolbarActions.tsx
```diff
@@ -68,22 +49,58 @@ export function useInstanceToolbarActions(view: IAwxView<Instance>) {
...
+
+export function useRunHealthCheckToolbarAction(view: IAwxView<Instance>, isPinned?: boolean) {
+  const { t } = useTranslation();
+  const runHealthCheck = useRunHealthCheck(view.unselectItemsAndRefresh);
+
+  const cannotRunHealthCheckDueToPending = useCallback(
+    (instance: Instance) => {
+      if (instance.health_check_pending)
+        return t(
+          `Instance has pending health checks. Wait for those to complete before attempting another health check.`
```
MilanPospisil, line 79: This never gets displayed, if the error message is returned, it will only fulfill the condition below and returns "Cannot run health checks on one or more of the selected instances".

frontend/awx/administration/instances/hooks/useInstanceToolbarActions.tsx
```diff
@@ -68,22 +49,58 @@ export function useInstanceToolbarActions(view: IAwxView<Instance>) {
...
+        );
+      return '';
+    },
+    [t]
+  );
+
+  const cannotRunHealthCheckDueToNodeType = useCallback(
+    (instance: Instance) => {
+      if (instance.node_type !== 'execution')
+        return t(`Health checks can only be run on execution instances.`);
```
MilanPospisil, line 89: Same as above.

frontend/awx/administration/instances/hooks/useInstancesColumns.tsx
prat98: Yeah I've been struggling to find out how to add the column to the expanded section I thought `ColumnTableOption.expanded` would work. Might be a framework bug?

prat98: Was thinking in the interest of time just to fix this during hardening or later in another PR. I can go ahead and make a Jira issue for it

frontend/awx/administration/instances/hooks/useInstanceRowActions.tsx
MilanPospisil: Im still having those privilleges issues. I have user as:
image

In my new application, I cant change enabled/disable and also forks, while in old, I can:
image

image

prat98: Added a commit to fix this issue should be resolved now

MilanPospisil:
Looks excellent so far! I have probably one last request, that should be handled in separate PR, since this PR is already too big and that was not problem in your code, but in previous code. When the detail is loaded, but fork slider continues loading and then finally renders and page gets updated. Could you please delay the loading whole page until slider has also data loaded, so everything gets loaded at once? Also, the old app seems to load slider much faster, can it be fixed?

There is this code for loading forks:
const fetchInstanceDetails = async () => {
      const instance = await requestGet<Instance>(awxAPI`/instances/${instanceId}/`);
      const instanceGroups = await requestGet<AwxItemsResponse<InstanceGroup>>(
        awxAPI`/instances/${instanceId}/instance_groups/`
      );
      setInstance(instance);
      setInstanceGroups(instanceGroups);
      setInstanceForks(
        computeForks(
          instance.mem_capacity,
          instance.cpu_capacity,
          parseFloat(instance.capacity_adjustment)
        )
      );
    };
    
    Can you please rewrite it so it uses useGet? Another problem is that it has no error handling, when the error occurs, it crashes whole page. Also the requests seems independent, it should not call sequential await, that is the probably cause of the loading delays. I suggest put data loading into separate hook that is called outside of row actions and row actions only accepts final loaded data, because useRowActions returns actions, it cant return information about error, that page needs to display it. And Im not sure if covering hook the try catch is good best practice...
    
  You can do it as separate PR. Thanks!

MilanPospisil:
LGTM :+1, my request can be done as separate PR.

MilanPospisil (approved):
LGTM :+1:

prat98:
> Looks excellent so far! I have probably one last request, that should be handled in separate PR, since this PR is already too big and that was not problem in your code, but in previous code. When the detail is loaded, but fork slider continues loading and then finally renders and page gets updated. Could you please delay the loading whole page until slider has also data loaded, so everything gets loaded at once? Also, the old app seems to load slider much faster, can it be fixed?
> 
> Can you please rewrite it so it uses useGet? (code below) Another problem is that it has no error handling, when the error occurs, it crashes whole page. Also the requests seems independent, it should not call sequential await, that is the probably cause of the loading delays. I suggest put data loading into separate hook that is called outside of row actions and row actions only accepts final loaded data, because useRowActions returns actions, it cant return information about error, that page needs to display it. And Im not sure if covering hook the try catch is good best practice...
> 
> You can do it as separate PR. Thanks!
> 
> There is this code for loading forks:
> 
> ```js
> const fetchInstanceDetails = async () => {
>       const instance = await requestGet<Instance>(awxAPI`/instances/${instanceId}/`);
>       const instanceGroups = await requestGet<AwxItemsResponse<InstanceGroup>>(
>         awxAPI`/instances/${instanceId}/instance_groups/`
>       );
>       setInstance(instance);
>       setInstanceGroups(instanceGroups);
>       setInstanceForks(
>         computeForks(
>           instance.mem_capacity,
>           instance.cpu_capacity,
>           parseFloat(instance.capacity_adjustment)
>         )
>       );
>     };
> ```

Sounds good to me can get this done in a separate PR

PR #2123, opened 2024-04-20
[ AAP-23071] Add boolean to credential details page

nixocio:
Add boolean to credential details page

image

### DIFF
--- a/cypress/e2e/awx/resources/credentials.cy.ts
+++ b/cypress/e2e/awx/resources/credentials.cy.ts
@@ -213,6 +213,28 @@ describe('Credentials', () => {
   });
 
   describe('Credentials: Details View', () => {
+    it('details page should render boolean field', () => {
+      const credentialName = 'E2E Credential ' + randomString(4);
+      cy.navigateTo('awx', 'credentials');
+      cy.clickButton(/^Create credential$/);
+      cy.get('[data-cy="name"]').type(credentialName);
+      cy.selectDropdownOptionByResourceName('credential-type', 'Container Registry');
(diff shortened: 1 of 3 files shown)


cypress/e2e/awx/resources/credentials.cy.ts
```diff
@@ -213,6 +213,28 @@ describe('Credentials', () => {
   });
 
   describe('Credentials: Details View', () => {
+    it('details page should render boolean field', () => {
+      const credentialName = 'E2E Credential ' + randomString(4);
+      cy.navigateTo('awx', 'credentials');
+      cy.clickButton(/^Create credential$/);
+      cy.get('[data-cy="name"]').type(credentialName);
+      cy.selectDropdownOptionByResourceName('credential-type', 'Container Registry');
+      cy.get('[data-cy="host"]').clear().type('https://host.com');
```
RedOrNot, line 222: can't all the navigation to the right place be in a function?
I see the same lines over and over again , lines 217-221 at least are repeating with different args at cy.selectDropdownOptionByResourceName function 
for example: 
lines 45-49, lines 80-84, lines 111-115
 
and so on

also - from a quick look, you can create 2 function that will cover even more code that repeat itself in this file

RedOrNot (changes requested):

cypress/e2e/awx/resources/credentials.cy.ts
RedOrNot: will simplify late

RedOrNot (approved):
LGTM

daphnemaeve (approved):

PR #2125, opened 2024-04-21
fixing the inventory hosts stubs

RedOrNot:


akus062381 (changes requested):
I reviewed the hosts.cy.ts spec file, and it looks very disorganized and difficult to read. I do not feel that the changes in this file are necessary and I still want to see the tests written in this file like they need to be.

RedOrNot:
@akus062381 we can organize the file
there is no need to create new file just because the other one isn't organized, if it's just a matter of adding more comments I can do it now.. 

also add more data to this file - where each and every part is covered

PR #2132, opened 2024-04-22
Fix loading state of active user contexts

jamestalton:
Changed the login around active user loading to handle the way SWR handles isLoading.


tznamena (approved):
Looks good, tested locally on AWX and even though the API calls are done, it no longer refreshes the login page. :+1:

PR #2133, opened 2024-04-22
Update Login Screen to match UXD designs

jamestalton:
Screenshot 2024-04-22 at 2 43 01 PM

Screenshot 2024-04-22 at 2 43 09 PM

Screenshot 2024-04-22 at 2 43 17 PM


frontend/awx/main/AwxLogin.tsx
```diff
@@ -43,6 +44,9 @@ export function AwxLogin(props: { children: React.ReactNode }) {
           refreshActiveAwxUser?.();
           void mutate(() => true);
         }}
+        icon={<ProductIcon style={{ height: 64 }} />}
+        brand={process.env.BRAND}
```
nixocio, line 48: For those env variables do we need have a fallback value?

nixocio (approved):

frontend/awx/main/AwxLogin.tsx
jamestalton: should not

PR #2137, opened 2024-04-22
AAP 8236 Org EE list

vishsanghishetty:
## Description
This PR is for [ticket](https://issues.redhat.com/browse/AAP-8236) 

## Acceptance Criteria
**Execution Environment tab within Organization**

- [x] creates a component ExecutionEnvironmentsList to reuse in the Organization ExecutionEnvironments tab
- [x] adds logic to hide the Organization column within the context of the Org -> EE tab
- [x] updates OPTIONS call


kialam (changes requested):
We should reuse the existing EE list component instead of creating a new component. Also, the toolbarfilter hook will need to be updated to be able to filter by resource id (og).

vidyanambiar (changes requested):
Works great! ❇️ 

A couple of comments regarding some customizations to the EE list when displayed within an Organization
- The existing AWX UI does not provide edit/delete options on the EE within the context of an organization. Can we check with PM/UXD if this is something we want to change in the new UI? 
- The `Organization` column seems redundant in the Org -> EE tab since we're already within the context of an org
- When the Org -> EE tab is opened the network tab indicates OPTIONS requests to `/api/v2/execution_environments/` (see call to the `useOptions` hook in the EEList component) but the OPTIONS request should be made to `/api/v2/organizations/:id/execution_environments`

image

vishsanghishetty:
@vidyanambiar 
The existing AWX UI does not provide edit/delete options on the EE within the context of an organization. Can we check with PM/UXD if this is something we want to change in the new UI?
Tina said, Projects -> Templates has actions and let's follow that.

vishsanghishetty:
@vidyanambiar 
updated the below 2, please review

1. The Organization column seems redundant in the Org -> EE tab since we're already within the context of an org
2. When the Org -> EE tab is opened the network tab indicates OPTIONS requests to /api/v2/execution_environments/ (see call to the useOptions hook in the EEList component) but the OPTIONS request should be made to /api/v2/organizations/:id/execution_environments

prat98 (approved):
Went through the PR and the changes requested. Don't see any other issues and I believe the requested changes have been addressed well. Looks good to me! 🚀

vidyanambiar (approved):

frontend/awx/access/organizations/hooks/useOrganizationsFilters.tsx
```diff
@@ -0,0 +1,18 @@
+import {
+  useCreatedByToolbarFilter,
+  useModifiedByToolbarFilter,
+} from '../../../common/awx-toolbar-filters';
+import { useDynamicToolbarFilters } from '../../../common/useDynamicFilters';
+
+export function useOrganizationsFilters() {
```
kialam, line 7: Remove because we already have a `useOrganizationsFilters` hook here: https://github.com/ansible/ansible-ui/blob/AAP-8236-Org-EE-List/frontend/awx/access/organizations/Organizations.tsx#L190

frontend/awx/administration/execution-environments/hooks/useExecutionEnvironmentsFilters.tsx
```diff
@@ -1,17 +1,24 @@
...
+  useModifiedByToolbarFilter,
+} from '../../../common/awx-toolbar-filters';
 
 export function useExecutionEnvironmentsFilters() {
+  const createdByToolbarFilter = useCreatedByToolbarFilter();
+  const modifiedByToolbarFilter = useModifiedByToolbarFilter();
   const toolbarFilters = useDynamicToolbarFilters({
     optionsPath: 'execution_environments',
-    preSortedKeys: ['name', 'id'],
+    preSortedKeys: ['name', 'image'],
```
kialam, line 12: We should add `id` back.

frontend/awx/administration/execution-environments/hooks/useExecutionEnvironmentsFilters.tsx
```diff
@@ -1,17 +1,24 @@
...
     preFilledValueKeys: {
-      name: {
+      image: {
         apiPath: 'execution_environments',
       },
-      id: {
+      name: {
```
kialam, line 17: We should add `id` back.

frontend/awx/administration/execution-environments/hooks/useExecutionEnvironmentsFilters.tsx
kialam: I would suggest ordering the search dropdown like so : name, id, image, created-by, modified-by

frontend/awx/administration/execution-environments/hooks/useExecutionEnvironmentsFilters.tsx
```diff
@@ -1,17 +1,24 @@
 import { useDynamicToolbarFilters } from '../../../common/useDynamicFilters';
+import {
+  useCreatedByToolbarFilter,
+  useModifiedByToolbarFilter,
+} from '../../../common/awx-toolbar-filters';
 
 export function useExecutionEnvironmentsFilters() {
```
kialam, line 7: This should take an additional prop, `url` so it can filter down the list of filterable values when it's within a tabbed view. Currently the UI shows all EEs within an Org:

filtered_ee

I would recommend using a similar approach to this [PR](https://github.com/ansible/ansible-ui/pull/1960/files). Let me know if you have any questions - happy to discuss further.

kialam (changes requested):
We should remove useOrganizationsFilters since we already have that hook. We should update useExecutionEnvironmentFilters to accept a prop to pass to the useDynamicFilters component to filter down available values when it's within a tabbed view. Nice work!

vishsanghishetty:
@kialam made the changes requested and the filters display the EEs within the Org

vishsanghishetty:
image
image

kialam (approved):

PR #2140, opened 2024-04-22
Ensure schedule summary and schedule rule preview times correspond

AlexSCorey:
This addresses AAP-23130.  Prior to this change it was possible the Rule preview date times that are different from the schedule summary.  This pr addresses that bug.
In the screenshot below the schedule summary shows a UTC time of 7:30.  The timezone on the schedule is `America/New_York` so when local toggled it should show a 3:30 pm time.

The labels in Rule 1 show the same start time, its just that it is in local time only and cannot be toggled.
Screenshot 2024-04-22 at 4 25 53 PM
Screenshot 2024-04-22 at 4 26 57 PM


frontend/awx/views/schedules/components/RulesPreview.tsx
```diff
@@ -10,80 +10,71 @@ import {
...
+            <DescriptionList data-cy={`rule-${index + 1}`} isCompact>
               <DescriptionListGroup>
-                <DescriptionListTerm>{t('Rule preview')}</DescriptionListTerm>
+                <DescriptionListTerm>{t('Rule preview local')}</DescriptionListTerm>
                 <LabelGroup numLabels={5}>
                   {rule
                     .all((_rule, i) => i < 10)
                     ?.map((value, i) => {
-                      return <Label key={i}>{formatDateString(value)}</Label>;
+                      const localDateTime = DateTime.fromISO(value.toISOString())
```
acruzgon, line 43: i think setting `setZone` to true is missing because times don't match in my machine
` DateTime.fromISO(value.toISOString(), { setZone: true })`

acruzgon (changes requested):

acruzgon (approved):
lgtm!

PR #2155, opened 2024-04-24
adding inventory-host-groups test and more comments

RedOrNot:


RedOrNot:
https://issues.redhat.com/browse/AAP-20198

cypress/e2e/awx/resources/hosts.cy.ts
```diff
@@ -229,6 +193,72 @@ describe('host and inventory host', () => {
...
+        cy.get(`[data-cy="row-id-${group.id}"] [data-cy="checkbox-column-cell"]`).click();
+        cy.clickModalButton('Confirm');
+        cy.contains('button', 'Close').click();
+        cy.contains(group.name);
+        deleteHost(inventory.id, inventory_host, host.name);
+      });
+    });
+  }
+
   function createAndEditAndDeleteHost(inventory_host: boolean, inventory: Inventory) {
```
MilanPospisil, line 262: I will suggest maybe different parameter, such as type: 'standalone_host' | 'inventory_host', because this way,  the call of the function is more readable, such as:

createAndEditAndDeleteHost('standalone_host')

this way it is much more clear what is that parameter call, true/false is not clear.

cypress/e2e/awx/resources/hosts.cy.ts
```diff
@@ -29,79 +29,33 @@ describe('host and inventory host', () => {
   });
 
   //tests
-  it.skip('can create, edit and delete a inventory host', () => {
+  it('can create, edit and delete a inventory host', () => {
```
MilanPospisil, line 32: Is this scaffolding from Sarah? I though the inventory hosts tests will be in inventory file?

RedOrNot: no
as you can see in line 38 it's not only inventory hosts

cypress/e2e/awx/resources/hosts.cy.ts
RedOrNot: do you mean pass a string?
as for now I've added a lot of comments to try and make it clearer 
I can add for this as well

RedOrNot: changed it base on suggestion - now it's string

MilanPospisil (approved):
LGTM :+1:

PR #2168, opened 2024-04-24
[AAP-22271] Add content to the user tokens tab

tznamena:
This PR adds code to populate user tokens table. Token details page will be implemented in another PR.

### DIFF
--- a/cypress/fixtures/userToken-app.json
+++ b/cypress/fixtures/userToken-app.json
@@ -1,0 +1,31 @@
+{
+  "id": 30,
+  "type": "o_auth2_access_token",
+  "url": "/api/v2/tokens/30/",
+  "related": {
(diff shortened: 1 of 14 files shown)


vidyanambiar (approved):

vidyanambiar (changes requested):
We don't need the created and modified columns to show up in the list view. They can be hidden columns for use in the details page.  
If we import `createdColumn` and `modifiedColumn` from `frontend/common/columns.tsx`, the correct settings will be applied.

vidyanambiar (approved):
Looks great! 🎉 

2 minor comments:
- Can we change `Back to User Tokens list` to `Back to User Tokens`. That's been the pattern elsewhere.
- Could you add a component test please to verify the details page.

PR #2175, opened 2024-04-25
Notifications actions

MilanPospisil:
Adds test action for list and detail.

Test  action calls API, that starts notification test and returns notification id (one notification template can have multiple notifications). The page then watches the running particular notification id and then it alerts the result (for detail), for list, the alert is not shown since there can be multiple notifications templates running.

### DIFF
--- a/frontend/awx/administration/notifiers/NotificationPage/NotificationPage.tsx
+++ b/frontend/awx/administration/notifiers/NotificationPage/NotificationPage.tsx
@@ -20,8 +20,7 @@ export function NotificationPage() {
   const { t } = useTranslation();
   const params = useParams<{ id: string }>();
 
-  const { runningNotifications, onNotifierStartTest, checkNotifiers } =
-    useNotificationsWatch('detail');
+  const { runningNotifications, onNotifierStartTest, checkNotifiers } = useNotificationsWatch();
 
   // set refresh interval to be faster when test is running
   const {
@@ -44,11 +43,11 @@ export function NotificationPage() {
     runningNotifications,
   });
 
-  useEffect( () => {
-    if (notificationTemplate)
-    {
-      checkNotifiers(notificationTemplate);
+  useEffect(() => {
+    if (notificationTemplate) {
+      checkNotifiers([notificationTemplate]);
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [notificationTemplate]);
 
   if (error) return <AwxError error={error} handleRefresh={refresh} />;
--- a/frontend/awx/administration/notifiers/NotificationPage/NotificationPage.tsx
+++ b/frontend/awx/administration/notifiers/NotificationPage/NotificationPage.tsx
@@ -16,13 +16,14 @@ import { usePageNavigate } from '../../../../../framework';
 import { useState } from 'react';
 import { usePageAlertToaster } from '../../../../../framework';
 import { StatusLabel } from '../../../../common/Status';
-import { RunningNotificationsType } from '../hooks/useNotifiersRowActions';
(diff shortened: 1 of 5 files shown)


frontend/awx/administration/notifiers/NotificationPage/NotificationPage.tsx
```diff
@@ -13,66 +13,46 @@ import { useNotifiersRowActions } from '../hooks/useNotifiersRowActions';
...
-        variant: 'info',
-        title: t('Notifier test result'),
-        children: <StatusLabel status={notification.status} />,
-      });
-      setTestedNotificationId(undefined);
-    }
-  }
+  const notificationId = runningNotifications[notificationTemplate.id.toString()];
+
+  checkNotifiers(notificationTemplate);
```
prat98, line 52: Should this be wrapped in a `useEffect()`?

frontend/awx/administration/notifiers/Notifiers.tsx
```diff
@@ -37,6 +49,12 @@ export function Notifiers() {
     notificationsOptions && notificationsOptions.actions && notificationsOptions.actions['POST']
   );
 
+  if (view.pageItems) {
+    view.pageItems.forEach((item) => {
+      checkNotifiers(item);
+    });
+  }
```
prat98, line 56: Same here should this also be wrapped in a `useEffect()`?

frontend/awx/administration/notifiers/NotificationPage/NotificationPage.tsx
MilanPospisil: Yeah, it may be in useEffect. Now it recomputes every render, useEffect may recompute it after pageItems change, so it may save some computational effor.

frontend/awx/administration/notifiers/hooks/useNotificationsWatch.tsx
```diff
@@ -0,0 +1,44 @@
...
+    checkNotifiers: (notificationTemplate: NotificationTemplate) => {
+      // search for notification id
+      const notificationId = runningNotifications[notificationTemplate.id.toString()];
+      const notifications = notificationTemplate.summary_fields?.recent_notifications;
+      const notification = notifications.find(
+        (notification) => notification.id.toString() === notificationId
+      );
+
+      if (notification && notification.status !== 'pending') {
+        if (type === 'detail') {
```
prat98, line 31: Did you mean to show toast for list view like this?
```suggestion
        if (type === 'list') {
```
In the old UI the toast is shown on the list view and not on the details view

frontend/awx/administration/notifiers/hooks/useNotifiersRowActions.tsx
```diff
@@ -56,7 +58,7 @@ export function useNotifiersRowActions(
         icon: RocketIcon,
```
prat98, line 58: Not a part of this PR but I noticed the old UI uses the `BellIcon` instead of the `RocketIcon` here

frontend/awx/administration/notifiers/hooks/useNotificationsWatch.tsx
prat98: Also saw the PR description, but feel like it's possible to have multiple notifications pop up for different notifications. Here's what I got when running two notifs after implementing the change suggested above
Screenshot 2024-04-26 at 11 49 00 AM

prat98: Showing it in both is fine too in my opinion

prat98 (approved):
Looks good to me! 🚀

PR #2174, opened 2024-04-25
Adds Instance Group Add and Edit Form along with InstanceGroupPage

AlexSCorey:
This addresses AAP-8262, AAP-8303, and part of AAP-7840.  7840 is an issue for the instance group details view.  This pr brings in just the InstanceGroupPage.tsx.  Primarily I did this so that I could assert on the page title after editing or creating and IG in the e2e tests.

PR #2183, opened 2024-04-25
Associate/Disassociate Instances from Instance Group

prat98:
This PR is for AAP-23129 which is about associating and disassociating an instance from an instance group. This PR adds... 

An associate button to the instance group instances list toolbar

- This button is enabled if the user has POST permissions to the instances endpoint.
- The associate button opens up a `MulitSelectDialog` that allows the user to select instances to associate.
- Upon selecting the instances to associate a `BulkActionDialog` is presented to confirm the users selection.

A disassociate button to instance group instances list toolbar

- This button is disabled based on the logic in a new function written [here](https://github.com/ansible/ansible-ui/pull/2183/files#diff-5980aa096a77feb4858d75eb1cd395ad2c6afec9f14cd2c380f60679c8443137R80)
- The disassociate button opens up a `BulkConfirmationDialog` here which allows to user to confirm their selection of instances to disassociate

List support for the useAwxView `queryParams` param

- This update allows passing a list of string values for multiple queries of the same key with a different value. Ex: 
- `const queryParam = {
    not__node_type: ['control', 'hybrid']
  }`

### DIFF
--- a/frontend/awx/administration/instance-groups/InstanceGroupPage/InstanceGroupPage.tsx
+++ b/frontend/awx/administration/instance-groups/InstanceGroupPage/InstanceGroupPage.tsx
@@ -59,11 +59,18 @@ export function InstanceGroupPage() {
           page: AwxRoute.InstanceGroups,
           persistentFilterKey: 'instance_groups',
         }}
-        tabs={[
-          { label: t('Details'), page: AwxRoute.InstanceGroupDetails },
-          { label: t('Instances'), page: AwxRoute.InstanceGroupInstances },
-          { label: t('Jobs'), page: AwxRoute.InstanceGroupJobs },
-        ]}
+        tabs={
+          instanceGroup?.is_container_group
+            ? [
+                { label: t('Details'), page: AwxRoute.InstanceGroupDetails },
+                { label: t('Jobs'), page: AwxRoute.InstanceGroupJobs },
+              ]
+            : [
+                { label: t('Details'), page: AwxRoute.InstanceGroupDetails },
+                { label: t('Instances'), page: AwxRoute.InstanceGroupInstances },
+                { label: t('Jobs'), page: AwxRoute.InstanceGroupJobs },
+              ]
+        }
         params={{ id: instanceGroup.id }}
       />
     </PageLayout>
--- a/frontend/awx/administration/instance-groups/InstanceGroupPage/hooks/useAssociateInstanceModal.tsx
+++ b/frontend/awx/administration/instance-groups/InstanceGroupPage/hooks/useAssociateInstanceModal.tsx
@@ -1,0 +1,52 @@
+import { useEffect, useState } from 'react';
+import { MultiSelectDialog, usePageDialogs } from '../../../../../../framework';
+import { useAwxView } from '../../../../common/useAwxView';
+import { Instance } from '../../../../interfaces/Instance';
+import { useInstancesColumns } from '../../../instances/hooks/useInstancesColumns';
+import { useInstancesFilters } from '../../../instances/hooks/useInstancesFilter';
+import { awxAPI } from '../../../../common/api/awx-utils';
+import { useTranslation } from 'react-i18next';
+
+export interface AssociateInstanceModalProps {
+  instanceGroupId: string;
+  onAssociate: (items: Instance[]) => void;
+}
+
+function AssociateInstanceModal(props: AssociateInstanceModalProps) {
+  const { t } = useTranslation();
+  const tableColumns = useInstancesColumns();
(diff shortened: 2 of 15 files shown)


frontend/awx/administration/instance-groups/InstanceGroupPage/InstanceGroupPage.tsx
```diff
@@ -59,11 +59,18 @@ export function InstanceGroupPage() {
           page: AwxRoute.InstanceGroups,
           persistentFilterKey: 'instance_groups',
         }}
-        tabs={[
-          { label: t('Details'), page: AwxRoute.InstanceGroupDetails },
-          { label: t('Instances'), page: AwxRoute.InstanceGroupInstances },
-          { label: t('Jobs'), page: AwxRoute.InstanceGroupJobs },
-        ]}
+        tabs={
```
AlexSCorey, line 62: I have a pr open that is also working in this file.  just fyi https://github.com/ansible/ansible-ui/pull/2174. 
and https://github.com/ansible/ansible-ui/pull/2189
 I've separated out container group and instance groups since their urls are different.

frontend/awx/administration/instance-groups/InstanceGroupPage/hooks/useAssociateInstanceModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+function AssociateInstanceModal(props: AssociateInstanceModalProps) {
+  const { t } = useTranslation();
+  const tableColumns = useInstancesColumns();
+  const toolbarFilters = useInstancesFilters();
+  const { instanceGroupId, onAssociate } = props;
+
+  const view = useAwxView<Instance>({
+    url: awxAPI`/instances/`,
+    queryParams: { not__node_type: ['control', 'hop'], not__rampart_groups__id: instanceGroupId },
```
AlexSCorey, line 23:
```suggestion
    queryParams: { not__node_type: 'control,hop', not__rampart_groups__id: instanceGroupId },
```
I think this change will work and it will prevent you from having to change the queryparams for hub and inside the `useAwxView`.

frontend/awx/administration/instance-groups/InstanceGroupPage/hooks/useAssociateInstanceToIG.tsx
```diff
@@ -0,0 +1,38 @@
...
+      actionFn: async (instance: Instance, signal: AbortSignal) => {
+        await postRequest(
+          awxAPI`/instance_groups/${instanceGroupId}/instances/`,
+          {
+            id: instance.id,
+          },
+          signal
+        );
+      },
+      processingText: t('Adding host to group...', { count: instances.length }),
```
AlexSCorey, line 34: Do we have a covention of use the elipse in this use case? `...`.  If not, I would advise against it here.

AlexSCorey (commented):
Lets add some tests here to ensure that associate and disassociate work.

frontend/awx/administration/instance-groups/InstanceGroupPage/hooks/useAssociateInstanceModal.tsx
prat98: I tried making the query with the comma and it wasn't working. And then I decided to add list support. The query with the comma doesn't filter out control or hop types.

frontend/awx/administration/instance-groups/InstanceGroupPage/hooks/useIGInstanceToolbarActions.tsx
```diff
@@ -1,13 +1,102 @@
...
 
 export function useIGInstanceToolbarActions(view: IAwxView<Instance>) {
   const healthCheckAction = useRunHealthCheckToolbarAction(view, true);
+  const associateAction = useIGInstanceAssociateToolbarAction(view);
+  const disassociateAction = useIGInstanceDisassociateToolbarAction(view);
 
   return useMemo<IPageAction<Instance>[]>(() => {
-    return [healthCheckAction];
-  }, [healthCheckAction]);
+    return [associateAction, disassociateAction, healthCheckAction];
```
AlexSCorey, line 24: There are a couple of places in this file where we have phrases like `id ?? "`.  If we are associating or disassociating and instance to an instance group id is a required field and we should not need to use the nullish coalescing operator.  My recommendation is that if id is undefined we return out of the function asap. Let me know if I can help with this.

AlexSCorey (commented):
We should add some e2e tests for associating/disassociating an Instance.

AlexSCorey (approved):
I've pulled this down and tested it out manually.  It works as expected.  Consider adding some e2e tests.

PR #2187, opened 2024-04-26
Fix inventory groups tests

MilanPospisil:
Credentials selection in the run command modal can flick sometimes,

The flicking causes failures, because cypress tries to get element, which dissapeares and appear again, but the new element is not selected, the old one dissapeared is and it fails.

image

Another problem was filter for inventory in run command test. Inventories were displayed slowly, so I added intercept and wait to wait for the filtered result.

### DIFF
--- a/cypress/e2e/awx/resources/inventoryGroup.cy.ts
+++ b/cypress/e2e/awx/resources/inventoryGroup.cy.ts
@@ -4,7 +4,7 @@ import { Organization } from '../../../../frontend/awx/interfaces/Organization';
 import { Credential } from '../../../../frontend/awx/interfaces/Credential';
 import { ExecutionEnvironment } from '../../../../frontend/awx/interfaces/ExecutionEnvironment';
 
-describe.skip('Inventory Groups', () => {
+describe('Inventory Groups', () => {
   let organization: Organization;
   let inventory: Inventory;
   let machineCredential: Credential;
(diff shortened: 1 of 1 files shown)


cypress/e2e/awx/resources/inventoryGroup.cy.ts
```diff
@@ -93,6 +93,7 @@ describe.skip('Inventory Groups', () => {
         //1) Use the inventory created in beforeEach block, access the groups tab of that inventory
         cy.navigateTo('awx', 'inventories');
         cy.filterTableBySingleSelect('name', inventory.name);
+        cy.wait(2000);
```
fullsushidev, line 96: Is there a way for us to intercept some of the API calls and wait for them instead of waiting X amount of time?

fullsushidev (changes requested):

cypress/e2e/awx/resources/inventoryGroup.cy.ts
```diff
@@ -135,6 +141,11 @@ describe.skip('Inventory Groups', () => {
             });
           }
         });
+
+        // this is for sync only, those modals are badly implemented and can occasionaly flick
+        // this is ugly fix, until app is repaired, this is necessity
+        cy.wait(2000);
```
fullsushidev, line 147: Just a reminder to include this on a jira task to be fixed later :)

fullsushidev (approved):
LGTM.

I just added a comment as a reminder to include a task for a future update of the component, and then we can update the `cy.wait` following Cypress best practices.

PR #2189, opened 2024-04-26
Adds Container Group add and edit along with container group page

AlexSCorey:
This addresses AAP-8264, AAP-8302, and part of AAP-8289.

It brings in Add, Edit for container groups and the Container group page component with all its functionality.  Edit/delete.

### DIFF
--- a/cypress/e2e/awx/administration/instanceGroups.cy.ts
+++ b/cypress/e2e/awx/administration/instanceGroups.cy.ts
@@ -1,6 +1,7 @@
 import { randomString } from '../../../../framework/utils/random-string';
 import { InstanceGroup } from '../../../../frontend/awx/interfaces/InstanceGroup';
-
+import { Organization } from '../../../../frontend/awx/interfaces/Organization';
+import { Credential } from '../../../../frontend/awx/interfaces/Credential';
 describe('Instance Groups', () => {
   let instanceGroup: InstanceGroup;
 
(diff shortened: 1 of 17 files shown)


cypress/e2e/awx/administration/instanceGroups.cy.ts
```diff
@@ -25,6 +27,133 @@ describe('Instance Groups', () => {
...
+          cy.get('#credential-select').type(credential.name);
+          cy.get('[data-cy="name"]').type(containerGroupName);
+          cy.getByDataCy('override').click({ force: true });
+          cy.get('[data-cy="max-concurrent-jobs"]').clear();
+          cy.get('[data-cy="max-concurrent-jobs"]').type('3');
+
+          cy.get('[data-cy="max-forks"]').clear();
+          cy.get('[data-cy="max-forks"]').type('4');
+          cy.clickButton(/^Create Container Group$/);
+          cy.verifyPageTitle(containerGroupName);
```
nixocio, line 58: We should delete the resources created by the test.

frontend/awx/administration/instance-groups/ContainerGroupForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+import { usePostRequest } from '../../../common/crud/usePostRequest';
+import { useGetItem } from '../../../common/crud/useGet';
+import { useParams } from 'react-router-dom';
+import { AwxError } from '../../common/AwxError';
+import { usePatchRequest } from '../../../common/crud/usePatchRequest';
+import { PageFormCredentialSelect } from '../../access/credentials/components/PageFormCredentialSelect';
+import { useFormContext, useWatch } from 'react-hook-form';
+import { PageFormSection } from '../../../../framework/PageForm/Utils/PageFormSection';
+import { SummaryFieldCredential } from '../../interfaces/summary-fields/summary-fields';
+import { POD_SPEC_DEFAULT_VALUE } from './constants';
```
nixocio, line 26: I believe this value comes from an OPTION request from the API. If we have this value provided by the API better to get from there in case the API is updated.

cypress/e2e/awx/administration/instanceGroups.cy.ts
AlexSCorey: As far as deleting, there is a delete function in the afterEach block.

PR #2192, opened 2024-04-26
Create secret management input component

daphnemaeve:
Adds button to open secret management wizard for credential types without kind of 'external'. 

Before:
Screenshot 2024-04-26 at 11 56 50 AM

After:
Screenshot 2024-04-26 at 11 56 28 AM

### DIFF
--- a/frontend/awx/access/credentials/CredentialForm.tsx
+++ b/frontend/awx/access/credentials/CredentialForm.tsx
@@ -381,7 +381,7 @@ function PageFormSecretInput({
           onClear={handleHideField}
           shouldHideField={shouldHideField}
           label={field.label}
-          placeholder={t('$ENCRYPTED')}
+          placeholder={t('ENCRYPTED')}
         >
           <InputGroup>
             <Tooltip content={t(`Reset`)}>
@@ -427,7 +427,7 @@ function PageFormSecretInput({
           }}
           shouldHideField={shouldHideField}
           label={field.label}
-          placeholder={t('$ENCRYPTED')}
+          placeholder={t('ENCRYPTED')}
         >
           <InputGroup>
             <InputGroupItem>
--- a/frontend/awx/access/credentials/CredentialForm.tsx
+++ b/frontend/awx/access/credentials/CredentialForm.tsx
@@ -368,11 +368,10 @@ function PageFormSecretInput({
   const [shouldHideField, setShouldHideField] = useState(field.secret && isEditMode);
   const { resetField } = useForm();
   const [clear, setClear] = useState(false);
-  const { setValue, clearErrors } = useFormContext();
+  const { setValue } = useFormContext();
 
   const handleHideField = () => {
     setShouldHideField(!shouldHideField);
-    console.log(field.id);
     resetField(field.id);
     setClear(!clear);
     //hide field
@@ -400,7 +399,7 @@ function PageFormSecretInput({
       <PageFormSecret
         //key={field.id}
         onClear={() => {
-          console.log(field.id);
+          //console.log(field.id);
           setValue(field.id, '');
           //resetField('security-token');
           setShouldHideField(!shouldHideField);
--- a/frontend/awx/access/credentials/CredentialForm.tsx
+++ b/frontend/awx/access/credentials/CredentialForm.tsx
@@ -1,4 +1,4 @@
-import { ReactNode, useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { useFormContext, useWatch } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
 import { useNavigate, useParams } from 'react-router-dom';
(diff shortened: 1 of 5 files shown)


frontend/awx/access/credentials/CredentialForm.tsx
```diff
@@ -295,38 +404,18 @@ function CredentialSubForm({ credentialType }: { credentialType: CredentialType
...
-          if (field?.multiline) {
-            return (
-              <CredentialMultilineInput
-                kind={credentialType.kind}
-                key={field.id}
-                field={field}
-                requiredFields={requiredFields}
-              />
-            );
-          } else if (credentialType.kind === 'ssh' && field.id === 'become_method') {
```
nixocio, line 312: This should not be removed...it is an edge case and we should keep this case.

frontend/awx/access/credentials/CredentialForm.tsx
```diff
@@ -269,14 +268,124 @@ function CredentialInputs({
...
+  requiredFields,
+}: {
+  field: CredentialInputField;
+  isEditMode: boolean;
+  credentialType: CredentialType;
+  requiredFields: string[];
+}) {
+  const [t] = useTranslation();
+  const [shouldHideField, setShouldHideField] = useState(field.secret && isEditMode);
+  const { resetField } = useForm();
```
nixocio, line 293:
```suggestion
 
```

frontend/awx/access/credentials/CredentialForm.tsx
```diff
@@ -269,14 +268,124 @@ function CredentialInputs({
...
+  };
+
+  if (field.multiline) {
+    if (field.secret && isEditMode) {
+      return (
+        <SecretManagementInputField
+          onClear={handleHideField}
+          shouldHideField={shouldHideField}
+          label={field.label}
+          placeholder={t('ENCRYPTED')}
```
nixocio, line 310:
```suggestion
          placeholder={t('$ENCRYPTED')}
```

frontend/awx/access/credentials/CredentialForm.tsx
```diff
@@ -269,14 +268,124 @@ function CredentialInputs({
...
+    if (field.secret && isEditMode) {
+      return (
+        <SecretManagementInputField
+          onClear={() => {
+            setValue(field.id, '');
+            setShouldHideField(!shouldHideField);
+          }}
+          shouldHideField={shouldHideField}
+          label={field.label}
+          placeholder={t('ENCRYPTED')}
```
nixocio, line 348:
```suggestion
          placeholder={t('$ENCRYPTED')}
```

frontend/awx/access/credentials/components/SecretManagementInputField.tsx
```diff
@@ -0,0 +1,89 @@
...
+          <Tooltip content={t(`Replace`)}>
+            <Button variant="control" onClick={() => onClear()}>
+              <Icon>
+                <RedoIcon />
+              </Icon>
+            </Button>
+          </Tooltip>
+          <TextInput
+            aria-label={t('hidden value')}
+            placeholder={placeholder ? placeholder : 'ENCRYPTED'}
```
nixocio, line 78:
```suggestion
            placeholder={placeholder ? placeholder : t($'ENCRYPTED')}
```

frontend/awx/access/credentials/CredentialForm.tsx
daphnemaeve: Agreed, I'll put that back in. Does this case also have the potential to be an encrypted field?

frontend/awx/access/credentials/CredentialForm.tsx
```diff
@@ -269,14 +268,124 @@ function CredentialInputs({
...
+  field: CredentialInputField;
+  isEditMode: boolean;
+  credentialType: CredentialType;
+  requiredFields: string[];
+}) {
+  const [t] = useTranslation();
+  const [shouldHideField, setShouldHideField] = useState(field.secret && isEditMode);
+  const { resetField } = useForm();
+  const [clear, setClear] = useState(false);
+  const { setValue } = useFormContext();
```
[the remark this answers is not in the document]
daphnemaeve: I don't believe declaring resetField is necessary because if we remove ```const { resetField } = useForm();``` as suggested, then it won't be used in this function and causes an error.

frontend/awx/access/credentials/CredentialForm.tsx
nixocio: @daphnemaeve it will have to manual test that one.

kialam:
Encrypted fields should only show up for fields that have values in them. See old UI for reference:
Screenshot 2024-05-15 at 10 20 05 AM
New UI:
Screenshot 2024-05-15 at 10 20 02 AM
Fields like Private Key Passphrase, etc. should not show up as encrypted.

kialam:
A few more things:

1. For the multiline input, the key icon is disabled in the edit form, and the undo/revert button should go to the left of the field:

Screenshot 2024-05-15 at 10 20 02 AM

2. Clicking revert and then undo to put back the previous value sends an empty string to the API instead of the previously stored value.
3. We should rename `SecretManagementInputField` since these fields are not tied to the Credential Plugins work.
4. I see some commits that were also in a previous PR that was merged: https://github.com/ansible/ansible-ui/pull/2228/commits. To avoid any potential conflicts or unexpected behavior I would either rebase against main and if the commits still show up I would delete them. I think squashing a few of the other commit messages would also clean this branch up and mark relevant work more clearly.

PR #2193, opened 2024-04-26
AWX User -> Roles 

vidyanambiar:
[AAP-20559](https://issues.redhat.com/browse/AAP-20559)

- Roles tab for AWX user with ability to remove roles and add roles via the roles wizard. 
- This PR moves common Access components (Access, TeamAccess, UserAccess, ResourceAccess) out of the `eda` directory into `frontend/common` so that it can be used for building AWX UIs as well.

image


lgalis (approved):
Looks good!

PR #2194, opened 2024-04-26
CredentialType - update Generate extra vars

lgalis:
Update the dataEditorValue on a value change - so the values set via setValue will show.
Screenshot from 2024-04-26 10-27-41
Screenshot from 2024-04-29 10-25-31


mkanoor:
https://issues.redhat.com/browse/AAP-22567

appuk:
@lgalis Can you please also add the validation for non empty injector configuration? Thanks!

PR #2195, opened 2024-04-27
[AAP-22424] Add 'workers offline' status

lgalis:
Screenshot from 2024-04-26 21-06-11


frontend/common/Status.tsx
```diff
@@ -89,6 +89,7 @@ function useLabel(status: string | undefined, t: (str: string) => string) {
     unknown: t('Unknown'),
     unreachable: t('Unreachable'),
     waiting: t('Waiting'),
+    'workers offline': t('Workers offline'),
```
nixocio, line 92:
```suggestion
    'workers-offline': t('Workers offline'),
```

Just to be consistent with the remaining work.

nixocio (approved):

PR #2199, opened 2024-04-29
Update EDA roles list to support adding, editing and deleting custom roles

mabashian:
This PR updates the EDA roles list, details and forms to interact with the new role_definitions endpoint. EDA will support both built-in and custom roles. Built-in roles cannot be modified or deleted. Custom roles can be created, modified and deleted given the appropriate permissions.

Each role has a single content_type. Based on that content_type, certain permissions can be assigned. The mapping for that does not currently exist in the API and is being managed on the client through the useEdaRoleMetadata hook.

There are two known issues at this point in time that (imo) shouldn't block a merge. The first is the alignment of the expanded row view:

320108744-2798e859-5ac9-466f-9816-1f105b06437e

The second is that some permissions are actually dependent on other permissions being present. A good example of this are the Change permissions. In order to apply a Change permission you also need to apply a View permission. The API will catch this:

320108744-2798e859-5ac9-466f-9816-1f105b06437e

but the UI doesn't really make this apparent to the user up front. I think we should try to improve on this UX moving forward but the current implementation is MVP.


vidyanambiar (approved):
lgtm! 👍 

As discussed the roles metadata is a bit out of date and will need to be updated in a follow-on PR when we have the updated metadata from the server.

PR #2200, opened 2024-04-29
Fixes breadcrumb routing for schedules at the root breadcrumb

AlexSCorey:
Addresses https://issues.redhat.com/browse/AAP-23524


akus062381 (approved):
looks good!

PR #2202, opened 2024-04-29
Adds missing fields to job details page

AlexSCorey:
This addresses https://issues.redhat.com/browse/AAP-7820

I use the old ui to compare with the fields in the UI to try to determine what was missing.  I looked at the following job types in only fairly basic configurations:

- Project sync
- command
- playbook
- workflow job

### DIFF
--- a/frontend/awx/views/jobs/JobDetails.tsx
+++ b/frontend/awx/views/jobs/JobDetails.tsx
@@ -31,52 +31,64 @@ export function JobDetails() {
   return (
     <PageDetails>
       <PageDetailsFromColumns columns={columns} item={job} />
-      {job.playbook && <PageDetail label={t('Playbook')}>{job.playbook}</PageDetail>}
-      {job.summary_fields.project_update?.status && (
-        <PageDetail label={t('Project update status')}>
-          <StatusCell
-            status={job.summary_fields.project_update?.status}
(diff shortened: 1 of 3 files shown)


frontend/awx/views/jobs/JobDetails.tsx
```diff
@@ -1,25 +1,119 @@
...
+  const getPageUrl = useGetPageUrl();
+  const pageNavigate = usePageNavigate();
   if (params.job_id && !job)
     return <EmptyStateNoData title="" description={t('Workflow job deleted')} />;
   if (!job) return <Skeleton />;
-
   return (
     <PageDetails>
       <PageDetailsFromColumns columns={columns} item={job} />
+      {job.playbook && <PageDetail label={t('Playbook')}>{job.playbook}</PageDetail>}
```
keithjgrant, line 34: PageDetail has an `isEmpty` prop you can use for all these conditionals. This approach should de-clutter the JSX quite a bit:

```suggestion
      <PageDetail label={t('Playbook')} isEmpty={!job.playbook}>{job.playbook}</PageDetail>
```

keithjgrant (approved):
Looks good. Just one suggestion to clean up the conditionals in JSX a bit

PR #2207, opened 2024-04-30
[AAP-22819] Update the api used to check in-use credentials

lgalis:
Screenshot from 2024-05-03 17-23-45


frontend/eda/common/EdaResourcesCommon.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const inUseResources: string[] = [];
+  await Promise.all(
+    resources.map(async (item) => {
+      const inUseRes = await ResourceInUse(item, query);
+      if (inUseRes?.count > 0) {
+        inUseResources.push(item.name);
+      }
+    })
+  );
+  return inUseResources;
```
AlexSCorey, line 24: I think you can return the Promise.all directly. 
```suggestion
return Promise.all(
    resources.map(async (item) => {
      const inUseRes = await ResourceInUse(item, query);
      if (inUseRes?.count > 0) {
        inUseResources.push(item.name);
      }
    })
  );
```

frontend/eda/common/EdaResourcesCommon.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const inUseResources: string[] = [];
+  await Promise.all(
+    resources.map(async (item) => {
+      const inUseRes = await ResourceRefInUse(item, query);
+      if (inUseRes?.references && inUseRes?.references.length > 0) {
+        inUseResources.push(item.name);
+      }
+    })
+  );
+  return inUseResources;
```
AlexSCorey, line 46: Same suggestion here.

AlexSCorey (commented):
This looks good.  I did not pull it down to test it.  I did make a few suggestions, but they truly are just suggestions.  Let me know if I can help.

frontend/eda/common/EdaResourcesCommon.tsx
lgalis: It is not possible, as only the in use credentials  have to be returned.

frontend/eda/common/EdaResourcesCommon.tsx
lgalis: Will not work, only the inUseResources have to be returned.

tiyiprh (approved):

PR #2209, opened 2024-04-30
Add terraform state inventory source

hakbailey:
Note: this is a resubmission of [PR 1868](https://github.com/ansible/ansible-ui/pull/1868) from a branch instead of a fork so checks can run. See previous review comments there.

Summary:

A new Terraform state inventory source was added to AWX in [PR 14840](https://github.com/ansible/awx/pull/14840). This PR updates the UI to include the new source type. It also adds Execution Environment selection to the inventory source add/edit forms, as an EE selection is required for the Terraform source -- users must provide and select an EE containing a Terraform binary, as we cannot ship one with the Terraform binary for licensing reasons.

In addition, this PR fixes a couple of related bugs: when trying to edit an inventory source that is not sourced from a project, the API returns a 400: Bad request error with detail "Cannot set source_path if not SCM type." due to a default value in the form for the source_path field. And the cypress `deleteAwxExecutionEnvironment` command had the wrong endpoint so this fixes that as well.


frontend/awx/main/ansibleDocsUrls.ts
```diff
@@ -15,4 +15,6 @@ export const ansibleDocUrls: { [key: string]: string } = {
     'https://docs.ansible.com/ansible/latest/collections/community/vmware/vmware_vm_inventory_inventory.html',
   constructed:
     'https://docs.ansible.com/ansible/latest/collections/ansible/builtin/constructed_inventory.html',
+  terraform:
+    'https://github.com/ansible-collections/cloud.terraform/blob/main/plugins/inventory/terraform_state.py',
```
AlexSCorey, line 19: I'm not 100% sure we want to link out to a github repo as official documentation.  I'm seeking input from people who would know.

frontend/awx/resources/sources/InventorySourceForm.tsx
```diff
@@ -131,14 +133,12 @@ export function EditInventorySource() {
...
+      execution_environment: inventorySource?.summary_fields?.execution_environment?.name,
       source: inventorySource?.source,
       credential: inventorySource?.summary_fields?.credential?.name,
       source_project: inventorySource?.summary_fields?.source_project,
       source_path: {
-        name:
-          inventorySource?.source_path?.length === 0
-            ? '/ (project root)'
-            : inventorySource?.source_path,
+        name: inventorySource?.source_path,
```
AlexSCorey, line 141: Given that this is in edit mode I do think we want to keep this line the way it was.  Inventory file is a required field and so as a user I would expected that field to be populated on when the form loads
Screenshot 2024-05-01 at 1 17 01 PM

AlexSCorey (changes requested):
Looks really good.  thanks for resubmitting this pr.  Just a few small things...and 1 lingering question for people about the documentation.

frontend/awx/main/ansibleDocsUrls.ts
AlexSCorey: Ok, did some digging.  Docs for this are pending publication.  The issue can be tracked here https://issues.redhat.com/browse/ACA-1187.  Might be worth while reaching out on that issue to see if they know the url that would link to terraform docs yet.  Sometimes they do know ahead of time where they will go.  For what its worth I suggested to Lynn Maynard that terraform docs should follow suit with the list of links to docs [here](https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.4/html-single/automation_controller_user_guide/index#ref-controller-inventory-sources)

frontend/awx/resources/sources/InventorySourceForm.tsx
hakbailey: I'm not sure I follow. If you're editing a source that already has an inventory file, it will load the file path. If you're creating a new source or changing the source type from something else to project, that field will be blank as there is no existing file chosen yet.

frontend/awx/main/ansibleDocsUrls.ts
hakbailey: @AlexSCorey that issue covers docs for the Terraform State inventory source in AAP, whereas here I think we want to link to the docs for the inventory plugin itself (which is also linked from those docs) so folks can understand what the plugin options are. We haven't done a cloud.terraform release yet with the new plugin so we can't link to published docs...we can either wait until that release happens and link to the plugin docs on Automation Hub, or we can link to the repo.

frontend/awx/resources/sources/InventorySourceForm.tsx
AlexSCorey: whoops, you are right.  I assumed the api worked a bit differently with the inventory file.

AlexSCorey (approved):
@hakbailey and I met offline and addressed all of my concerns.  This will merge after the cloud documentation merges.

gravesm:
We should be able to link to docs on AH now: https://console.redhat.com/ansible/automation-hub/repo/published/cloud/terraform/content/inventory/terraform_state/

frontend/awx/main/ansibleDocsUrls.ts
hakbailey: @AlexSCorey I updated the docs link and rebased onto current main to fix a merge conflict in the cypress test file (part of the credential test had been changed since I last worked on this). It looks like there are two failing tests but I don't _think_ they are related to this PR. Let me know if any other changes are needed, thanks!

PR #2213, opened 2024-04-30
Adds host modal details

AlexSCorey:
This addresses AAP-15469.  It adds the host event modal in the job output.  There will be follow on issues and prs.  Right now it does not show any YAML event data which would be rendered in the JSON/YAML tabs.

Old UI
Screenshot 2024-04-30 at 7 27 44 PM
Screenshot 2024-04-30 at 7 29 35 PM

### DIFF
--- a/framework/PageDetails/PageDetailCodeEditor.tsx
+++ b/framework/PageDetails/PageDetailCodeEditor.tsx
@@ -14,6 +14,7 @@ export function PageDetailCodeEditor(props: {
   label?: string;
   value: string;
   helpText?: string | ReactNode;
+  toggleLanguage?: boolean;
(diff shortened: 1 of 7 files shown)


marshmalien (commented):
I pulled this down and noticed that all of the job output rows were clickable. We should add a check that ensures the row is a clickable host event row. It looks like we can get that information from the jobEvent `const jobEvent = getJobOutputEvent(counter);`. 

https://github.com/ansible/awx/blob/devel/awx/ui/src/screens/Job/JobOutput/isHostEvent.js

keithjgrant (approved):
Looks good apart from the one minor issue Marly mentioned 👍🏻

PR #2215, opened 2024-05-01
[AAP-23637] Hook up run command wizard to toolbar buttons

prat98:
This PR is for hooking up the run command buttons on the inventory host and groups toolbars to the run command wizard.
It also removes the inventory group page action to run a command because the run command wizard only runs commands against entire inventories. So having a run command page action on a group is not necessary in my opinion. The ability to run a command from the inventory groups page is also not there in the old UI

### DIFF
--- a/frontend/awx/administration/execution-environments/components/PageFormExecutionEnvironmentSelect.tsx
+++ b/frontend/awx/administration/execution-environments/components/PageFormExecutionEnvironmentSelect.tsx
@@ -19,11 +19,18 @@ export function PageFormExecutionEnvironmentSelect<
   additionalControls?: ReactElement;
   isRequired?: boolean;
   label?: string;
+  placeholder?: string;
   isDisabled?: boolean;
   // tooltip?: string;
 }) {
-  const { name, organizationId, executionEnvironmentIdPath, executionEnvironmentPath, ...rest } =
-    props;
+  const {
+    name,
+    organizationId,
+    executionEnvironmentIdPath,
+    executionEnvironmentPath,
+    placeholder,
+    ...rest
+  } = props;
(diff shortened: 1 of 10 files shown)


akus062381:
Hey! I know this isn't specifically what this PR addresses, but since it's right in the same section of the UI, I was wondering whether you could edit the Placeholder text here, to something more like "Select Execution Environment" ? This is in the Execution Environment step of the Run Command Wizard.

image

akus062381:
When the user gets to the Review step of the Wizard, and clicks the Finish button, the UI doesn't respond. And no error shows in the Network tab. All that happens is the Finish button turns red. Are you able to fix this too? 

image

nixocio (commented):
hey @prat98 is there a related issue?

prat98:
@nixocio yes it's https://issues.redhat.com/browse/AAP-23637

nixocio (approved):

prat98:
> When the user gets to the Review step of the Wizard, and clicks the Finish button, the UI doesn't respond. And no error shows in the Network tab. All that happens is the Finish button turns red. Are you able to fix this too?
> 
> image

Fixed this issue, there was a problem with the way the request was being sent. Also updated the error adapter to pick up this error and show a generic error message in case the server returns an error like this again

prat98:
> Hey! I know this isn't specifically what this PR addresses, but since it's right in the same section of the UI, I was wondering whether you could edit the Placeholder text here, to something more like "Select Execution Environment" ? This is in the Execution Environment step of the Run Command Wizard.
> 
> image

Add a placeholder prop and updated the placeholder text to what you suggested

akus062381 (approved):
This looks great, thanks so much!

PR #2216, opened 2024-05-01
Fix Hidden Application Form Submit and Cancel Buttons

prat98:
Application form page submit and cancel buttons were hidden due to page layout. This PR fixes that.


vishsanghishetty:
@prat98 thanks for fixing this. I noticed this the other day and slipped out of my mind to file.

vishsanghishetty:
tested locally, this fixes the issue. Submit and Cancel buttons are visible on the Create form

PR #2221, opened 2024-05-02
Controller -> Fix Failing Tests- InventoryGroups and Project

akus062381:


kialam (changes requested):
I am unable to run an ad hoc command in the UI. This test case may be showing an actual bug in the UI.

kialam (approved):

PR #2222, opened 2024-05-02
Fix tests

MilanPospisil:
Run Ad Hoc command in Inventories -> Detail -> Groups -> Detail -> Run ad hoc command action.

The credentials modal flicking in wizard, when the select credential is clicked too fast after clicking next wizard page in previous wizard section. Waiting is included here to prevent flicking.


RedOrNot (approved):
LG 
thanks Milan

PR #2223, opened 2024-05-02
 [AAP-23629] Survey fixes and enhancements

jerabekjiri:
[AAP-23629](https://issues.redhat.com/browse/AAP-23629): **Fix validation error in `Prompt on Launch`**
Screenshot from 2024-05-02 11-40-17

**Use `PageFormMultiSelect` component for multi select option (with `PageFormCreatableSelect` user was able to create new option in typeahead option)**
Screenshot from 2024-05-01 15-00-40

### DIFF
--- a/frontend/awx/common/SurveyStep.tsx
+++ b/frontend/awx/common/SurveyStep.tsx
@@ -53,7 +53,7 @@ export function SurveyStep({
   );
 
   useEffect(() => {
-    const survey: { [key: string]: string | string[] | number | { name: string }[] } = {};
+    const survey: { [key: string]: string | string[] | number } = {};
     survey_spec?.spec?.forEach((obj: Spec) => {
       if (obj.default === '') {
         return;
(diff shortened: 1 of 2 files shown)


marshmalien (approved):
Pulled down and tested locally. LGTM 👍

frontend/awx/common/SurveyStep.tsx
```diff
@@ -53,16 +53,14 @@ export function SurveyStep({
   );
 
   useEffect(() => {
-    const survey: { [key: string]: string | number | { name: string }[] } = {};
+    const survey: { [key: string]: string | string[] | number | { name: string }[] } = {};
```
marshmalien, line 56: Do we still need to include `{ name: string }[]` in the survey key:value type?

jerabekjiri: probably not, I'll verify

PR #2224, opened 2024-05-02
Management jobs notifications list

MilanPospisil:
image

Also fixed the loading of approvals query /notification_templates_approvals/ that was sent also in notification list types that old application did not send and those requests failed. Now it does not sends the approvals requests for:

  resourceType === 'system_job_templates' ||
    resourceType === 'job_templates' ||
    resourceType === 'projects'


prat98 (approved):
Changes look good!

PR #2231, opened 2024-05-03
 Jobs output expandable section

MilanPospisil:
https://issues.redhat.com/browse/AAP-14459
https://issues.redhat.com/browse/AAP-14717

Find job with type Playbook run. Then detail and outputs.

What is added:
Topmost expand all expands Playbooks and Tasks outputs.
When section is expanded, it renders (...) rectangle under line number.

image

### DIFF
--- a/frontend/awx/views/jobs/JobOutput/JobOutputEvents.tsx
+++ b/frontend/awx/views/jobs/JobOutput/JobOutputEvents.tsx
@@ -86,8 +86,7 @@ export function JobOutputEvents(props: IJobOutputEventsProps) {
       [uuid]: collapsed,
       [counter]: collapsed,
     }));
-    if (collapsed === false)
-    {
+    if (collapsed === false) {
       setCollapsedAll(false);
     }
   };


frontend/awx/views/jobs/JobOutput/JobEventEllipsis.tsx
```diff
@@ -0,0 +1,16 @@
+import styled from 'styled-components';
+
+const Wrapper = styled.div`
+  border-radius: 1em;
+  background-color: var(--pf-global--BackgroundColor--light-200, white);
```
keithjgrant, line 5: This doesn't work in dark mode. I think this might be the wrong variable name; my dev tools say it isn't set to anything.

keithjgrant (changes requested):
This is great! Just a couple things I would suggest improving:

If you collapse all, then expand the Play, the "collapse all" button is still in the collapsed state and clicking it once doesn't expand/collapse anything. I think you might need to `setCollapsedAll(false)` any time a single item is expanded or collapsed.

frontend/awx/views/jobs/JobOutput/JobEventEllipsis.tsx
MilanPospisil: What about set it to white directly? It works in dark and white mode.

keithjgrant (approved):

PR #2232, opened 2024-05-03
[AAP-8163] Add user and team access tabs for inventories

lgalis:


vishsanghishetty:
@lgalis QA'ed locally
Credentials ✅  was able to add and remove user access to credentials
Inventory ✅ was able to add and remove user access to credentials. minor correction needed - after adding user access, when the page redirects and lands on the user access tab is not highlighted
image

vishsanghishetty (approved):

PR #2242, opened 2024-05-06
[AAP-17737] Update awxErrorAdapter to parse detail

nixocio:
Update awxErrorAdapter to parse detail from credential type error.

image


vishsanghishetty:
QA'ed, was able to reproduce the error [AAP-17737](https://issues.redhat.com/browse/AAP-17737) and the error message is handled as per the API is as expected

image

vishsanghishetty (approved):

prat98 (approved):
Looks good! Tested component tests locally and created error with `detail` key which was displayed to the user. It all worked as expected!

PR #2254, opened 2024-05-07
[AAP-8313 ] Disable edit credential field for kind vault and id vault_id

nixocio:
Disable edit credential field for kind vault and id vault_id

image


kialam (changes requested):
We should also disable the key button. Otherwise looks good.

nixocio:
> We should also disable the key button. Otherwise looks good.

Change updated.

kialam (approved):

PR #2256, opened 2024-05-07
Updates to support adding users to orgs

vidyanambiar:
[AAP-23146](https://issues.redhat.com/browse/AAP-23146)

1. Hide organization admin, member roles from the Org -> User roles wizard in Awx, Eda.
2. Allow EDA APIs to be filtered based on list based query params
3. Props to customize label and title for the roles wizard step for selecting AWX, EDA roles


akus062381 (approved):
Looks good!

PR #2270, opened 2024-05-09
Fix rendering form action buttons

tznamena:
This PR fixes rendering of the form for creating users. Before this change the "Create" and "Cancel" buttons will be outside of the view and page can't even be scrolled down, user must hit tab to get to those buttons. With this fix the buttons are in view.


vidyanambiar (approved):
lgtm!

akus062381 (approved):
Works well!

PR #2278, opened 2024-05-10
Notifier tests - check the detail after adding

MilanPospisil:
Check that notifier was added correctly - in detail, if fields get displayed.


cypress/e2e/awx/administration/notifiers/notifiersListView.cy.ts
```diff
@@ -127,6 +129,10 @@ function createNotification(type: string) {
     selectOrganization(orgName);
 
     cy.get(`[data-cy="Submit"]`).click();
+
+    // test defail
```
AlexSCorey, line 133: typo
```suggestion
    // test detail
```

AlexSCorey (approved):
I like this approach.  well done!

PR #2297, opened 2024-05-13
[AAP-23669] Instance Peering Bugfix

prat98:
This PR is for AAP-23669 which essentially about an issue with the peering instances list showing instances that are not possible to peer. This PR updates the network requests to get the list of peering instance candidates to match the request made in the old UI.


akus062381 (approved):
Your changes look great, thank you!

PR #2299, opened 2024-05-13
Adds logic to fix page crash when editing a schedule to add a rule

AlexSCorey:
Addresses AAP-23592.  Prevents page crash on schedules.


akus062381 (approved):
Can't reproduce the bug. Good work!

PR #2303, opened 2024-05-14
Fixes schedules-mgt-job-resource-link

vishsanghishetty:
BEFORE 
Clicking on the related resource column link of a Management job schedule
image

AFTER
Clicking on the related resource column link of a Management job schedule now
image


acruzgon (approved):
looks great!

PR #2305, opened 2024-05-14
Persist default values in re-enabling webhook and provisioning callback 

jerabekjiri:
Re-enabling `webhook_key` subform option or provisioning callback subform causes default values to be empty.

before:
chrome-capture-2024-5-14 (1)

after:
chrome-capture-2024-5-14


akus062381 (approved):
Thoroughly tested this, and it's looking great! Nice work!

PR #2318, opened 2024-05-15
update persistent filters in templates list

daphnemaeve:
Small change to change the persistent filters in ```TemplatesList.tsx``` from 'templatesList' to 'templates'.


vishsanghishetty (approved):
reviewed locally, works great!

PR #2319, opened 2024-05-15
Credential sub form tests

AlexSCorey:
This brings in tests for the credential sub form.  The tests simply confirm the existence of the necessary fields, and whether those fields are required.


kialam:
These tests look good. I wonder if it makes sense, since the subform is rendered dynamically based on the credential type selected in the main form, to have the tests assert the correct behavior is happening, vs using hard-coded values. For example, selecting Machine from the dropdown should render all fields as returned by the mock payload (`inputs.fields`) for `credential_type === machine`.

kialam (approved):

PR #2320, opened 2024-05-15
add roles tab tests for users

appuk:
https://issues.redhat.com/browse/AAP-21633


vishsanghishetty (commented):
I was able to run the tests successfully all the tests except users-role-tab hitting 400s

vishsanghishetty (approved):

PR #2324, opened 2024-05-15
Fix infinite credential type requests in InventoryRunCommand test

marshmalien:
* Update test fixture value to `"next": null` to prevent `useAwxGetAllPages` from making additional requests
* Type credential value instead of selecting from flakey popup modal

_Screenshot of failing test_
Screenshot 2024-05-15 at 3 50 41 PM


nixocio (approved):

akus062381 (approved):
Looks good!

PR #2330, opened 2024-05-16
Update AWX roles lists

keithjgrant:
Use new RBAC resources for Roles and Teams->Roles lists


frontend/awx/access/teams/TeamPage/TeamRoles.tsx
```diff
@@ -1,99 +1,15 @@
...
-      errorStateTitle={t('Error loading roles')}
-      emptyStateTitle={t('Team does not have any roles.')}
-      emptyStateDescription={t('To get started, add roles to the team.')}
-      emptyStateButtonIcon={<PlusCircleIcon />}
-      emptyStateButtonText={t('Add role to team')}
-      emptyStateButtonClick={() =>
-        pageNavigate(AwxRoute.AddRolesToTeam, { params: { id: team.id } })
-      }
-      {...view}
+    <ResourceAccess
```
vidyanambiar, line 8: Just a heads up - this particular change is related to a work item that @ZitaNemeckova will be addressing [AAP-20560](https://issues.redhat.com/browse/AAP-20560)

vidyanambiar (changes requested):
- Can we have the associated permissions for a role displayed in an expanded row (similar to EDA)?

- There is an error on pagination to page 3

image

keithjgrant:
> * There is an error on pagination to page 3

Looks like this was caused by the System Auditor role, which has no `content_type` field. I added a catch to cast that to "null" for the sake of the URL. Do you know if this just happens to be bad data in our environment?

vidyanambiar:
@keithjgrant The System Auditor role is unique in that it has a null content-type because it is applicable system-wide. This is expected.

vidyanambiar (approved):
Looks great! 🎉 

It would be good to get some component tests added.

PR #2333, opened 2024-05-17
PageForm hook up onClickAdditionalAction prop.

kialam:
This modifies the PageForm component to support a secondary button along with the default Submit and Cancel buttons in the PageForm and any actions associated with it.


kialam:
Closing in favor of using `additionalActions` prop.

PR #2337, opened 2024-05-17
Re-enable previously commented out test

vidyanambiar:
[AAP-24173](https://issues.redhat.com/browse/AAP-24173)

This test was skipped several months ago due to flakiness.
I updated the asserted tab count, ran the test locally a few times and it appears to pass consistently.


MilanPospisil (approved):
LGTM :+1

PR #2339, opened 2024-05-17
[AAP-21473] Update the no token info on the rulebook activation form

lgalis:
Screenshot from 2024-05-17 17-02-00


tiyiprh (approved):
LGTM!

PR #2351, opened 2024-05-20
Adds default values for boolean fields. Fixes navigation on edit save

AlexSCorey:
This addresses AAP-23689.  It also resolves an issue where in edit mode, saving an instance would navigate back to the instances list instead of the to the details view.


jerabekjiri (approved):
LGTM :+1:

PR #2360, opened 2024-05-21
Replace Roles in teams with one like EDA

ZitaNemeckova:


vidyanambiar (approved):
lgtm 👍 

It would be good to get some component tests added.

PR #2362, opened 2024-05-21
Addresses TODO 24149

AlexSCorey:
The scope ballooned more than I wanted it to here.  

This addresses AAP-24149:

This issue addressed a TODO that asked to replace AccessTable with ResourceAccessList.  However, ResourceAccessList was an outdated component by this point.  So, I decided to replace AccessTable with the current mock ups which separates UserRoles, and TeamsRoles tabs.  Each of those tabs use a full page wizard to add roles to the teams and users.

Additionally, I remove ResourceAccessList.tsx.  However, to do that I had to introduce a new workflow to the Teams --->Access tab.  This tab became called Users and follows the existing workflow throughout the app to add and remove users, and specific team roles.  The result is that now a user can be added to a team, and that user can be given specific team roles.....currently on main, a user can be added to a team, but cannot be given specific team roles.

Current implementation (on main)for Teams:
teamonMain

Change this PR brings in:
teamonFeature


frontend/awx/main/routes/useAwxOrganizationsRoutes.tsx
```diff
@@ -41,14 +43,14 @@ export function useAwxOrganizationRoutes() {
               element: <OrganizationDetails />,
             },
             {
-              id: AwxRoute.OrganizationUsers,
+              id: AwxRoute.OrganizationUsersAccess,
               path: 'users',
```
ZitaNemeckova, line 47: We are trying to unify url paths to use `user-access`. Could you change it, please?

frontend/awx/main/routes/useAwxOrganizationsRoutes.tsx
```diff
@@ -41,14 +43,14 @@ export function useAwxOrganizationRoutes() {
...
-              element: <OrganizationAccess />,
+              element: <OrganizationUserAccess />,
             },
             {
-              id: AwxRoute.OrganizationTeams,
+              id: AwxRoute.OrganizationTeamsAccess,
               path: 'teams',
```
ZitaNemeckova, line 52: We are trying to unify url paths to use `team-access`. Could you change it, please?

ZitaNemeckova (approved):
Tested in UI. LGTM 👍 Please change URL paths as mentioned before merging.

frontend/awx/main/routes/useAwxOrganizationsRoutes.tsx
vidyanambiar: @ZitaNemeckova I think that would only apply to resources such as projects, inventories, rulebook activations etc. where we have `User Access` and `Team Access` tabs on the resource.
For the Organizations page, the tab name and path can still be `users` as per the design and downstream implementation.

frontend/awx/main/routes/useAwxOrganizationsRoutes.tsx
vidyanambiar: same comment at [above](https://github.com/ansible/ansible-ui/pull/2362#discussion_r1617507202). This path matches what we have downstream for Organization -> Teams.

PR #2365, opened 2024-05-21
AWX Role add & edit

keithjgrant:


vidyanambiar (commented):
Works well 👍 

- Can we get the permissions to display in the details page, similar to the EDA role details?
- The role name does not need to be a link in the details page. 
- It would be good to get some unit test coverage added for roles list/details and forms either in this or a follow-on PR.

vidyanambiar (approved):

PR #2372, opened 2024-05-22
Fix case where filter values might get null

jamestalton:


vishsanghishetty:
approving this, James fixed this during a call and it worked!

vishsanghishetty (approved):

PR #2380, opened 2024-05-22
URL cleanup

keithjgrant:
* removes unused `:instanceType` param from instance group URLs
* removes unused `:resourceType` param from role URLs
* changes underscores in URLs to hyphens for consistency with the rest of the app

Note: after a closer look, I decided the `:inventory_type` param is still needed in the inventories page, so I left it in place. A constructed inventory can be fetched from the API at `/inventories/:id`, but using the URL param to fetch from `/constructed_inventories/:id` instead returns a few extra fields


vidyanambiar (approved):
looks good!

PR #2384, opened 2024-05-23
[AAP-24342] Create Separate Applications Table Component

prat98:
This PR creates a separate applications table component which allows for reusability. This change is needed specifically to support changes in the downstream which require having an tabbed applications section for both Awx and Platform level OAuth Applications. You can look at AAP-24342 for more info.


vishsanghishetty:
QA'ed locally, and everything works as expected 🚀

vishsanghishetty (approved):

PR #2388, opened 2024-05-23
[AAP-24296] Add Encrypted field to CredentialTextInput

daphnemaeve:
Adds encrypted field to CredentialTextInput that is shown when field isn't empty but is a secret field.

Screenshot 2024-05-23 at 4 50 47 PM

### DIFF
--- a/cypress/e2e/awx/resources/credentials.cy.ts
+++ b/cypress/e2e/awx/resources/credentials.cy.ts
@@ -86,8 +86,16 @@ describe('Credentials', () => {
       cy.verifyPageTitle('Edit Credential');
       cy.get('[data-cy="vault-id"]').should('have.attr', 'disabled');
       cy.get('[data-cy="vault-password"]').should('be.visible');
-      cy.get('[data-cy="vault-password"]').clear().type('new password');
+      cy.get('[data-cy="vault-password"]').then(($input) => {
+        expect($input.val()).to.eq('ENCRYPTED');
+      });
+      cy.get('[data-cy="ask_vault_password"]').check();
       cy.clickButton(/^Save credential$/);
+      cy.get('[data-cy="name"]').contains(credentialName);
+      cy.contains('Vault Identifier').should('be.visible');
+      cy.get('[data-cy="vault-identifier"]').contains('id');
+      cy.contains('Vault Password').should('be.visible');
+      cy.get('[data-cy="vault-password"]').contains('Prompt on launch');
       //delete created credential
       cy.clickPageAction('delete-credential');
       cy.get('#confirm').click();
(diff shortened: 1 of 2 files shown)


skochay330 (changes requested):

kialam:
This is looking really close. I noticed that empty fields have the revert/redo button but in the old UI those buttons are not present unless a field has an encrypted value:
Old UI:
Screenshot 2024-06-07 at 2 27 06 PM


New UI:
Screenshot 2024-06-07 at 2 26 31 PM

frontend/awx/access/credentials/CredentialForm.tsx
```diff
@@ -818,70 +862,133 @@ function CredentialTextInput({
...
+    );
+    setShouldHideField(true);
+  };
 
   useEffect(() => {
-    // mark field as managed by a credential plugin
-    if (accumulatedPluginValues.some((cp) => cp.input_field_name === field.id)) {
-      setValue(field.id, renderFieldValue(field), { shouldDirty: true });
+    if (shouldHideField && isSecretField && isInitialValueEncrypted && !isPromptOnLaunchChecked) {
+      setValue(field.id, t('ENCRYPTED'), { shouldDirty: false });
```
kialam, line 880: This actually updates the stored password value that the API returns (`$encrypted$`) to `ENCRYPTED` and sends it as the new string to the API, which we don't want. We should set it to `initialValues[field.id]` so that it doesn't change the value.

kialam (changes requested):

marshmalien (approved):
👍 Great work! It's a very complex area of the project. I pulled down and compared the changes locally against the old UI. Looked good!

nixocio:
@marshmalien @kialam I rebased and update credential tests, now there is one credential test failing but it is a known failure due to a cypress command that needs to be updated.

cypress/e2e/awx/resources/credentials.cy.ts
```diff
@@ -613,7 +621,7 @@ describe('Credentials Tabbed View - Job Templates', function () {
     cy.selectDropdownOptionByResourceName('inventory', awxInventory.name);
     cy.selectDropdownOptionByResourceName('project', `${(this.globalProject as Project).name}`);
     cy.selectDropdownOptionByResourceName('playbook', 'hello_world.yml');
-    cy.selectItemFromLookupModal('credential-select', machineCredential.name);
```
kialam, line 616: We should revert this change. `selectItemFromLookupModal ` is the correct command.

marshmalien (approved):

cypress/e2e/awx/resources/credentials.cy.ts
nixocio: @kialam reverted.

nixocio:

kialam (approved):

skochay330 (approved):

nixocio:

PR #2392, opened 2024-05-27
fix inventory type and id in hosts -> groups navigation

MilanPospisil:
Navigation was broken in Hosts -> Groups -> click the group link that will navigate it to inventories -> groups. Now the hosts does not have inventory_type param and it uses hook for columns that get this missing param. I added props to the hook so the params is get from host summary inventory field.


jerabekjiri (approved):
Nice :+1:

PR #2401, opened 2024-05-29
[AAP-24295] Credentials multiline encryption

skochay330:
Jira issue: https://issues.redhat.com/browse/AAP-24295

Before:
Screenshot 2024-05-29 at 12 30 41 AM

After: 
Screenshot 2024-05-29 at 12 34 29 AM

### DIFF
--- a/frontend/awx/access/credentials/components/CredentialMultilineInput.tsx
+++ b/frontend/awx/access/credentials/components/CredentialMultilineInput.tsx
@@ -1,5 +1,5 @@
-import { Button } from '@patternfly/react-core';
-import { KeyIcon } from '@patternfly/react-icons';
+import { Button, Tooltip } from '@patternfly/react-core';
+import { KeyIcon, UndoIcon } from '@patternfly/react-icons';
 import { PageFormFileUpload } from '../../../../../framework/PageForm/Inputs/PageFormFileUpload';
 import { CredentialInputField, CredentialType } from '../../../interfaces/CredentialType';
 import { useTranslation } from 'react-i18next';
@@ -7,8 +7,9 @@ import { Credential } from '../../../interfaces/Credential';
 import { useGetItem } from '../../../../common/crud/useGet';
 import { awxAPI } from '../../../common/api/awx-utils';
 import { CredentialPluginsInputSource } from '../CredentialPlugins/hooks/useCredentialPluginsDialog';
-import { useCallback, useEffect } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { useFormContext } from 'react-hook-form';
+import { initialValues } from '../CredentialForm';
 
 export function CredentialMultilineInput({
   field,
@@ -18,6 +19,7 @@ export function CredentialMultilineInput({
   accumulatedPluginValues,
   setAccumulatedPluginValues,
   setPluginsToDelete,
+  initialValues,
 }: {
   field: CredentialInputField;
   requiredFields: CredentialType['inputs']['required'];
@@ -26,16 +28,21 @@ export function CredentialMultilineInput({
   accumulatedPluginValues: CredentialPluginsInputSource[];
   setAccumulatedPluginValues?: (values: CredentialPluginsInputSource[]) => void;
   setPluginsToDelete?: React.Dispatch<React.SetStateAction<string[]>>;
+  initialValues?: initialValues;
 }) {
   const { t } = useTranslation();
-  const { setValue, clearErrors } = useFormContext();
+  const [shouldHideField, setShouldHideField] = useState(field.secret);
+  const { getValues, setValue, clearErrors } = useFormContext();
+  const watchedFieldValue = getValues();
   const onClear = () => {
     setValue(field.id, '', { shouldDirty: false });
     clearErrors(field.id);
+    if (accumulatedPluginValues.some((cp) => cp.input_field_name === field.id)) {
+      setPluginsToDelete?.((prev: string[]) => [...prev, field.id]);
+    }
     setAccumulatedPluginValues?.(
       accumulatedPluginValues.filter((cp) => cp.input_field_name !== field.id)
     );
-    setPluginsToDelete?.((prev: string[]) => [...prev, field.id]);
   };
   const useGetSourceCredential = (id: number) => {
     const { data } = useGetItem<Credential>(awxAPI`/credentials/`, id);
@@ -87,6 +94,18 @@ export function CredentialMultilineInput({
     }
   }, [setValue, accumulatedPluginValues, renderFieldValue, field]);
 
+  const clearField = () => {
+    setValue(field.id, '', { shouldDirty: false });
+    setShouldHideField(!shouldHideField);
+  };
+  const hideField = () => {
+    setValue(field.id, initialValues?.[field.id], { shouldDirty: true });
+    setShouldHideField(!shouldHideField);
+    setAccumulatedPluginValues?.(
+      accumulatedPluginValues.filter((cp) => cp.input_field_name !== field.id)
+    );
+  };
+
   return (
     <>
       <PageFormFileUpload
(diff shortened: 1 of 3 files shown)


frontend/awx/access/credentials/components/CredentialMultilineInput.tsx
```diff
@@ -26,16 +32,24 @@ export function CredentialMultilineInput({
...
+  const watchedFieldValue = getValues();
   const { setValue, clearErrors } = useFormContext();
   const onClear = () => {
     setValue(field.id, '', { shouldDirty: false });
     clearErrors(field.id);
     setAccumulatedPluginValues?.(
       accumulatedPluginValues.filter((cp) => cp.input_field_name !== field.id)
     );
-    setPluginsToDelete?.((prev: string[]) => [...prev, field.id]);
+    watchedFieldValue?.[field.id]?.includes('Value is managed by')
```
nixocio, line 50: This string is selected to be translated...we might re-evaluate this logic.

frontend/awx/access/credentials/components/CredentialMultilineInput.tsx
```diff
@@ -100,10 +126,39 @@ export function CredentialMultilineInput({
...
-            <Button icon={<KeyIcon />} variant="control" onClick={handleModalToggle} />
+            isSecret && watchedFieldValue?.[field.id] === '$encrypted$' ? (
+              <div style={{ display: 'grid' }}>
+                <Button
+                  icon={<KeyIcon />}
+                  variant="control"
+                  onClick={handleModalToggle}
+                  isDisabled
+                />
+                <Tooltip content="Replace">
```
nixocio, line 144: This one should be selected to be translated.

frontend/awx/access/credentials/components/CredentialMultilineInput.tsx
```diff
@@ -100,10 +126,39 @@ export function CredentialMultilineInput({
...
+                </Tooltip>
+              </div>
+            ) : isSecret &&
+              initialValues?.[field.id] &&
+              watchedFieldValue?.[field.id] !== '$encrypted$' &&
+              !watchedFieldValue?.[field.id]?.includes('Value is managed by') &&
+              !pluginsToDelete?.includes(field.id) ? (
+              <div style={{ display: 'grid' }}>
+                <Button icon={<KeyIcon />} variant="control" onClick={handleModalToggle} />
+                <Tooltip content="Revert">
```
nixocio, line 155: This one should be selected to be translated.

frontend/awx/access/credentials/components/CredentialMultilineInput.tsx
```diff
@@ -100,10 +126,39 @@ export function CredentialMultilineInput({
...
+                  isDisabled
+                />
+                <Tooltip content="Replace">
+                  <Button icon={<UndoIcon />} variant="control" onClick={() => clearField()} />
+                </Tooltip>
+              </div>
+            ) : isSecret &&
+              initialValues?.[field.id] &&
+              watchedFieldValue?.[field.id] !== '$encrypted$' &&
+              !watchedFieldValue?.[field.id]?.includes('Value is managed by') &&
```
nixocio, line 151: Same comment as above. Using strings that will be translated in logic.

nixocio (changes requested):
@skochay330 on the old implementation we had buttons in different sides of the input.


image

Could we mimic the same design?

kialam:
> @skochay330 on the old implementation we had buttons in different sides of the input.
> 
> image
> 
> Could we mimic the same design?

@tiyiprh @trahman73 thoughts on this? We'd have to do a bit more logic to try and make that happen, whereas the current implementation has both actionable buttons stacked on the right hand side.

trahman73:
If the stacked button implementation is working fine, we are good to go.

frontend/awx/access/credentials/components/CredentialMultilineInput.tsx
kialam: @nixocio good point. @skochay330 i think we can check `accumulatedPluginValues` to see if the credential was a plugin, and then mark it for deletion. I would do this prior to filtering it out of the `accumulatedPluginValues` array:

```tsx
// if field value was previously managed by a credential plugin, mark it for deletion
    if (accumulatedPluginValues.some((cp) => cp.input_field_name === field.id)) {
      setPluginsToDelete?.((prev: string[]) => [...prev, field.id]);
    }
```

frontend/awx/access/credentials/components/CredentialMultilineInput.tsx
```diff
@@ -100,10 +126,39 @@ export function CredentialMultilineInput({
...
+              !pluginsToDelete?.includes(field.id) ? (
+              <div style={{ display: 'grid' }}>
+                <Button icon={<KeyIcon />} variant="control" onClick={handleModalToggle} />
+                <Tooltip content="Revert">
+                  <Button icon={<UndoIcon />} variant="control" onClick={() => hideField()} />
```
kialam, line 156:
```suggestion
                  <Button icon={<UndoIcon />} variant="control" onClick={hideField} />
```

frontend/awx/access/credentials/components/CredentialMultilineInput.tsx
```diff
@@ -100,10 +126,39 @@ export function CredentialMultilineInput({
...
+            isSecret && watchedFieldValue?.[field.id] === '$encrypted$' ? (
+              <div style={{ display: 'grid' }}>
+                <Button
+                  icon={<KeyIcon />}
+                  variant="control"
+                  onClick={handleModalToggle}
+                  isDisabled
+                />
+                <Tooltip content="Replace">
+                  <Button icon={<UndoIcon />} variant="control" onClick={() => clearField()} />
```
kialam, line 145:
```suggestion
                  <Button icon={<UndoIcon />} variant="control" onClick={clearField} />
```

frontend/awx/access/credentials/components/CredentialMultilineInput.tsx
```diff
@@ -26,16 +32,24 @@ export function CredentialMultilineInput({
   accumulatedPluginValues: CredentialPluginsInputSource[];
   setAccumulatedPluginValues?: (values: CredentialPluginsInputSource[]) => void;
   setPluginsToDelete?: React.Dispatch<React.SetStateAction<string[]>>;
+  isSecret?: boolean;
+  pluginsToDelete?: string[];
+  initialValues?: initialValues;
 }) {
   const { t } = useTranslation();
+  const [shouldHideField, setShouldHideField] = useState(isSecret);
```
kialam, line 40: Since we have access to `field` already, we can set isSecret to just field.secret instead of passing it as a prop.

frontend/awx/access/credentials/components/CredentialMultilineInput.tsx
kialam, line 151: If this is to check whether or not the value is managed by a credential plug in you can check for it's existence in `accumulatedPluginValues` array.

skochay330:
> This is looking really close to complete 🥇I noticed this while testing the branch locally: Given this flow:
> 
> * User has encrypted field
> * Replace
> * Select credential plugin and click Finish on the modal
> * There is no Revert button to allow them to go back to the previous value
> 
> We should allow the user to revert back to a previously set value if there is one.

Old UI does not have the revert button when there is a plugin selected, but if we want that then I can just remove the condition on line 151 we talked about here https://github.com/ansible/ansible-ui/pull/2401#discussion_r1619490049

tiyiprh:
> > This is looking really close to complete 🥇I noticed this while testing the branch locally: Given this flow:
> > 
> > * User has encrypted field
> > * Replace
> > * Select credential plugin and click Finish on the modal
> > * There is no Revert button to allow them to go back to the previous value
> > 
> > We should allow the user to revert back to a previously set value if there is one.
> 
> Old UI does not have the revert button when there is a plugin selected, but if we want that I can play around with the conditions some more. @trahman73 @tiyiprh thoughts?

I'm not sure if I fully understand but if there is a use case for adding a revert button and it doesn't require too much extra effort, I would say we should add it. But at this point, if it's extra work maybe it can be a 2.5-next item and we just do feature parity and mirror what is in the current UI.

nixocio (approved):
@skochay330 good job.

PR #2406, opened 2024-05-29
Credentials tests

vishsanghishetty:
## Description
This PR is for [ticket](https://issues.redhat.com/browse/AAP-22922) 

## Acceptance Criteria
**Tests**
- [x] Credential creation of 30 different credential types
- [x] `credentialsTestData` with test data for field names and credential type `data-cys` for the 30 different credential types
- [x] Create a new credential, assign a team, and apply role(s) to test
- [x] Create a new credential, assign a user, and apply role(s) to test
- [x] Can create a job template within the context of the credential job template tab

### DIFF
--- a/cypress/e2e/awx/resources/credentials.cy.ts
+++ b/cypress/e2e/awx/resources/credentials.cy.ts
@@ -1,8 +1,11 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import { randomString } from '../../../../framework/utils/random-string';
 import { Credential } from '../../../../frontend/awx/interfaces/Credential';
 import { Organization } from '../../../../frontend/awx/interfaces/Organization';
 import { AwxUser } from '../../../../frontend/awx/interfaces/User';
 import { awxAPI } from '../../../support/formatApiPathForAwx';
+import { randomE2Ename } from '../../../support/utils';
+import { Team } from '../../../../frontend/awx/interfaces/Team';
 
 describe('Credentials', () => {
   let organization: Organization;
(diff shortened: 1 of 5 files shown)


cypress/support/awx-commands.ts
```diff
@@ -514,6 +514,15 @@ Cypress.Commands.add(
   }
 );
 
+Cypress.Commands.add(
+  'clickTableRowRoleAction',
```
acruzgon, line 518: just curious why not use `clickTableRowPinnedAction` or `clickTableRowAction`, they look very similar to this new one

cypress/e2e/awx/resources/credentials.cy.ts
```diff
@@ -493,3 +482,307 @@ describe('Credentials', () => {
...
+      cy.createAwxTeam(awxOrganization).then((awxTeam) => {
+        createdAwxTeam = awxTeam;
+      });
+      cy.createAwxInventory().then((inv) => {
+        awxInventory = inv;
+      });
+    });
+  });
+
+  after(() => {
```
akus062381, line 576: I believe this should be an afterEach block, since these resources are created in a beforeEach block; will help with proper resource cleanup.

akus062381 (changes requested):

cypress/e2e/awx/resources/credentials.cy.ts
vishsanghishetty: since there is only one test in the `describe` block, using `after` is fine and works as expected. `afterEach` would be when we have multiple it blocks inside the `describe` and we want the resource cleanup to happen `after` each test

cypress/support/awx-commands.ts
vishsanghishetty: @acruzgon good catch, we have so many custom commands now 😄 always good to know a new one
updating the command.

akus062381 (changes requested):
I'd still like to see the after block changed to an afterEach block, to be congruous with the beforeEach block. Especially if another test ever needs to be added to that describe block in future. Thanks!

vishsanghishetty:
@acruzgon @akus062381 I have updated the PR. I would like to get this PR merged if you can approve it.

acruzgon (approved):
lgtm!

akus062381 (approved):

PR #2407, opened 2024-05-30
Fix Inventory hosts groups missing run command button

MilanPospisil:
The run command button was missing in inventories -> detail -> hosts -> detail -> groups.

Note that the groups are showing only in normal inventories, not smart or constructed.


fullsushidev (approved):

prat98 (approved):
Changes look good to me 🚀

PR #2411, opened 2024-05-30
Fix error when creating credential with custom type without input

nixocio:
Fix error when creating credential with custom type without input. Also, add test.

To test: 

1. Create a custom credential type without inputs
2. Create a credential using the previous item
3. No error should be raised


akus062381 (approved):
Looks good!

vishsanghishetty (approved):
QA'ed, works as expected, thanks for fixing this!

PR #2412, opened 2024-05-31
Add Sync All Button for inventory sources

MilanPospisil:
Adds Sync All buttons in Inventories -> Detail -> Tab Sources -> Sources list

Sync all is calling single API, so its not classic bulk action which calls multiple API calls for every item. It calls single API endpoint that starts syncing for all items.

Also row action to launch sync (and in detail) is now much more responsive because it immediately calls refresh of the page instead of waiting for periodic reload, which may happen somewhere from 0 to 5 seconds.

### DIFF
--- a/frontend/awx/resources/inventories/hooks/useInventoriesSourcesToolbarActions.tsx
+++ b/frontend/awx/resources/inventories/hooks/useInventoriesSourcesToolbarActions.tsx
@@ -92,7 +92,7 @@ export function useInventoriesSourcesToolbarActions(
         type: PageActionType.Button,
         selection: PageActionSelection.None,
         isPinned: false,
-        label: t('Sync All'),
+        label: t('Launch inventory update'),
         onClick: syncAll,
         isDisabled: () => cannotLaunchInventorySourcesUpdate,
       },
--- a/frontend/awx/resources/inventories/hooks/useInventoriesSourcesToolbarActions.tsx
+++ b/frontend/awx/resources/inventories/hooks/useInventoriesSourcesToolbarActions.tsx
@@ -1,5 +1,5 @@
 import { ButtonVariant } from '@patternfly/react-core';
-import { PlusCircleIcon, TrashIcon } from '@patternfly/react-icons';
+import { PlusCircleIcon, RocketIcon, TrashIcon } from '@patternfly/react-icons';
 import { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useParams } from 'react-router-dom';
@@ -91,7 +91,7 @@ export function useInventoriesSourcesToolbarActions(
       {
         type: PageActionType.Button,
         selection: PageActionSelection.None,
-        isPinned: false,
+        icon: RocketIcon,
         label: t('Launch inventory update'),
         onClick: syncAll,
         isDisabled: () => cannotLaunchInventorySourcesUpdate,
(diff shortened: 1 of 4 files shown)


frontend/awx/resources/inventories/hooks/useInventoriesSourcesToolbarActions.tsx
```diff
@@ -61,7 +88,47 @@ export function useInventoriesSourcesToolbarActions(view: IAwxView<InventorySour
...
+  const alertToaster = usePageAlertToaster();
+  const { t } = useTranslation();
+
+  return () => {
+    void (async () => {
+      try {
+        await postRequest(url, {});
+        void refresh();
+      } catch (error) {
+        alertToaster.addAlert({
```
skochay330, line 125: I don't see any alerts when the sync fails- when should this be visible?

frontend/awx/resources/inventories/hooks/useInventoriesSourcesToolbarActions.tsx
```diff
@@ -61,7 +88,47 @@ export function useInventoriesSourcesToolbarActions(view: IAwxView<InventorySour
         isDanger: true,
         isDisabled: (sources: InventorySource[]) => cannotDeleteResources(sources, t),
       },
+      {
+        type: PageActionType.Button,
```
skochay330, line 92: I know we have this as a separate button in old UI, but maybe it would be more consistent with framework patterns if we put it in the bulk actions kebab menu? @trahman73 @tiyiprh what do ya'll think?

skochay330 (commented):
This looks good so far! Just a couple of UI issues- alerts don't seem to be working as expected, and maybe we want to move the sync button inside of the kebab

frontend/awx/resources/inventories/hooks/useInventoriesSourcesToolbarActions.tsx
tiyiprh: I think that makes sense!

frontend/awx/resources/inventories/hooks/useInventoriesSourcesToolbarActions.tsx
MilanPospisil: This should be visible only when sync request fails, not the sync operation running as task. I dont see any alerts of failing operation in the old app...

frontend/awx/resources/inventories/hooks/useInventoriesSourcesToolbarActions.tsx
MilanPospisil: I can do that, but we have plenty of buttons outside of kebab. It depends on how frequently will user use this option, clicking kebab is additional click, which is unecessary... Kebab menu makes sense when there is no space for all buttons, here, we have plenty of space.

tiyiprh: Do you all know if syncing is common for users? If so, agree that it can be outside of the kebab but if it's not common for users to sync, it should remain in the kebab.

AlexSCorey: It probably isn't super common to sync from the list.  it is probably much more common to sync an inventory source from within a workflow job template...which doesn't involve the list.  Thus I think putting the button in the kebab on the toolbar would be fine.

jerabekjiri:
Should we unify the text in buttons? The old UI uses `Sync` (`Sync All`), while the new UI uses `Launch inventory update`. @MilanPospisil @AlexSCorey @tiyiprh

AlexSCorey:
@MilanPospisil I'm ok with `Launch inventory update` as long as we are consistent across the new UI

skochay330:
We probably want to add the rocket icon and update the wording on the dropdown item to indicate that it would be launching the inventory update for **all** items

Screenshot 2024-06-04 at 4 00 24 PM

tiyiprh:
> We probably want to add the rocket icon and update the wording on the dropdown item to indicate that it would be launching the inventory update for **all** items
> 
> Screenshot 2024-06-04 at 4 00 24 PM

I agree that we should add a launch icon. For the wording, since the toolbar actions indicate a bulk action, it can probably just be "Launch inventory updates"

skochay330 (approved):

PR #2417, opened 2024-05-31
Run command ignoring selection fixed

MilanPospisil:
Run command ignores selected hosts/groups. It sends in body POST: limit="all", even when some hosts or groups are selected, the limit is still "all".

This PR fixes it for toolbars and row actions.

It uses URL as list of selected items for short selections, for longer, it uses local storage.

### DIFF
--- a/frontend/awx/resources/inventories/InventoryRunCommand.tsx
+++ b/frontend/awx/resources/inventories/InventoryRunCommand.tsx
@@ -31,7 +31,6 @@ export function InventoryRunCommand() {
 
   const pageNavigate = usePageNavigate();
   const { data: inventory } = useGet<Inventory>(awxAPI`/inventories/${id as string}/`);
-  
 
   const navigate = useNavigate();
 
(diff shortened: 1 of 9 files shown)


frontend/awx/resources/inventories/InventoryRunCommand.tsx
```diff
@@ -20,11 +20,20 @@ import { Inventory, RunCommandWizard } from '../../interfaces/Inventory';
...
 
 export function InventoryRunCommand() {
   const { t } = useTranslation();
   const getPageUrl = useGetPageUrl();
   const { id } = useParams();
+  const [searchParams] = useURLSearchParams();
+  let limit = searchParams.get('limit') || 'all';
+  const key = searchParams.get('key');
+
+  if (key) {
```
MilanPospisil, line 33: TODO: Clear item!

MilanPospisil: Done.

AlexSCorey (approved):
This looks good.  I do think we should change the name of that 1 variable but I'll leave the final decision up to you.

PR #2420, opened 2024-05-31
[AAP-24645] Instance Details Switch

prat98:
Moved instance switch to instance details section so topology view side bar has access to it. Customers can enable and disable an instance from the instances list, instance detail page and topology sidebar now (Just added).
Related JIRA: AAP-24645

### DIFF
--- a/cypress/e2e/awx/administration/instances.cy.ts
+++ b/cypress/e2e/awx/administration/instances.cy.ts
@@ -112,7 +112,7 @@ describe('Instances: Add/Edit', () => {
     });
   });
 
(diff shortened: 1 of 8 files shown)


AlexSCorey (changes requested):
Consider using the PageActionSwitch in the details file instead of creating a new Switch file.  Also, lets make the text `Unavailable` red like it is in the list view. Screenshot 2024-06-03 at 3 13 48 PM

prat98:
Made the requested changes

AlexSCorey (commented):
Looks good to me.

AlexSCorey (approved):

PR #2427, opened 2024-06-03
Adds link to org in Application Details and Copy to EEs

AlexSCorey:
This addresses AAP-24577 and AAP-24766.


prat98 (approved):
Changes look good just need to update the copy action. It currently has an edit icon instead of a copy icon next to it
Screenshot 2024-06-04 at 10 19 38 AM

PR #2428, opened 2024-06-03
[AAP-24675] Add download button for jobs output

prat98:
This PR addresses AAP-24675. Users should now be able to download logs from a job page where possible. This download button is in the kebab menu and is hidden if there is no `related.stdout` value for a job


frontend/awx/views/jobs/hooks/useJobHeaderActions.tsx
```diff
@@ -70,7 +72,16 @@ export function useJobHeaderActions(onComplete: (jobs: UnifiedJob[]) => void) {
         ouiaId: 'job-detail-delete-button',
         isDanger: true,
       },
+      {
```
AlexSCorey, line 75: Lets put the download button above the delete button and but a separator between the 2 items.

AlexSCorey (changes requested):

frontend/awx/views/jobs/hooks/useJobHeaderActions.tsx
prat98: done

PR #2430, opened 2024-06-03
Adds pod spec override to Container group Details

AlexSCorey:
It also enables the functionality to save Inventory Source Vars on the constructed inventory form.

AAP-24821 and AAP-24764


marshmalien (approved):
pulled down and lgtm!

PR #2431, opened 2024-06-03
Hide client secret field if client type is public

prat98:
This PR fixes an issue where the client secret modal displays a client secret field for an application with a `public` client type. The empty field throws an error when attempting to use the copy button for the field. More info can be found here https://github.com/ansible/aap-ui/pull/297


vidyanambiar (approved):
lgtm!

PR #2437, opened 2024-06-04
Inventories - Reorder tabs

MilanPospisil:


jerabekjiri (approved):
LGTM, but this feels like the only screen that doesn't have the access tabs on the last

MilanPospisil:
> LGTM, but this feels like the only screen that doesn't have the access tabs on the last

Agreed, we must change the order of the rest to match this, provided UIX is aware that all of Access tabs must go first right after detail tab.

PR #2441, opened 2024-06-04
[AAP-24644] Fixes scroll on Topology Sidebar

AlexSCorey:
This addresses AAP-24644 by adding styling via styled components.  This works for now, but in the future we should figure out a better solution.


jerabekjiri:
Looks like there is a white panel at the bottom of my screen in the topology view.

Screenshot from 2024-06-05 12-05-32

prat98 (approved):
LGTM! 🚀

PR #2443, opened 2024-06-05
Access control for managing/adding roles to users and teams

vidyanambiar:
- `Add roles` for a team is enabled if the logged in user has edit permissions on the team.
- `Add roles` for a user is enabled if the logged in user has edit permissions on the user.
- The `Manage (org) roles` dialog hides the "Manage roles" action if the logged in user does not have permissions to edit the organization.

### DIFF
--- a/frontend/awx/access/teams/TeamPage/AwxTeamRoles.tsx
+++ b/frontend/awx/access/teams/TeamPage/AwxTeamRoles.tsx
@@ -1,15 +1,41 @@
 import { useParams } from 'react-router-dom';
 import { ResourceAccess } from '../../../../common/access/components/ResourceAccess';
 import { AwxRoute } from '../../../main/AwxRoutes';
+import { useOptions } from '../../../../common/crud/useOptions';
+import { ActionsResponse, OptionsResponse } from '../../../interfaces/OptionsResponse';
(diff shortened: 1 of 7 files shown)


AlanCoding:
Our permission policies for "associations" are very old and have not ever really changed. For associating/linking two objects there is a notion of a "primary" object, we'll call it A, and the "associated" object, we'll call it B, so the B is associated to A

A <- B

But you see that as we create duplicate API endpoints for doing the same thing, `/api/v2/roles/:id/users/` and `/api/v2/users/:id/roles/`, you automatically create a situation where they have "different" rules, because the "parent" object in the endpoint may be either the primary object or the associated object. So the logic has to translate that.

The policy is:
 - user needs change/admin access to A
 - user needs read access to B

For role assignments, the content object is the primary object, and the user/team is the associated object.

> `Add roles` for a user is enabled if the logged in user ~has edit permissions on~ the user.

So all that leadup is to suggest the edit ~has edit permissions on~ --> "can view"

vidyanambiar:
@AlanCoding Thanks for sharing your perspective here.
As discussed in the RBAC standup meeting this morning, we should leave the Add/Manage roles button enabled for users/teams that the logged in user can view. If there is a gap with the logged in user's permissions the role assignment API will fail.
I'm closing this PR.

PR #2447, opened 2024-06-05
Watch dog for poll command to prevent infinite loop

MilanPospisil:
Fix to prevent infinite loop, now it will wait only 10 times.


akus062381 (approved):
Thank you!!

PR #2453, opened 2024-06-05
Fix secret fields with prompt on launch options.

kialam:


nixocio (approved):
There not related tests failures that are being skipped by another PR.

PR #2416, opened 2024-05-31
Adds Activity stream icon to all nested details views

AlexSCorey:
This addresses AAP-24610, AAP-24609 and AAP-24608.

It adds the activity stream icon everywhere it is needed in the nested/details views.

PR #2456, opened 2024-06-05
Adds activity stream to all awx page views

AlexSCorey:
This does the same work as https://github.com/ansible/ansible-ui/pull/2416 however it takes a little bit of a different approach.  #2416 needs to be redone because it breaks the package build.  I


prat98 (approved):
LGTM!

PR #2460, opened 2024-06-05
Add Cancel Sync Inventory Option

prat98:
This PR addresses AAP-24920. It adds the option to cancel an inventory sync when one is running. In addition, websockets are used to track status changes of the inventory sync jobs.


AlexSCorey (approved):
Looks good to me.

PR #2466, opened 2024-06-06
lint-staged

jamestalton:
https://github.com/lint-staged/lint-staged

Much faster cleaner way to ensure a commit only contains files that pass eslint and prettier.
Only runs prettier and eslint on the files that changed.


vishsanghishetty (approved):
:rocket

PR #2468, opened 2024-06-06
Adds logic for inventory type routes from activity stream

AlexSCorey:
This addresses AAP-24153.  It adds logic to the activity stream code that helps determine which inventory type the user clicks on so the correct url can be created.


prat98 (approved):
LGTM! 🚀

PR #2470, opened 2024-06-06
Adds help text to job details

AlexSCorey:
This addresses AAP-24726.

It also adds a helpText property to `ITableColumnCommon` so that we can define help text on each column and then <PageDetailsFromColumns> can consume, and render the help text in the details views


prat98 (approved):
LGTM! 🚀

PR #2471, opened 2024-06-06
Add optional props for passing empty state and button labels to ControllerTokens

tznamena:
This PR adds optional props in the ControllerTokens component so that empty state labels can be modified.


vidyanambiar (approved):
lgtm!

PR #2474, opened 2024-06-06
Add file upload option & password fields to settings

prat98:
This PR is for AAP-24760 and AAP-24670. It adds a password field to all secrets in the settings forms and adds fileupload forms to the SAML settings section

### DIFF
--- a/frontend/awx/administration/settings/AwxSettingsForm.tsx
+++ b/frontend/awx/administration/settings/AwxSettingsForm.tsx
@@ -12,6 +12,7 @@ import { usePatchRequest } from '../../../common/crud/usePatchRequest';
 import { AwxPageForm } from '../../common/AwxPageForm';
 import { awxAPI } from '../../common/api/awx-utils';
 import { PageFormFileUpload } from '../../../../framework/PageForm/Inputs/PageFormFileUpload';
+import { useTranslation } from 'react-i18next';
 
 export interface AwxSettingsOptionsResponse {
   actions: {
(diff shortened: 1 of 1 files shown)


frontend/awx/administration/settings/AwxSettingsForm.tsx
```diff
@@ -168,6 +169,38 @@ export function AwxSettingsForm(props: {
 
 export function OptionActionsFormInput(props: { name: string; option: AwxSettingsOptionsAction }) {
   const option = props.option;
+
+  if (option.label.endsWith('Secret')) {
```
nixocio, line 173: I read this code quickly @prat98 would this label being translated?

prat98: Hmm good question maybe I should just do `.contains(t('Secret'))`. That way I capture the word in the label regardless. Certain translations may update the order of the words.

nixocio: I do not remember seeing such thing before. In general we might need to create an object or other data structure with an id that is not selected to be translated.

AlexSCorey (approved):
We spoke offline.  Looks good to me.!

frontend/awx/administration/settings/AwxSettingsForm.tsx
prat98: Hmm so what is your proposal?

PR #2477, opened 2024-06-06
[AAP-25091] Uses resource id for edit schedule instead of params

AlexSCorey:
This resolves AAP-25091.  Instead of relying on the params to get the resource ID we get that information from the form.


skochay330 (approved):
Works perfectly, thank you!

PR #2494, opened 2024-06-10
GitHub Action Update Success Job

jamestalton:
We need to cut our Cypress costs.
Part of that is to stop using the Cypress dashboard for component tests.
This will save ~75% in costs.
There is a new job  as part of the pipeline that indicates if that pipeline was success.
This makes it easier to have that as the only required check for PRs.
Since the team has decided that E2E are not required, this removes the requirement of E2E from the success job.


nixocio (approved):
@jamestalton explained the why there. Ty

PR #2499, opened 2024-06-10
[AAP-24169, AAP-24170, AAP-24171] Removes unused file and lots of TODOS

AlexSCorey:
This addresses AAP-24169, AAP-24170, AAP-24171 which are all TODOS that are in a file called `launchHandlers.tsx`.  This file is no longer being used and I deleted the file.


vidyanambiar (approved):

nixocio:

PR #2500, opened 2024-06-10
Use GET action to generate content type options in AWX role wizard instead of POST

mabashian:
Normal users won't have POST in the OPTIONS response but they will have GET


vidyanambiar (approved):
lgtm!

nixocio:

PR #2501, opened 2024-06-10
Allow write-in values for async single selects

keithjgrant:
Adds a new `writeInOption` prop for async single select. If given, this callback can create an on-the-fly select option based on the user's search query


vidyanambiar (approved):
Looks good!  
Do we have a use case for this or is it being added as nice to have?

PR #2503, opened 2024-06-11
inventories run command test for constructed inventories

MilanPospisil:


cypress/e2e/awx/resources/inventoryHost/runCommandFunction.ts
```diff
@@ -0,0 +1,98 @@
...
+        cy.getByDataCy('checkbox-column-cell').click();
+
+        cy.contains('button', 'Confirm').click();
+
+        cy.getByDataCy('Submit').click();
+
+        // Credentials tab
+
+        // not sure why this is needed, but it ensures the test stability
+        cy.wait(4000);
```
MilanPospisil, line 63: There may be some issue with the tabs in framework, or with selection component, but without some waiting after moving to the new tab, this tests fails constantly in the selector in CI, but not localy.

ZitaNemeckova (approved):

PR #2504, opened 2024-06-11
Fix multiple lookup-button elements

ZitaNemeckova:
Use cases:
- For Job template when there's more than 50 items for inventories and projects there's gonna be two `cy.get('svg[data-cy="lookup-button"]')` elements -> it's not possible to click on correct one
- After clicking  `cy.get('svg[data-cy="lookup-button"]')` we need to wait for the table to load before we look for anything


nixocio:

nixocio (approved):
We got in a call and we test this change as a group.

kialam:
Ran the following tests locally against the test server and they pass:
workflowVisualizerCRUD
executionEnvironments
applications
inventoryGroup

PR #2514, opened 2024-06-12
Add assertion to singleSelectBy to ensure it is visible before perfor…

kialam:
…ming an action.


akus062381 (approved):

jamestalton:
```
cy.contains('.pf-v5-c-menu__item-text', exactMatch ? regExp : value)
```

Will already wait until a component that has a class of .pf-v5-c-menu__item-text containing the value is found before trying to act on it.
Adding the .should('be.visible') is not fixing that a dropdown has not loaded.

kialam:
Closing as the added assertion is unnecessary.

PR #2517, opened 2024-06-12
Removes deprecated hook and replaces it with the current hook

AlexSCorey:
This removes a deprecated usePagDialog hook and replaces it with usePageDialogs.  no functionality should change.  

In an effort to clean up our code base and remove deprecated items in order to improve our QOL and improve our testing issues I think this PR is pertinent.


h-kataria:
@AlexSCorey do we need tests to verify changes in this PR?

AlexSCorey:
@h-kataria Since this does not bring in any functionality changes from a UI perspective I believe that current tests should be sufficient

PR #2519, opened 2024-06-13
Fix notifiers tests

MilanPospisil:
Fix assert that item is deleted.
Filter for item now waits for single item row (and header row).

Also wait for the reload more longer.

### DIFF
--- a/cypress/e2e/awx/administration/notifiers/notifiersTabs.cy.ts
+++ b/cypress/e2e/awx/administration/notifiers/notifiersTabs.cy.ts
@@ -241,12 +241,10 @@ function testToggle(
     cy.get(`[aria-label="${type_enable}"]`).click();
 
     // reload page to check if the toggle is working and try to disable it
-    cy.reload();
-    cy.get(`[aria-label="${type_disable}"]`, { timeout: 25000 }).click();
+    cy.get(`[aria-label="${type_disable}"]`, { timeout: 5000 }).click();
 
     // check if it is disabled again
-    cy.reload();
-    cy.get(`[aria-label="${type_enable}"]`, { timeout: 25000 });
+    cy.get(`[aria-label="${type_enable}"]`, { timeout: 5000 });
   });
 }
 
--- a/cypress/e2e/awx/administration/notifiers/notifiersTabs.cy.ts
+++ b/cypress/e2e/awx/administration/notifiers/notifiersTabs.cy.ts
@@ -17,7 +17,7 @@ describe('Notifications', () => {
     });*/
   });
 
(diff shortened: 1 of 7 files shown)


himdel (approved):
Sounds reasonable given the contraints :) :+1:

cypress/e2e/awx/administration/notifiers/notifiersTabs.cy.ts
```diff
@@ -242,11 +242,11 @@ function testToggle(
 
     // reload page to check if the toggle is working and try to disable it
     cy.reload();
-    cy.get(`[aria-label="${type_disable}"]`, { timeout: 15000 }).click();
+    cy.get(`[aria-label="${type_disable}"]`, { timeout: 25000 }).click();
```
vishsanghishetty, line 245: I think we can do this without using reload
see if this code is useful which I have written for Management Job Schedules and Notification tabs
```
cy.get('.pf-v5-c-switch__label.pf-m-on')
                  .should('have.text', 'Schedule enabled')
                  .should('be.visible');
```

                // Toggle the schedule to disabled
                cy.getByDataCy('toggle-switch').click();

                // Assert the schedule is disabled
                cy.get('.pf-v5-c-switch__label.pf-m-off')
                  .should('have.text', 'Schedule disabled')
                  .should('be.visible');`

vishsanghishetty: or this
`//enable and disable start, success and failure switches of the notifier
              cy.toggleAndAssert(['start', 'success', 'failure']);`

cypress/e2e/awx/administration/notifiers/notifiersTabs.cy.ts
```diff
@@ -242,11 +242,11 @@ function testToggle(
...
 
     // check if it is disabled again
     cy.reload();
-    cy.get(`[aria-label="${type_enable}"]`, { timeout: 15000 });
+    cy.get(`[aria-label="${type_enable}"]`, { timeout: 25000 });
```
vishsanghishetty, line 249: same as above

vishsanghishetty (commented):
I am going to review further and leave comments if any. thanks for the fixes

cypress/e2e/awx/administration/notifiers/notifiersTabs.cy.ts
MilanPospisil: Im not sure if this will not lead to problems, because we will immidiately call switch on and switch off and that could be processed in parallel in backend and thus leading to problems. But ok, lets try it without reload.

PR #2525, opened 2024-06-14
flaky tests -> execution environments 

acruzgon:
This PR is:

- Removing the use of `Global Organization` to create Execution Environments
- Addendum to Instance Group User access Tab test to also stop using Global organization.
- Skipping additional `Teams` tests


cypress/e2e/awx/resources/executionEnvironments.cy.ts
```diff
@@ -265,15 +271,15 @@ describe.skip('Execution Environments: Edit and Bulk delete', () => {
     cy.verifyPageTitle(executionEnvironment.name + '-edited');
     cy.hasDetail('Name', executionEnvironment.name + '-edited');
     cy.hasDetail('Image', 'executionenvimage');
-    cy.hasDetail('Organization', 'Global Organization');
+    cy.hasDetail('Organization', organization.name);
   });
 
   it('can bulk delete multiple EEs from the list view and assert deletion', function () {
```
nixocio, line 277: @acruzgon do we still need to use function here? Since we are no longer `this`. Perhaps we could update just to be consistent everywhere that is not longer necessary.

acruzgon: @nixocio agree! it's not needed anymore so i'll remove it.

nixocio (approved):

PR #2529, opened 2024-06-14
Fixes flake in credentials and some job e2e tests

AlexSCorey:
One goal for this pr, within the jobs tests, is to rely less on the api to finish a job before we can act on it...delete it.  To do this I moved some tests out of the e2e test file and into a component test file and to focus less on what the api returns to us, and more on what the user can do.  In those tests we assert that a user can delete a job.  This is an improvement IMO because we no longer have to wait for the job to finish.  We can use stubbed data to load the table with completed jobs, and we can assert that the user can go through the steps necessary to delete a job.

### DIFF
--- a/cypress/e2e/awx/resources/credentials.cy.ts
+++ b/cypress/e2e/awx/resources/credentials.cy.ts
@@ -11,7 +11,6 @@ import { randomE2Ename } from '../../../support/utils';
 
 describe('Credentials', () => {
   let organization: Organization;
-  let credential: Credential;
   let user: AwxUser;
 
   before(() => {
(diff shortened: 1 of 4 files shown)


cypress/e2e/awx/resources/credentials.cy.ts
```diff
@@ -579,31 +593,31 @@ describe('Credentials Tabbed View - Job Templates', function () {
...
-          machineCredential = cred;
-        });
-      });
-      cy.createAwxTeam(awxOrganization).then((awxTeam) => {
-        createdAwxTeam = awxTeam;
-      });
-      cy.createAwxInventory().then((inv) => {
-        awxInventory = inv;
+    awxOrganization = this.globalOrganization as Organization;
+    // cy.createAwxOrganization().then((awxOrg) => {
```
acruzgon, line 597:
```suggestion
```

cypress/e2e/awx/resources/credentials.cy.ts
```diff
@@ -579,31 +593,31 @@ describe('Credentials Tabbed View - Job Templates', function () {
...
+    // awxOrganization = awxOrg;
```
acruzgon, line 598:
```suggestion
```

cypress/e2e/awx/resources/credentials.cy.ts
```diff
@@ -579,31 +593,31 @@ describe('Credentials Tabbed View - Job Templates', function () {
...
+        machineCredential = cred;
       });
     });
+    cy.createAwxTeam(awxOrganization).then((awxTeam) => {
+      createdAwxTeam = awxTeam;
+    });
+    cy.createAwxInventory().then((inv) => {
+      awxInventory = inv;
+    });
+    // });
```
acruzgon, line 615:
```suggestion
```

cypress/e2e/awx/resources/credentials.cy.ts
```diff
@@ -623,40 +637,10 @@ describe('Credentials Tabbed View - Job Templates', function () {
...
-        });
-        cy.contains(/^Success$/);
-        cy.clickButton(/^Close$/);
-        cy.clearAllFilters();
-      });
   });
 });
 
-describe.skip('Credentials Tabbed View - Team and User Access', function () {
+describe('Credentials Tabbed View - Team and User Access', function () {
```
acruzgon, line 643: we can remove `function` from here since it's not used

cypress/e2e/awx/views/jobs.cy.ts
```diff
@@ -110,104 +109,6 @@ describe('Jobs: List', () => {
   });
 });
 
-describe('Jobs: Delete', () => {
```
acruzgon, line 113: what's the reason to remove delete tests for jobs? i think we could revamp these tests instead of removing them

acruzgon (changes requested):

cypress/e2e/awx/views/jobs.cy.ts
AlexSCorey: I've removed them from this file and put them in `Jobs.cy.tsx`

acruzgon (approved):
following the best practice to avoid creating unused resources, the next two tests are not using the `credential` created in that describe block:

- `can display error toast message when running a test from the create credential form`
- `can display success toast message when running a test from the create credential form`

i'll leave it up to you. other than that it looks good!

PR #2535, opened 2024-06-17
Enable kebab dropdown keyboard accessibility

marshmalien:
https://issues.redhat.com/browse/AAP-12914
https://issues.redhat.com/browse/AAP-12952

Update to PatternFly's latest implementation of [dropdown](https://www.patternfly.org/components/menus/dropdown) and remove deprecated implementation. Some breaking changes addressed:
* Pass dropdown `position` as Dropdown `popperProps`
* Remove `<Tooltip>` from dropdown items and pass tooltip properties as DropdownItem `tooltipProps `
* Pass dropdown item icon as DropdownItem `children`
* Replace `<DropdownSeparator>` with `<Divider>`
* Replace linkable dropdown item `component` prop with `to`

keyboard

### DIFF
--- a/framework/PageActions/PageActionDropdown.tsx
+++ b/framework/PageActions/PageActionDropdown.tsx
@@ -1,16 +1,19 @@
-import { ButtonVariant, Tooltip } from '@patternfly/react-core';
 import {
+  Button,
+  ButtonVariant,
+  Divider,
   Dropdown,
   DropdownItem,
-  DropdownPosition,
-  DropdownSeparator,
-  DropdownToggle,
-  KebabToggle,
-} from '@patternfly/react-core/deprecated';
-import { CircleIcon } from '@patternfly/react-icons';
+  DropdownList,
+  DropdownPopperProps,
+  Icon,
+  MenuToggle,
+  MenuToggleElement,
+  Tooltip,
+} from '@patternfly/react-core';
+import { CircleIcon, EllipsisVIcon, CaretDownIcon } from '@patternfly/react-icons';
 import { ComponentClass, FunctionComponent, useEffect, useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
-import { Link } from 'react-router-dom';
 import styled from 'styled-components';
 import { PFColorE, getPatternflyColor } from '../components/pfcolors';
 import { getID } from '../hooks/useID';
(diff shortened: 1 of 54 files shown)


framework/PageActions/PageActionDropdown.tsx
```diff
@@ -165,9 +118,64 @@ export function PageActionDropdown<T extends object>(props: PageActionDropdownPr
...
+    );
+
   return (
     <Tooltip content={tooltipContent} trigger={tooltipContent ? undefined : 'manual'}>
-      {dropdown}
+      <Dropdown
+        isOpen={dropdownOpen}
+        onSelect={() => setDropdownOpen(false)}
+        onOpenChange={(isOpen) => setDropdownOpen(isOpen)}
+        popperProps={{
```
jamestalton, line 136: When the dropdown opens near the bottom of a table, it causes the whole page to get a scroll bar.
The fix is to mount the dropdown at the document.body.
We can also leverage the preventOverflow in that case to auto position the dropdown.
Then we can remove the position prop from the component.
```
        popperProps={{
          appendTo: () => document.body,
          preventOverflow: true,
          enableFlip: true,
        }}
```
This may affect tests, but fixes the problem with the overflow scrolling!

framework/PageActions/PageActionDropdown.tsx
```diff
@@ -259,13 +282,25 @@ function PageDropdownActionItem<T extends object>(props: {
         <Tooltip key={action.label} content={tooltip} trigger={tooltip ? undefined : 'manual'}>
           <DropdownItem
             isAriaDisabled={Boolean(isDisabled)}
-            icon={Icon ? <Icon /> : undefined}
-            component={<Link to={to}>{action.label}</Link>}
+            to={to}
```
jamestalton, line 285: Action type button has
```
data-cy={getID(action)?.split('.').join('-')}
```
Should action type lionk have the same?

cypress/e2e/awx/administration/instances.cy.ts
```diff
@@ -173,12 +173,12 @@ describe('Instances: Delete', () => {
   it('can remove an instance from instance list toolbar', () => {
     cy.intercept('PATCH', '/api/v2/instances/*').as('removedInstance');
     cy.get('[data-cy="actions-dropdown"]').click();
-    cy.get('[data-cy="remove-instance"]').should('have.attr', 'aria-disabled', 'true');
+    cy.get('[data-cy="remove-instance"] button').should('have.attr', 'aria-disabled', 'true');
```
jamestalton, line 176: Why is button added here?
It may be needed, but I worry that it will make it harder on devs to figure out when they need something like this.
Any way to make it so that this is not needed?

jamestalton (changes requested):
Awesome changes. I'd also like to see us fix the overflow scrolling in this!

jamestalton (dismissed):

nixocio (approved):

nixocio (approved):

keithjgrant (approved):

PR #2537, opened 2024-06-18
[AAP-23092] Make verbosity levels dropdown

prat98:
This PR is for AAP-23092 it replaces a text input for verbosity levels with a proper dropdown


AlexSCorey (approved):
This looks good to me.

PR #2538, opened 2024-06-18
Update cypress login and token usage [AAP-25472]

jamestalton:
- New cy.login that handles login for all projects
- cy.login is not needed in tests anymore as it is called before and beforeEach
- This is performant because of the cypress session support
- Removed the use of the AWX token in AWX custom commands
- This was causing issues in downstream testing
- This is not needed as it was only used for globalOrg and EDA activations, both which will work now without the AWX token.
- Moved the creation of globalOrg and globalProject into a location that happens after login.
- Renamed globalOrganization to globalAwxOrganization to make it clearer.


akus062381 (approved):
Exciting changes, thank you

vishsanghishetty (approved):

nixocio (approved):
💯

PR #2544, opened 2024-06-19
Fix topology view tests

MilanPospisil:
Flaky test unskipped, but I cant reproduce the failures after many retries...

### DIFF
--- a/cypress/e2e/awx/administration/topology-view.cy.ts
+++ b/cypress/e2e/awx/administration/topology-view.cy.ts
@@ -11,7 +11,6 @@ describe('Topology view', () => {
   let user: AwxUser;
   let organization: Organization;
 
-
   before(() => {
     cy.login();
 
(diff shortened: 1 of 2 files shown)


ZitaNemeckova:
Gonna re-run it 3 times here and if it's always green I'll approve.

cypress/e2e/awx/administration/topology-view.cy.ts
```diff
@@ -82,8 +82,8 @@ describe('Topology view', () => {
       });
     });
   });
-  // FLAKY_06_13_2024
```
nixocio, line 85: Would we able to find the root of the flakiness?

MilanPospisil: Ah, I see why it is probably running well now. It is using global org, some test may have done something with it and now when we are moving away from global, it starts to work.

I will remove global org here too!

ZitaNemeckova (approved):

PR #2547, opened 2024-06-19
[AAP-25642] Fix breadcrumbs params in `Inventory > Inventory Source > Schedule` details 

jerabekjiri:
Issue: [AAP-25642](https://issues.redhat.com/browse/AAP-25642)

### DIFF
--- a/frontend/awx/main/routes/useAwxInventoryRoutes.tsx
+++ b/frontend/awx/main/routes/useAwxInventoryRoutes.tsx
@@ -63,9 +63,13 @@ export function useAwxInventoryRoutes() {
             <SchedulePage
               initialBreadCrumbs={[
                 { label: t('Inventories'), to: AwxRoute.Inventories },
-                { id: 'inventory', to: AwxRoute.InventoryPage },
-                { label: t('Inventory Sources'), to: AwxRoute.InventorySources },
-                { id: 'data', to: AwxRoute.InventorySourcePage },
+                { id: 'inventory', to: AwxRoute.InventoryDetails },
+                {
+                  id: 'inventory_sources',
+                  label: t('Inventory Sources'),
+                  to: AwxRoute.InventorySources,
+                },
+                { id: 'data', to: AwxRoute.InventorySourceDetail },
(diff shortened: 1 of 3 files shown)


AlexSCorey (commented):
I _think_ I understand the goal here....to add a link in the breadcrumb for the inventory source when on the schedules list inside an inventory source.  However, currently if you are on the schedules list inside a resource, we don't use a breadcrumb link for the details view of that resource in other parts of the app.  This is because the details view is another tab in line with the details tab.  See job templates schedules list as an example below.

Screenshot 2024-06-20 at 12 17 08 PM


Below is an the schedules list inside an inventory source, on this branch.  As you can see the breadcrumb links to the inventory source called `gergree`, but just clicking on the detail tab takes you to the same place as that link would.  If we are going to make this change we should be consistent throughout the app.


Screenshot 2024-06-20 at 12 18 57 PM


Regarding the screenshot on the issue.  We can hard code the `:inventory_type` param value because only regular inventories have inventory sources....and only inventory sources have schedules.

jerabekjiri:
@AlexSCorey 
I see your point and it makes sense :+1:, so tldr `gergree` should not be link.

plus I also noticed that breadcrumbs are little different in `schedules list` vs `schedule detail` in the inventory source.

Schedule list:
Screenshot from 2024-06-21 13-00-46

Schedule detail:
Screenshot from 2024-06-21 13-00-30

I'm wondering which breadcrumbs structure should we follow to keep it consistent?

tiyiprh:
> @AlexSCorey I see your point and it makes sense 👍, so tldr `gergree` should not be link.
> 
> plus I also noticed that breadcrumbs are little different in `schedules list` vs `schedule detail` in the inventory source.
> 
> Schedule list: Screenshot from 2024-06-21 13-00-46
> 
> Schedule detail: Screenshot from 2024-06-21 13-00-30
> 
> I'm wondering which breadcrumbs structure should we follow to keep it consistent? CC @tiyiprh @aratti96

I think we should follow the pattern in the second image. There was some discussion on this issue and in a PR Vidya worked on before where we decided on the pattern: https://issues.redhat.com/browse/AAP-9170

PR #2548, opened 2024-06-19
flaky tests -> teams

acruzgon:
This PR is:

- Adding two additional `describe` blocks in order to separate tests that are not using resources created during `beforeEach`
- Replacing unnecessary `filterTableByMultiSelect` with `filterTableBySingleSelect`
- Removing test `can render the team details page` because it's a component tests instead of an integration test
- Removing test `can edit a team from the details page` because it's a component tests instead of an integration test
- Removing unnecessary `function()`
- Adding new test for bulk deletion


vishsanghishetty (approved):
🚀

PR #2551, opened 2024-06-20
[AAP-25659] Show error if resources are missing in Schedule wizard

jerabekjiri:
Issue: [AAP-25659](https://issues.redhat.com/browse/AAP-25659)

Screenshot from 2024-06-26 14-53-39

### Don't let create schedule if resources are missing
**empty state:**
Screenshot from 2024-06-26 14-53-17
~~(custom icon requires https://github.com/ansible/ansible-ui/pull/2585)~~

**disabled button:**
Screenshot from 2024-06-24 16-44-53

### DIFF
--- a/cypress/e2e/awx/views/schedules.cy.ts
+++ b/cypress/e2e/awx/views/schedules.cy.ts
@@ -94,7 +94,7 @@ describe.skip('Schedules - Create and Delete', () => {
     });
   });
 
-  describe.skip('Schedules - Create schedule of resource type Project', () => {
+  describe('Schedules - Create schedule of resource type Project', () => {
     let organization: Organization;
     let project: Project;
 
--- a/cypress/e2e/awx/views/schedules.cy.ts
+++ b/cypress/e2e/awx/views/schedules.cy.ts
@@ -74,9 +74,27 @@ describe.skip('Schedules - Create and Delete', () => {
       cy.clickModalButton('Delete schedule');
       cy.deleteAwxJobTemplate(jobTemplate, { failOnStatusCode: false });
     });
+
+    it("can't create a schedule with missing resources", () => {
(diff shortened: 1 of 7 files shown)


marshmalien (changes requested):
We don't need to show an error when a workflow job template is missing an inventory. Only when the launch config `inventory_needed_to_start` is true AND the schedule is missing an inventory. 


https://github.com/ansible/awx/blob/devel/awx/ui/src/components/Schedule/shared/ScheduleForm.js#L246-L260

```const missingRequiredInventory = useCallback(() => {
  let missingInventory = false;
  if (
    launchConfig?.inventory_needed_to_start &&
    !schedule?.summary_fields?.inventory?.id
  ) {
    missingInventory = true;
  }
  return missingInventory;
}, [launchConfig, schedule]);

marshmalien (approved):
looks great to me!

AlexSCorey (changes requested):
It seems that we have a difference in error handling for editing a resource with missing dependencies, and creating a schedule on a resource with missing dependencies.

The first image is for editing a schedule that is part of job template `hj`.  
Screenshot 2024-07-03 at 11 36 55 AM

The second image is for creating a schedule that is part of job template `hj` even though the resource name isn't correct (that looks like another bug).  
Screenshot 2024-07-03 at 11 38 02 AM

The methods of error handling should be the same in both cases. The in edit scenario we do not show the error message, and we should.

vidyanambiar (approved):
lgtm!

AlexSCorey (approved):
All the issues I have found are resolved.  Looks good to me.  Well done!

AlexSCorey:
I file https://issues.redhat.com/browse/AAP-26667 in response to the bug I found about the name of the resource on the review step.

PR #2556, opened 2024-06-20
Skip failing WFJT Surveys Tests

akus062381:


cypress/e2e/awx/resources/workflowJobTemplateSurveys.cy.ts
```diff
@@ -12,7 +12,7 @@ describe('Workflow Job Templates Surveys', function () {
     cy.awxLogin();
   });
 
-  describe('WFJT Surveys: Create, Edit and Delete', function () {
```
nixocio, line 15: For the previous skipped ones we add a flag such ```// FLAKY_06_13_2024``` to allow to search, and identify when that flaky was introduced.

akus062381: Oh great, I'll add that, thank you

PR #2577, opened 2024-06-24
Fixes selectors for flaky test credentials test

vishsanghishetty:
Fixes selectors for flaky test credentials test
`edit credential type that renders a sub form`

### DIFF
--- a/cypress/e2e/awx/resources/credentials.cy.ts
+++ b/cypress/e2e/awx/resources/credentials.cy.ts
@@ -533,7 +533,7 @@ describe('Create Credentials of different types', () => {
     cy.verifyPageTitle(credentialName);
     cy.getByDataCy('name').contains(credentialName);
     cy.contains('Private Key Passphrase').should('be.visible');
-    cy.get('private-key-passphrase').contains('Prompt on launch');
+    cy.getByDataCy('private-key-passphrase').contains('Prompt on launch');
     cy.contains('Password').should('be.visible');
     cy.getByDataCy('password').contains('Prompt on launch');
     cy.getByDataCy('edit-credential').click();
--- a/cypress/e2e/awx/resources/credentials.cy.ts
+++ b/cypress/e2e/awx/resources/credentials.cy.ts
@@ -479,15 +479,12 @@ describe('Create Credentials of different types', () => {
         cy.verifyPageTitle('Edit Credential');
         const ModifiedCredentialName = credentialName + ' - edited';
         cy.getByDataCy('name').type(ModifiedCredentialName);
-        cy.getByDataCy('username').should('be.visible');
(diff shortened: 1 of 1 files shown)


vidyanambiar:
I pushed up a commit to address the following failure:
Screenshot 2024-06-20 at 10 29 07 AM

The issue was that the function `removeRoleFromListRow` was using the team/user name to delete the role assignment from the view. The team/user name is the same for both role assignments. If the UI did not refresh in time to remove the deleted assignment, the same assignment got selected for removal a second time resulting in a 404 response. 
The change here was to use the distinct role name to delete the role assignment to ensure that we are deleting distinct assignments.

vishsanghishetty:
> I pushed up a commit to address the following failure: https://cloud.cypress.io/projects/77sud6/runs/7115/overview/c383195c-3e6d-4394-907c-b0240fc85a9e/replay?roarHideRunsWithDiffGroupsAndTags=1&ts=1719408209785.5&att=1 image (33)
> 
> The issue was that the function `removeRoleFromListRow` was using the team/user name to delete the role assignment from the view. The team/user name is the same for both role assignments. If the UI did not refresh in time to remove the deleted assignment, the same assignment got selected for removal a second time resulting in a 404 response. The change here was to use the distinct role name to delete the role assignment to ensure that we are deleting distinct assignments.

@vidyanambiar thanks for the fix Vidya!

cypress/e2e/awx/resources/credentials.cy.ts
```diff
@@ -354,21 +354,27 @@ describe('Create Credentials of different types', () => {
             (item.required as { field: string; dataCy: string }[]).forEach((credentialType) => {
               switch (item.name) {
                 case 'HashiCorp Vault Secret Lookup':
-                  cy.get(`[data-cy="${credentialType.dataCy}"]`).type(`${credentialType.field}`);
+                  cy.getByDataCy(`[data-cy="${credentialType.dataCy}"]`).type(
```
nixocio, line 357: hey @vishsanghishetty I believe we decide to remove this specific test in another PR, but the usage of `getByDataCy` would not require to pass `data-cy`again? Just the name of the locator?

See: https://github.com/ansible/ansible-ui/pull/2601

nixocio (approved):

cypress/e2e/awx/resources/credentials.cy.ts
vishsanghishetty: @nixocio Yes, I was the one who wrote the creation of 30 different credentials test. I see that it is skipped, fine with me. since this PR is updated with the new changes, it will skip this test. I might have changed this locator when I made bulk changes in the file. hope that answers your question.

vidyanambiar (approved):

PR #2578, opened 2024-06-24
Consolidate credentials sub form tests

nixocio:
Consolidate credentials sub form tests, also add data-cy to PageDetail


marshmalien (approved):
Looks good to me. We talked about consolidating the visible and value assertions using one of our existing commands `getByDataCy` or `containsByDataCy`.

vishsanghishetty (approved):

nixocio:
@marshmalien I will address that in a follow up PR.

PR #2583, opened 2024-06-25
Add cancel job test

ZitaNemeckova:
Last part of https://issues.redhat.com/browse/AAP-22952

I tried every single playbook and none of them run more then 30s. So I decided to mock the API response for running job to prevent flakiness. In the future I'd like to create a playbook that just waits/sleeps for like five minutes to use here.


MilanPospisil (approved):
LGTM :+1:

PR #2589, opened 2024-06-25
Remove unused cypress helpers

himdel:
Clean up Cypress helpers that are not currently being used.

```
sg -p 'Cypress.Commands.add($A, $$$)' -r '$A' | grep '│+' | cut -d\' -f2 | sort > list.all
cat list.all | while read name; do echo -n "$name        "; rg  'cy.'"$name"\\b cypress frontend framework | wc -l; done | sort -k2 -n > list.uses
grep ' 0$' list.uses | col1 > list0
cat list0 | while read name; do sg -p 'Cypress.Commands.add('\'"$name"\'', $$$)' -r '' -U; git commit -a -m 'remove unused helper '"$name" ;  done
npm run prettier:fix; git commit -a -m 'prettier:fix'
```

(and manual eslint fixes for unused imports)
(and run again because some helpers were only used by removed helpers, while `list0` is nonempty :))

NOTE this does NOT remove helpers that are only used in skipped tests, the assumption is that those tests are more likely to get unskipped rather than deleted

(artifacts in https://gist.github.com/himdel/423046c1a29a6d225575558bafac989b)

### DIFF
--- a/cypress/e2e/hub/execution-environments.cy.ts
+++ b/cypress/e2e/hub/execution-environments.cy.ts
@@ -28,7 +28,7 @@ describe('Execution Environments', () => {
     });
   });
 
-  it('can add and delete a new execution environment', () => {
+  it.skip('can add and delete a new execution environment', () => {
     cy.createHubRemoteRegistry().then((remoteRegistry) => {
       const eeName = `execution_environment_${randomString(3, undefined, { isLowercase: true })}`;
       const upstreamName = `upstream_name_${randomString(3, undefined, { isLowercase: true })}`;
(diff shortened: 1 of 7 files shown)


AlexSCorey (approved):
As long as tests pass I'm happy to merge this.

ZitaNemeckova (approved):

cypress/e2e/hub/execution-environments.cy.ts
```diff
@@ -38,6 +38,8 @@ describe('Execution Environments', () => {
       cy.intercept('GET', hubAPI`/_ui/v1/execution-environments/registries/?limit=50`).as(
         'registries'
       );
+      // FIXME: this is sometimes add-.., sometimes create-..., depending on whether there are already any added
```
himdel, line 41: (addressing this bit properly in https://github.com/ansible/ansible-ui/pull/2636)

PR #2591, opened 2024-06-25
Quick fix run ad hoc command

MilanPospisil:


cypress/e2e/awx/resources/inventoryHost/runCommandFunction.ts
```diff
@@ -60,7 +60,7 @@ export function runCommand(params: {
         // Credentials tab
 
         // not sure why this is needed, but it ensures the test stability
-        //cy.wait(4000);
+        cy.wait(4000);
```
nixocio, line 63: Could we just bump the timeout for certain command?  Or get certain element that should be present on the screen and add a custom timeout as well?

MilanPospisil: I will try to wait for whole page elements to be loaded and we will see...

nixocio (approved):

PR #2604, opened 2024-06-26
increase debounce time on multiselect, text filter

keithjgrant:
testing the impact of debounce on e2e tests


keithjgrant:
I can't see any evidence this is impacting e2e tests.

I do think it’s a worthwhile change for general user performance, but I’m not going to push to merge this just prior to release.

PR #2612, opened 2024-06-27
Disable adding role assignments to managed and global EEs

vidyanambiar:
[AAP-25268](https://issues.redhat.com/browse/AAP-25268)

System managed and globally available EEs are superuser-only and are not applicable for object-level role assignments.

#### Object assignment flow
`Execution Environment -> User Access -> Add Roles` and `Execution Environment -> Team Access -> Add Roles`
- For these types of EEs, the `Team Access` and `User Access` tabs must be hidden in the details view.
- If the user tried to access those tabs using the direct URL, they will be redirected to the details tab.

#### User assignment flow
`User -> Roles -> Add Roles`

- In the Add Roles wizard, if the Resource Type is selected as Execution Environment, the next step for selection of resources will display a filtered list of EEs (filtering out any managed/global EEs)

#### Team assignment flow
`Team -> Roles -> Add Roles`

- In the Add Roles wizard, if the Resource Type is selected as Execution Environment, the next step for selection of resources will display a filtered list of EEs (filtering out any managed/global EEs)


keithjgrant (approved):

akus062381 (approved):
Testing manually and with automation, all looks good :-)

PR #2617, opened 2024-06-28
Removal of Header Description Tooltip When Full Description is Shown

stevensonmichel:
**Description**
The description of the header in every page of AWX has an associated tooltip that shows up when the cursor hovers on the Description.

**Issue**
Since the description of the header is always present, the tooltip becomes unnecessary. 

**Solution**
As a solution, the tooltip is completely removed, which leaves only the description in the header. Also, when the page shrinks less than medium size, the description disappears. Under such circumstances, the user can click on the info icon to inquire for more details about the respective component. Thus, the tooltip becomes unnecessary in all cases.

**Files Changed**
- `PageHeader.tsx` located in the `framework` directory.

Fixes issue [AAP-9069]


stevensonmichel:
> @stevensonmichel I'm not sure I follow what this PR changes. Do you have an issue you can link and/or before and after screenshots?

@AlexSCorey I added the issue in the PR description.

himdel:
Looks like previously, we would only ever see a one-line help text ending in an ellipsis when too long, with a Tooltip when hovering.

After, we see the full text, wrapped, with no truncation or Tooltips.

Both before & after, when the screen is too narrow, there is no text at all.
On screens using the `(?)` button, the text can be seen in a Popover 

---

Before:



After: 



---

@stevensonmichel Assuming that wrapping long text was intentional, this is looking good :+1: 
But .. we should probably ensure that the question mark button is actually part of all the titles .. it's definitely missing from multiple Hub screens, so maybe we need to do a sweep to add it everywhere?

stevensonmichel:
The wrapping was indeed intentional, from a discussion I had with Tina. I am gonna make sure the info button wherever it is necessary.

himdel:
Looks like the missing help buttons are being addressed in https://github.com/ansible/ansible-ui/pull/2782 :).
So I guess this one just needs green tests now :)

himdel (approved):
LGTM :) :+1: 

The sonarcloud "no duplicate props" failures seem to be incorrect, looks like the duplication is between lines removed and lines added, and not present in code.

PR #2621, opened 2024-07-01
Refactor WorkflowOutputNavigation

ZitaNemeckova:
Instead of hardcoded strings use Routes and params

Before/After:
Works the same on the upstream.

Related to https://issues.redhat.com/browse/AAP-25915


MilanPospisil (approved):
LGTM :+1:

PR #2622, opened 2024-07-01
Fix flaky instances test

jamestalton:
The test was navigating to 'instances'
```
    cy.wait('@newInstance')
      .its('response.body')
      .then((thisInstance: Instance) => {
        cy.navigateTo('awx', 'instances');
```

But in the UI when a user clicks submit, after the POST, the UI navigates to the created instance details page.

So sometimes the test ends up on "Instances" and other times ends up on the instance details.
This causes the test to fail.


acruzgon (approved):
lgtm! ran locally against the CI
image

nixocio (approved):

PR #2625, opened 2024-07-01
AAP-9048: Update login page to use branded logos

jamestalton:
https://issues.redhat.com/browse/AAP-9048

Branding really want the UI to use the official logos in login.
This work is needed for downstream to properly handle the AAP branding, custom branding, and custom description.

Uses: https://www.patternfly.org/components/login-page/

Screenshot 2024-07-02 at 1 01 13 PM

- Updates the Login Page to use the PatternFly components which fix dark theme and layout issues.

Fixes current layout issues seem below
Screenshot 2024-07-02 at 10 04 36 AM


frontend/common/SocialAuthLogin.tsx
```diff
@@ -46,14 +27,14 @@ export function SocialAuthLogin(props: SocialAuthLoginProps) {
   }
 
   return (
-    <Container>
-      <Heading>{t`Log in with`}</Heading>
-      <Grid>
+    <>
+      <Title headingLevel="h3">{t`Log in with`}</Title>
```
nixocio, line 31: From the PF example they do not add this section. Do we need this one?

frontend/awx/main/AwxLogin.tsx
```diff
@@ -37,16 +36,15 @@ export function AwxLogin(props: { children: React.ReactNode }) {
 
   if (!activeAwxUser) {
     return (
-      <Login
+      <AnsibleLogin
         authOptions={authOptions}
-        apiUrl="/api/login/"
+        loginApiUrl="/api/login/"
```
nixocio, line 41: Interesting that the other projects need to have the functions to wrap those URLS.

frontend/common/AnsibleLogin.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  const { loginApiUrl } = props;
+  const onSubmit = useCallback(async () => {
+    try {
+      const loginPageResponse = await fetch(loginApiUrl, {
+        credentials: 'include',
+        headers: { Accept: 'application/json,text/*' },
+      });
+      if (!loginPageResponse.ok) {
+        throw await createRequestError(loginPageResponse);
```
nixocio, line 62: Previously, we were using `LoginForm` and `GenericForm`, and any errors raised would bubble up to these components. `GenericForm` would then parse the raised errors and display them to the user. Now, with the use of PatternFly (PF) native components, I'm not sure if raising this error here will still result in it being displayed to the user. I believe we are relying on the ErrorBoundary now.
 @jamestalton, what do you think?

frontend/common/SocialAuthLogin.tsx
jamestalton: The PF version shows icon where we show buttons with text.

Instead of having buttons like
- Login with Google
- Login with Github
- Login with OIDC

The UXD design is to have one "Login with" and then buttons with simpler labels.
This removes all the repetitive text.

frontend/awx/main/AwxLogin.tsx
jamestalton: In the AWX API the login is not under the /api/v2 path so we cannot wrap it.
Since that login is not used downstream, this works .

frontend/common/AnsibleLogin.tsx
jamestalton: The PF component has the ability to show the error text is in a standard way across products.
We wrap the submit in a try/catch and then set the error test with setHelperText(err.message).
The error boundary should not be catching/handling any errors.

nixocio (approved):

PR #2629, opened 2024-07-02
Fix input inventories order

MilanPospisil:
https://issues.redhat.com/browse/AAP-26468


jerabekjiri:
looks like the old UI has a way to reorder items Screenshot from 2024-07-02 17-53-50

might be a lot of work for just this specific case of reordering items in the modal in the new UI and not sure if it's worth it

jerabekjiri (approved):
nice work :+1:

PR #2630, opened 2024-07-02
Prevent instance group fallback in inventories tooltip

MilanPospisil:
https://issues.redhat.com/browse/AAP-24866


jerabekjiri (approved):
nice! :+1:

PR #2636, opened 2024-07-02
Hub - unify on "Create execution environment", fix test failing because of it

himdel:
The `can add and delete a new execution environment` test in `cypress/e2e/hub/execution-environments.cy.ts` fails sporadically, because when there are no EEs, the button has `data-cy=create-execution-environment`, while when there are EEs, the button has `data-cy=add-execution-environment`. These labels are generated from the button text, which is also inconsistent.

Fixing the button text to say "Create execution environment" on both the empty state screen and the list screen.

(the only place remaining that says "Add execution environment" is now `useAwxRoleMetadata`, which seems related to awx role names)

### DIFF
--- a/cypress/e2e/hub/execution-environments.cy.ts
+++ b/cypress/e2e/hub/execution-environments.cy.ts
@@ -28,7 +28,7 @@ describe('Execution Environments', () => {
     });
   });
 
-  it.skip('can add and delete a new execution environment', () => {
+  it('can add and delete a new execution environment', () => {
     cy.createHubRemoteRegistry().then((remoteRegistry) => {
       const eeName = `execution_environment_${randomString(3, undefined, { isLowercase: true })}`;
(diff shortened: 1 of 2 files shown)


himdel:
Cc @tiyiprh is "Create execution environment" for both buttons the right direction?  :)

ZitaNemeckova (approved):
Code-wise 👍 Waiting on UXD approval.

ZitaNemeckova:
@himdel Let's merge this. There's no consistency across projects yet.

himdel:
Agreed, we can always change it later, one less skipped test :)

tiyiprh:
> Cc @tiyiprh is "Create execution environment" for both buttons the right direction? :)

Sorry @himdel was on PTO last week but yes that's correct!

himdel:
Thanks!, no worries, sorry for rushing you :)

PR #2639, opened 2024-07-03
Unify Groups from buttons

MilanPospisil:
https://issues.redhat.com/browse/AAP-24926

I renamed buttons in groups add/edit to match the correct naming style of the application.

However are the buttons in the empty state implemented wrongly? I see that pattern in the new UI, so this should also follow the new UI style.


jerabekjiri (approved):
LGTM :+1:

PR #2649, opened 2024-07-04
Controller tests fix workflow approval tests

MilanPospisil:


cypress/support/auth.ts
```diff
@@ -117,7 +117,7 @@ Cypress.Commands.add('awxLogout', () => {
   cy.getByDataCy('account-menu')
     .click()
     .then(() => {
-      cy.intercept('GET', awxAPI`logout/`).as('logout');
+      cy.intercept('GET', `/api/logout/`).as('logout');
```
akus062381, line 120: Will this work against a downstream build?

MilanPospisil: I have asked about it James Talton and he said this login and logout is not used in downstream. So I hope :-)

akus062381 (approved):
We will hope that the Cypress tests running in the Jenkins build will show these passing :-)

PR #2661, opened 2024-07-09
Break apart project tests into smaller spec files and remove redundant project creation/sync

jamestalton:
Projects is the longest running spec.
- Breaks the projects spec into sub specs
- Removes creating a projects in beforeEach where not needed thus removing waits on project sync


vishsanghishetty (approved):

akus062381 (approved):
Thanks James! I've been wanting this same way of organizing all the specs for a while :-)

PR #2668, opened 2024-07-09
Order by ID instead of time for AWX ActivityStream 

TheRealHaoLiu:
Timestamp for activity stream is not indexed result in slow query with large data set. Switching to ID (which effectively will is order by created time) to improve performance


kialam:
E2E https://github.com/ansible/aap-ui/actions/runs/11861862996

kialam:
Closing in favor of https://github.com/ansible/aap-ui/pull/933

PR #2670, opened 2024-07-09
unskip inventory checks in activity stream page

ShaiahWren:
This pr unskips the following tests in `cypress/e2e/awx/administration/activity-stream.cy.ts`. Both tests have a description and have been run multiple times in the cypress UI tool with success.

`  it.skip('event column displays correct info', function () {`
`  it.skip('event details modal displays correct info', function () {`


skochay330 (approved):
Tests are passing on multiple runs, LGTM!

PR #2676, opened 2024-07-09
Fix labels for inventory type filter

daphnemaeve:
Jira issue: https://issues.redhat.com/browse/AAP-25463

Replaces ```Kind``` filter with ```Inventory Type``` filter in the Inventory list page. ```Inventory Type``` filters by the same value, but contains the correct labels for filter.

Before:
Screenshot 2024-07-09 at 4 18 28 PM

After:
Screenshot 2024-07-09 at 4 18 04 PM


kialam (changes requested):
The dropdown options are ordered differently than the order specified in `useInventoryTypeToolbarFilter()`. Once this is fixed this should be good to go.

kialam (approved):
👍

PR #2677, opened 2024-07-10
Update credentials and credential type cleanup in E2E tests

lgalis:
Update credentials and credential type cleanup in E2E tests


akus062381 (approved):
Looks good!

PR #2693, opened 2024-07-11
[AAP-25029] Fix Select EE Dialog

prat98:
This PR is for AAP-25029 which updates all old EE selects to the new `PageFormSingleSelectAwxResource` components

### DIFF
--- a/frontend/awx/resources/templates/WorkflowVisualizer/components/JobTemplateDetails.tsx
+++ b/frontend/awx/resources/templates/WorkflowVisualizer/components/JobTemplateDetails.tsx
@@ -213,13 +213,24 @@ export function JobTemplateDetails({
         label={t('Execution environment')}
         overriddenValue={template.summary_fields?.execution_environment?.name}
         isOverridden={
-          executionEnvironment?.id !== template.summary_fields?.execution_environment?.id
+          typeof executionEnvironment === 'object'
+            ? executionEnvironment?.id !== template.summary_fields?.execution_environment?.id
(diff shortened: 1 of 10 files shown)


frontend/awx/resources/templates/WorkflowVisualizer/components/JobTemplateDetails.tsx
```diff
@@ -213,13 +213,24 @@ export function JobTemplateDetails({
         label={t('Execution environment')}
         overriddenValue={template.summary_fields?.execution_environment?.name}
         isOverridden={
-          executionEnvironment?.id !== template.summary_fields?.execution_environment?.id
+          typeof executionEnvironment === 'object'
```
MilanPospisil, line 216: Great job, however, I see potential problems:

Since executionEnvironment can be now also id, which is number and can be 0, the condition above: 

isEmpty={!executionEnvironment}

is broken when ee has id 0.

And is it really necessary to have it also object, where does the object come from, when the aim of this PR is to have it number?

MilanPospisil:
I have tested the pages and it seems to work correctly. Great job @prat98 ! Please try to address one found small potential issue and its approved!

frontend/awx/resources/templates/WorkflowVisualizer/components/JobTemplateDetails.tsx
prat98: Just pushed a fix to ensure EE is always an object. This would address the issue with isEmpty and the need for conditional checks for object

jamestalton (approved):

PR #2702, opened 2024-07-12
Cypress Fix Component ChunkLoadError

jamestalton:
Testing putting all dependencies in one chunk to fix the load errors.

Rerunning multiple times.

I think this might fix some of the errors but not the ones to do with Monaco workers for json and yaml.
Might do a follow on PR to handle that.


vishsanghishetty (approved):
thanks for putting this together!

PR #2706, opened 2024-07-12
Refactor Activity stream link out

kialam:
- Replace logic to determine resource type using a param instead of relying on URL structure.


AlexSCorey (approved):
I pullled this down and tested it locally.  Works well.  Good job!

PR #2710, opened 2024-07-15
Fix JT and WFJT survey tests

jerabekjiri:
- replace `useMatch` with template type 
- fix `Failure to launch: Cannot read properties of undefined (reading 'map')` 
Screenshot from 2024-07-15 10-29-09


AlexSCorey (approved):
This looks good to me.  Well done!

PR #2712, opened 2024-07-15
[AAP-27011] Fix Job Output Details Functionality

prat98:
This is a PR for AAP-27011 which fixes the issue with the job output details functionality when clicking on certain job output rows. See Jira issue for more context.


marshmalien (approved):
Tested locally. LGTM!

PR #2728, opened 2024-07-16
Fix and refactor inventoryGroup tests for upstream and downstream

MilanPospisil:
Refactored inventoryGroup run command test to use shared function, because run command is executed in multiple places through the application.

Fix test in can run an ad-hoc command against a group's related group list view, added filter for particular item, previous code assumes the item is displayed on the page without filter.

### DIFF
--- a/cypress.awx.config.ts
+++ b/cypress.awx.config.ts
@@ -3,7 +3,7 @@ import { defineConfig } from 'cypress';
 import setValue from 'set-value';
 import { baseConfig } from './cypress.base.config';
 
-baseConfig.e2e!.specPattern = 'cypress/e2e/awx/resources/inventoryGroup.cy.ts';
+baseConfig.e2e!.specPattern = 'cypress/e2e/awx/**/*.cy.ts';
 baseConfig.e2e!.baseUrl = 'https://localhost:4101';
 baseConfig.component!.specPattern = 'frontend/awx/**/*.cy.{js,jsx,ts,tsx}';
 setValue(baseConfig, 'component.devServer.webpackConfig.devServer.port', 4201);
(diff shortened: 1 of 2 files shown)


ShaiahWren (commented):
Inventory Groups - List view:
failures 
- none

Inventory Groups - Related Groups Tab: 
failures:
- it(can add and disassociate new related groups). `POST 201 /api/v2/groups/` (create group) results in ui crash with message `Error: Model is disposed!` and as a result, the "Back to groups" button does not work. Fails are intermittent
   
- it(can add, edit and then disassociate existing related groups). `(fetch)OPTIONS 200 /api/v2/groups/` (Back to groups) results in ui crash with message `Error: Model is disposed!` This fails consistently, every time. 
- it(can run an ad-hoc command against a group's related group list view). The afterEach command has a cypress error for deleting organizations.


Inventory Groups - Hosts Tab: 
failures
- can edit a related host from the host tab. `
(fetch)GET 200 /api/v2/inventories/417/hosts/?order_by=name&page=1&page_size=10` results in ui crash with message `Error: Model is disposed!` and as a result, the filter in the following command does not work. Failures intermittent

The afterEach command has a cypress error for deleting organizations.

MilanPospisil:
I cant reproduce any of the error 'Model was disposed' that @ShaiahWren or @AlexSCorey reported. What browser are you using?

MilanPospisil:
I looked into it. It seems tests are failing in firefox and not in chrome, but the problem is not in the tests itself. I will write about it in aap-ui, this is UI issue. So this test is ok and can be merged. 

However the problem has to be solved as separate UI issue.

AlexSCorey (dismissed):

PR #2729, opened 2024-07-16
[AAP-27036] User preferences screen inconsistent UX

jamestalton:
Adds a details page to the user preferences.


framework/PageSettings/PageSettingsDetails.tsx
```diff
@@ -0,0 +1,67 @@
...
+      },
+    ],
+    [navigate, t]
+  );
+  const options = usePageSettingsOptions();
+
+  return (
+    <PageLayout>
+      <PageHeader
+        title={t('User Preferences')}
```
AlexSCorey, line 39:
```suggestion
        title={t('User preferences')}
```
Based on recent changes from @tiyiprh we probably want this to me sentence case.

framework/PageSettings/PageSettingsForm.tsx
```diff
@@ -0,0 +1,53 @@
...
+      navigate('..');
+      return Promise.resolve();
+    },
+    [navigate, setSettings]
+  );
+  const onCancel = useCallback(() => navigate('..'), [navigate]);
+  const options = usePageSettingsOptions();
+  return (
+    <PageLayout>
+      <PageHeader title={t('User Preferences')} titleHelp={t('Per user preferences.')} />
```
AlexSCorey, line 27:
```suggestion
      <PageHeader title={t('User preferences')} titleHelp={t('Per user preferences.')} />
```
Sentence casing here too.

framework/PageSettings/usePageSettingOptions.tsx
```diff
@@ -0,0 +1,117 @@
...
+    // Comment this onw out as it causes issue for now
+    // {
+    //   name: 'formLayout',
+    //   label: t('Form layout'),
+    //   helpText: t('Select the layout for forms.'),
+    //   options: [
+    //     { label: t('Vertical'), value: 'vertical' },
+    //     { label: t('Horizontal'), value: 'horizontal' },
+    //   ],
+    // },
```
AlexSCorey, line 87: Can we remove it, and file an issue to come back to it later if we want?
```suggestion
    
```

cypress/e2e/awx/administration/settings.cy.ts
```diff
@@ -4,10 +4,12 @@ describe('AWX Settings', () => {
...
     cy.navigateTo('awx', 'settings-preferences');
     cy.verifyPageTitle('User Preferences');
+    cy.clickButton(/^Edit$/);
     cy.singleSelectBy('#tablelayout', 'Compact');
     cy.getByDataCy('Submit').click();
     cy.navigateTo('awx', 'settings-preferences');
     cy.get('#tablelayout').should('contain', 'Compact');
+    cy.clickButton(/^Edit$/);
     cy.singleSelectBy('#tablelayout', 'Comfortable');
     cy.getByDataCy('Submit').click();
```
AlexSCorey, line 14: We should probably make an assertion on this test. Though it isn't a blocker IMO

framework/PageDetails/PageDetail.tsx
```diff
@@ -31,9 +32,13 @@ export function PageDetail(props: {
...
+      <DescriptionListDescriptionStyled id={id} data-cy={id}>
         {children}
-      </DescriptionListDescription>
+      </DescriptionListDescriptionStyled>
     </DescriptionListGroup>
   );
 }
+
+const DescriptionListDescriptionStyled = styled(DescriptionListDescription)`
+  opacity: 0.8;
```
AlexSCorey, line 43: We might want to use a PF utility class here for styling. https://www.patternfly.org/developer-resources/global-css-variables

AlexSCorey (changes requested):
Good work.  I think the only thing we must change before merge is the text casing.  Otherwise the other things are optional IMO.

framework/PageSettings/PageSettingsDetails.tsx
tiyiprh: Since this is a page title, it should be title case matching the navigation. If it were a create page it would be sentence case. Sorry there are slight differences that could be confusing.

framework/PageDetails/PageDetail.tsx
jamestalton: Changing the opacity makes the text easier to spot and makes it distinct from the label.
Need it to be the same color as before, and still have good contrast for accessibility.
Needs to support dark and light theme.

Many frameworks out there use opacity on text for just this reason.

PF does not have this built in.  I have discussed it with them but it has not landed in PF5.
I do not see any utility classes that would solve this.

AlexSCorey (approved):
Looks good to me!

PR #2733, opened 2024-07-16
[UXD][AAP-25027] AWX - updates to Users section breadcrumbs & headers

tiyiprh:
https://issues.redhat.com/browse/AAP-25027

Updates to Users section breadcrumbs & headers section breadcrumbs & headers


daphnemaeve (commented):
Looks good! Two last tests are failing in ```cypress/e2e/awx/access/users/users.cy.ts``` because it is expecting the old page title. Editing the two places where we verify the page title ```cy.verifyPageTitle('Edit User');``` in that file should resolve this!

daphnemaeve (approved):

PR #2737, opened 2024-07-16
Remove unreachable job statuses from workflow output status bar

marshmalien:
https://issues.redhat.com/browse/AAP-26858

Remove the "unreachable" status for nodes that have not been run and only display the statuses of completed jobs in the workflow output status bar 

_Example: 1 success, 1 never run_
Screenshot 2024-07-16 at 2 22 14 PM

_Example: 3 success, 1 fail, 1 cancel, 1 never run_
Screenshot 2024-07-16 at 2 15 17 PM


nixocio (approved):

vishsanghishetty (approved):
looks good to me, thanks for fixing!

PR #2738, opened 2024-07-16
[AAP-26919] Adds error handle to credentials step in template launch

AlexSCorey:
This PR adds credentials error handling for the job template prompt on launch workflow.  Essentially, if a job with has credentials with prompt on launch, if the user removes credentials for the template while in the prompt on launch workflow they must replace that credential with another of the same type.  They can add additional credentials as well.

This PR also refactors the prompt on launch workflow to bring it in line with the prompt on launch workflow used in the workflow visualizer and in schedules.


prat98 (approved):
Changes look good to me here! There is a case we may want to cover to ensure only one cred is selected for a credtype with the exception being vault creds

PR #2740, opened 2024-07-16
[AAP-27278] Fixes no call to launch end point on schedules

AlexSCorey:
Addresses AAP-27278.

at the top level schedule list we were failing to make a call to the launch end point.

### DIFF
--- a/cypress/e2e/awx/views/schedules.cy.ts
+++ b/cypress/e2e/awx/views/schedules.cy.ts
@@ -78,7 +78,7 @@ describe('Schedules - Create and Delete', () => {
       cy.navigateTo('awx', 'schedules');
       cy.verifyPageTitle('Schedules');
       cy.getByDataCy('create-schedule').click();
-      cy.verifyPageTitle('Create Schedule');
+      cy.verifyPageTitle('Create schedule');
       cy.selectDropdownOptionByResourceName('schedule_type', 'Job template');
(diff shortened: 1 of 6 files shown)


MilanPospisil:
I identified another problem with schedules, Im not sure if it separate issue, but it is maybe good to include fix into this PR?

How to reproduce it: 
Edit the schedules with Type: Management job

you will see fail:
TypeError
Cannot read properties of undefined (reading 'job_tags')
Call Stack
 setLaunchToWizardData
  frontend/awx/views/schedules/wizard/ScheduleSelectStep.tsx:131:9

cypress/e2e/awx/views/schedules.cy.ts
```diff
@@ -78,7 +78,7 @@ describe('Schedules - Create and Delete', () => {
       cy.navigateTo('awx', 'schedules');
       cy.verifyPageTitle('Schedules');
       cy.getByDataCy('create-schedule').click();
-      cy.verifyPageTitle('Create Schedule');
+      cy.verifyPageTitle('Create schedule');
```
acruzgon, line 81: this was already addressed in [here](https://github.com/ansible/ansible-ui/pull/2754/files#diff-6710eea15d62a82a05d60f79fceb32cf813b84e6929970b2dc9b90243d0aad6cR81)

acruzgon (approved):

PR #2772, opened 2024-07-19
[AAP-27455] Update docs helper function to link to either Upstream or Downstream Docs

prat98:
This PR is for AAP-27455. It aims to ensure both upstream and downstream builds have the appropriate doc links throughout the UI. The changes here were tested upstream and seem to be working fine. This is still a draft as this solution still needs to be tested downstream. In addition, the problem with figuring out the version of docs we want to link for downstream still needs to be solved. Currently I am just linking to the 2.4 docs but ideally we would want to be able to figure out what version of the docs we want through checking the version of AAP in a deployment and doing some mapping to discern the correct docs version. Another solution would be to try and default to latest at all times. Currently this is hard because there is no way to provide a `latest` version in the product docs path like we can for upstream docs. We would need to talk to the docs team to support a `latest` path to do this.

### DIFF
--- a/frontend/awx/common/util/getDocsBaseUrl.ts
+++ b/frontend/awx/common/util/getDocsBaseUrl.ts
@@ -1,6 +1,6 @@
 import { Config } from '../../interfaces/Config';
 
-export interface docPathDictionary {
+export interface DocPathDictionary {
   credentialTypes: string;
   credentials: string;
   organizations: string;
@@ -30,7 +30,7 @@ export interface docPathDictionary {
 
 export function getDocsBaseUrl(
   config: Config | null | undefined,
-  doc: keyof docPathDictionary
+  doc: keyof DocPathDictionary
 ): string {
   const version = config?.platformVersion || '2.5';
   const licenseType = config?.license_info?.license_type;
@@ -41,7 +41,7 @@ export function getDocsBaseUrl(
   }
 }
 
-const upstreamPaths: docPathDictionary = {
+const upstreamPaths: DocPathDictionary = {
   credentialTypes: 'userguide/credential_types.html',
   credentials: 'userguide/credentials.html',
   organizations: 'userguide/organizations.html',
@@ -70,7 +70,7 @@ const upstreamPaths: docPathDictionary = {
   schedules: 'userguide/scheduling.html',
 };
 
-const downstreamPaths: docPathDictionary = {
+const downstreamPaths: DocPathDictionary = {
   credentialTypes: 'automation_controller_user_guide/assembly-controller-custom-credentials',
   credentials: 'automation_controller_user_guide/controller-credentials',
   organizations: 'automation_controller_user_guide/assembly-controller-organizations',
--- a/frontend/awx/common/util/getDocsBaseUrl.ts
+++ b/frontend/awx/common/util/getDocsBaseUrl.ts
@@ -32,7 +32,7 @@ export function getDocsBaseUrl(
   config: Config | null | undefined,
   doc: keyof docPathDictionary
 ): string {
-  const version = '2.4';
+  const version = config?.platformVersion || '2.5';
   const licenseType = config?.license_info?.license_type;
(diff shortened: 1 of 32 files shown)


frontend/awx/common/util/getDocsBaseUrl.ts
```diff
@@ -1,10 +1,106 @@
...
+  eeMigration: 'upgrade-migration-guide/upgrade_to_ees.html',
+  jobTemplateSurveys: 'userguide/job_templates.html#surveys',
+  index: 'userguide/index.html',
+  hosts: 'userguide/hosts.html',
+  inventories: 'userguide/inventories.html',
+  constructedInventories: 'userguide/inventories.html#constructed-inventories',
+  managePlaybooksSC: 'userguide/projects.html#manage-playbooks-using-source-control',
+  projects: 'userguide/projects.html',
+  templates: 'userguide/job_templates.html',
+  // workflowTemplates: 'userguide/automation_controller_user_guide/controller-workflow-job-templates',
```
AlexSCorey, line 66: Why is this commented out?

frontend/awx/common/util/getDocsBaseUrl.ts
```diff
@@ -1,10 +1,106 @@
...
+  index: 'automation_controller_user_guide/index',
+  hosts: 'automation_controller_user_guide/controller-hosts',
+  inventories: 'automation_controller_user_guide/controller-inventories',
+  constructedInventories:
+    'automation_controller_user_guide/controller-inventories#ref-controller-constructed-inventories',
+  managePlaybooksSC:
+    'automation_controller_user_guide/controller-projects#ref-projects-manage-playbooks-with-source-control',
+  projects: 'automation_controller_user_guide/controller-projects',
+  templates: 'automation_controller_user_guide/controller-job-templates',
+  // workflowTemplates: 'automation_controller_user_guide/controller-workflow-job-templates',
```
AlexSCorey, line 99: Do we need this?

frontend/awx/common/util/getDocsBaseUrl.ts
```diff
@@ -1,10 +1,106 @@
...
+  workflowVisBuild: string;
+  jobs: string;
+  schedules: string;
+}
+
+export function getDocsBaseUrl(
+  config: Config | null | undefined,
+  doc: keyof DocPathDictionary
+): string {
+  const version = config?.platformVersion || '2.5';
```
AlexSCorey, line 35: I'm a bit nervous about hard coded 2.5 here.  Can we make it `latest` or do something so that we don't have the update this with each release?

AlexSCorey (changes requested):
I'd like to see at least 1 test or `getDocsBaseUrl` just to ensure it returns the expected string.

frontend/awx/common/util/getDocsBaseUrl.ts
keithjgrant: unfortunately the new downstream docs don't support a `latest` path like our old docs did, so this is the safest fallback

frontend/awx/common/util/getDocsBaseUrl.ts
prat98: Removed comment

frontend/awx/common/util/getDocsBaseUrl.ts
prat98: No it was just a remanent removed the comment

prat98:
> I'd like to see at least 1 test or `getDocsBaseUrl` just to ensure it returns the expected string.

@AlexSCorey I definitely agree a test here would be good but is it fine if we prioritize this for after Wednesday?

AlexSCorey (approved):
Looks good.  lets add the tests later

AlexSCorey (approved):
reapprovals

keithjgrant (approved):

nixocio:
As a team we decided that the failures were not related to those changes.

PR #2776, opened 2024-07-23
Repair schedules edit problem

MilanPospisil:
https://issues.redhat.com/browse/AAP-27555

Repaired 1st issue - wizard data were partialy deleted in details step in useEffect by mistake, which depeneded on whenever useEffects was able to start running or not - in tests, next was clicked before the useEffect run.

The effect was that rules and other data were somethimes missing in next wizard pages.


jerabekjiri (approved):
Tested locally, nice fix! :+1:

PR #2777, opened 2024-07-23
Wrapper over LabelGroup

MilanPospisil:
Fixes second part of issue: https://issues.redhat.com/browse/AAP-27555

Taken from ansible-hub-ui from Martins fix. 

https://github.com/ansible/ansible-hub-ui/blob/master/src/components/patternfly-wrappers/label-group.tsx


himdel (approved):
LGTM, prevents the "show more" label from submitting forms :)

(ideally this should be a patternfly fix, but LGTM :+1: )

PR #2786, opened 2024-07-23
[AAP-27450] Adds validation to prevent duplicate credential types 

AlexSCorey:
Across awx there are a few places where a user can select mutiple credentials.  In each of those cases there are limits.  
Limits
1) Cannot select more than 1 credential per credential type
except for Vault credentials
2) Cannot select more than 1 vault credential with the same vault ID.  Vault ID is an optional field.

### DIFF
--- a/frontend/awx/access/credentials/hooks/useCredentialsValidate.tsx
+++ b/frontend/awx/access/credentials/hooks/useCredentialsValidate.tsx
@@ -1,0 +1,55 @@
+import { useCallback } from 'react';
+import { useTranslation } from 'react-i18next';
+import { PromptFormValues } from '../../../resources/templates/WorkflowVisualizer/types';
+import { Credential } from '../../../interfaces/Credential';
+import { requestGet } from '../../../../common/crud/Data';
+import { awxAPI } from '../../../common/api/awx-utils';
+
+export function useCredentialsValidate() {
+  const { t } = useTranslation();
+  return useCallback(
+    async (selectedCredentials: PromptFormValues['credentials']) => {
+      const originalCredentials: number[] = [];
+      const newCredentials: Credential[] = [];
+      selectedCredentials?.forEach((credential) =>
+        !('summary_fields' in credential)
+          ? originalCredentials.push(credential.id)
+          : newCredentials.push(credential)
+      );
+      const fullDefaultCredentials: Credential[] = await Promise.all(
+        originalCredentials.map((defCred) =>
+          requestGet<Credential>(awxAPI`/credentials/${defCred.toString()}/`)
+        )
+      );
+      const allCredentials = [...(fullDefaultCredentials || []), ...newCredentials];
+      const vaultIds = allCredentials
+        .filter((credential) => credential.kind === 'vault' && credential.inputs.vault_id)
+        .map((vaultCred) => vaultCred.inputs.vault_id.toString());
+      const otherCredentialTypes = allCredentials
+        .filter((credential) => credential.kind !== 'vault')
+        .map((nonVaultCred) => nonVaultCred.summary_fields.credential_type.name);
+
+      const hasDuplicateVaultIds: boolean =
+        vaultIds.filter(
+          (vaultId, _index, array) => array.indexOf(vaultId) !== array.lastIndexOf(vaultId)
+        ).length > 0;
+      const duplicateCredentialTypes: string[] = [
+        ...new Set(
+          otherCredentialTypes.filter(
+            (cred, _index, array) => array.indexOf(cred) !== array.lastIndexOf(cred)
+          )
+        ),
+      ];
+
+      if (duplicateCredentialTypes.length > 0) {
+        return t(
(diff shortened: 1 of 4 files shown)


framework/PageInputs/PageAsyncMultiSelect.tsx
```diff
@@ -30,6 +30,7 @@ export interface PageAsyncMultiSelectProps<ValueT>
   onBrowse?: () => void;
 
   compareOptionValues?: (a: ValueT, b: ValueT) => boolean;
+  validate?: (items: ValueT[]) => Promise<string | undefined>;
```
marshmalien, line 33: The `validate` prop appears to be unused in PageAsyncMultiSelect. We could probably remove it.

framework/PageForm/Inputs/PageFormAsyncMultiSelect.tsx
```diff
@@ -40,6 +40,7 @@ export type PageFormAsyncMultiSelectProps<
   | 'onBrowse'
   | 'queryLabel'
   | 'compareOptionValues'
+  | 'validate'
```
marshmalien, line 43: PageFormAsyncMultiSelect has a `validate` prop defined on line 25. If we remove it from PageAsyncMultiSelect then we should remove it here as well.

frontend/awx/access/credentials/hooks/useCredentialsValidate.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { Credential } from '../../../interfaces/Credential';
+import { requestGet } from '../../../../common/crud/Data';
+import { awxAPI } from '../../../common/api/awx-utils';
+
+export function useCredentialsValidate() {
+  const { t } = useTranslation();
+  return useCallback(
+    async (selectedCredentials: PromptFormValues['credentials']) => {
+      const originalCredentials: number[] = [];
+      const newCredential: Credential[] = [];
```
marshmalien, line 13:
```suggestion
      const newCredentials: Credential[] = [];
```
We could rename `newCredential` to `newCredentials` for consistency

frontend/awx/common/PageFormMultiSelectAwxResource.tsx
```diff
@@ -65,11 +66,25 @@ export function PageFormMultiSelectAwxResource<
...
+                'summary_fields' in resource &&
+                typeof resource.summary_fields === 'object' &&
+                resource.summary_fields !== null &&
+                'credential_type' in resource.summary_fields &&
+                typeof resource.summary_fields.credential_type === 'object' &&
+                resource.summary_fields.credential_type !== null &&
+                'name' in resource.summary_fields.credential_type
+              ) {
+                returnObject.label = `${resource.name} | ${resource.summary_fields.credential_type.name as string}`;
+              }
```
marshmalien, line 85: Instead of adding specific logic for credential resources in this shared component, we could allow label customization through props. We could pass in a custom function, such as `formatLabel`, and use that with `PageFormMultiSelectAwxResource`.

```
const returnObject = {
  value: resource as PathValue<FormData, Name>,
  description: resource.description,
  label: props.formatLabel ? props.formatLabel(resource) : resource.name,
};
```

marshmalien (changes requested):
Tested locally and validation works as expected 🎉  We should cleanup the unused props.  The other code comments are more so suggestions and not blockers to merging. 

Screenshot 2024-07-26 at 11 30 38 AM

frontend/awx/access/credentials/hooks/useCredentialsValidate.tsx
```diff
@@ -0,0 +1,54 @@
...
+        originalCredentials.map((defCred) =>
+          requestGet<Credential>(awxAPI`/credentials/${defCred.toString()}/`)
+        )
+      );
+      const vaultIds = [...(fullDefaultCredentials || []), ...newCredential]
+        .filter((credential) => credential.kind === 'vault' && credential.inputs.vault_id)
+        .map((vaultCred) => vaultCred.inputs.vault_id.toString());
+      const otherCredentialTypes = [...(fullDefaultCredentials || []), ...newCredential]
+        .filter((credential) => credential.kind !== 'vault')
+        .map((nonVaultCred) => nonVaultCred.summary_fields.credential_type.name);
```
marshmalien, line 29: We could combine these into a single `allCredentials` array for simplicity

```
  const allCredentials = [...fullDefaultCredentials, ...newCredentials];

  const vaultIds = allCredentials
    .filter((credential) => credential.kind === 'vault' && credential.inputs.vault_id)
    .map((vaultCred) => vaultCred.inputs.vault_id.toString());

  const otherCredentialTypes = allCredentials
    .filter((credential) => credential.kind !== 'vault')
    .map((nonVaultCred) => nonVaultCred.summary_fields.credential_type.name);
    ```

marshmalien (approved):
🎉 Great fix!

nixocio (approved):

PR #2801, opened 2024-07-25
Fix inventories constructed flaky tests

daphnemaeve:
Jira issue: https://issues.redhat.com/browse/AAP-26768

### DIFF
--- a/cypress/e2e/awx/resources/inventoriesConstructed.cy.ts
+++ b/cypress/e2e/awx/resources/inventoriesConstructed.cy.ts
@@ -81,6 +81,7 @@ describe('Constructed Inventories CRUD Tests', () => {
       .document()
       .its('body')
       .find('.pf-v5-c-menu__content')
+      .should('have.length', 1)
       .within(() => {
         cy.getByDataCy('search-input').within(() => {
           cy.get('input').clear().type(invNames[0]);
--- a/cypress/e2e/awx/resources/inventoriesConstructed.cy.ts
+++ b/cypress/e2e/awx/resources/inventoriesConstructed.cy.ts
@@ -41,7 +41,6 @@ describe('Constructed Inventories CRUD Tests', () => {
   });
 
   after(() => {
-    //cy.deleteAwxInventory(newInventory);
     cy.deleteAwxInventory(newInventory);
     cy.deleteAwxInstanceGroup(instanceGroup);
     cy.deleteAwxOrganization(organization);
@@ -183,7 +182,7 @@ describe('Constructed Inventories CRUD Tests', () => {
       .its('response.statusCode')
       .then((statusCode) => {
         expect(statusCode).to.be.equal(200);
-        cy.verifyPageTitle(constructedInv.name);
+        cy.verifyPageTitle(newInventory.name);
       });
 
     cy.intercept('POST', awxAPI`/inventory_sources/*/update`).as('syncInventory');
--- a/cypress/e2e/awx/resources/inventoriesConstructed.cy.ts
+++ b/cypress/e2e/awx/resources/inventoriesConstructed.cy.ts
@@ -157,15 +157,8 @@ describe('Constructed Inventories CRUD Tests', () => {
       "product_dev": "account_alias == 'product_dev'"
       }}`
     );
-    cy.intercept('PATCH', awxAPI`/constructed_inventories/*`).as('saveInv');
     cy.clickButton(/^Save inventory$/);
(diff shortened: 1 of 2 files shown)


cypress/e2e/awx/resources/inventoriesConstructed.cy.ts
```diff
@@ -22,36 +19,34 @@ describe('Constructed Inventories CRUD Tests', () => {
...
 
   beforeEach(() => {
-    inventoriesList = [];
-    invNames = [];
-    for (let i = 0; i < invToCreate; i++) {
-      cy.createAwxInventory(organization).then((inv) => {
-        inventoriesList.push(inv);
-      });
-    }
+    cy.createAwxInventory(organization).then((inv) => {
```
AlexSCorey, line 30: I think we are creating this Inventory to prepopulate the input inventories field on the constructed inventory form.  If thats true, lets create this inventory 1 time in the `before`, and then delete it 1 time in the `after` hooks.

cypress/e2e/awx/resources/inventoriesConstructed.cy.ts
```diff
@@ -64,28 +59,30 @@ describe('Constructed Inventories CRUD Tests', () => {
...
+        cy.getByDataCy('search-input').within(() => {
+          cy.get('input').clear().type(inventory.name);
+        });
+        cy.contains('.pf-v5-c-menu__item-text', inventory.name)
+          .parent()
+          .within(() => {
+            cy.get('input').click();
+          });
+      });
+
```
AlexSCorey, line 85:
```suggestion
    //get inventories select
  cy.multiSelectBy('input-inventories', [inventory.name])
```
I think that'll work.

cypress/e2e/awx/resources/inventoriesConstructed.cy.ts
```diff
@@ -158,14 +150,55 @@ describe('Constructed Inventories CRUD Tests', () => {
       });
   });
 
-  it.skip('shows a failed sync on the constructed inventory if the user sets strict to true and enters bad variables', () => {
-    //Create a constructed inventory in the beforeEach hook
+  it('shows a failed sync on the constructed inventory if the user sets strict to true and enters bad variables', () => {
```
AlexSCorey, line 153: This feel like its testing the API more than its testing the UI.  In essence I think what we should care about is that we can submit a new inventory, I think other tests have covered that, and that we show the sync status on the details page..... that could be component test.  

I this test is solid and hasn't show any flake over several runs I say keep it.  Otherwise, I think we can remove it.

AlexSCorey (changes requested):
Looks really good.  Just 1 small change to use a new cypress command to select the inputs inventory. Then, consider removing that 1 test.

cypress/e2e/awx/resources/inventoriesConstructed.cy.ts
daphnemaeve: I'd definitely prefer to use the custom command, but it seems to be a source of flakiness for this test/field. The changes I've made run consistently and use the same code as the singleSelectBy, just searching instead for a button with the correct text, and adding an assertion that there should only be one result.

cypress/e2e/awx/resources/inventoriesConstructed.cy.ts
daphnemaeve: Okay! I agree that it'd be okay to remove if it's flaky, but it so far seems to be passing over several local runs both downstream and upstream. I can also run the PR check here again to ensure that it isn't flaky, just to be sure!

cypress/e2e/awx/resources/inventoriesConstructed.cy.ts
daphnemaeve: Met with Alex and decided to implement ```cy.multiSelectByDataCy()```

daphnemaeve:
@AlexSCorey helped verify that tests are passing both upstream and downstream, so rerunning checks and will merge when passing.

vidyanambiar (approved):

vishsanghishetty:
thanks for adding the changes @daphnemaeve

vishsanghishetty (approved):

vishsanghishetty (dismissed):

vidyanambiar (approved):
Tested against dev server and VM ✅

vidyanambiar (approved):
Tested against the dev server + VM ✅

PR #2802, opened 2024-07-25
Reduces org calls

AlexSCorey:
We know that org creation in the gateway takes lots of resources during testing.  This PR removes as many org create calls as possible while still keeping tests green.  Generally, doing that involves wrapping the whole test suite in a call to create and then delete the org in a `before` and `after` functions.


skochay330:
There are backend investigations in progress to address this issue, closing the PR

PR #2816, opened 2024-07-26
[AAP-27827] Updates to Workflow Approvals section table headers/field labels casing

tiyiprh:
https://issues.redhat.com/browse/AAP-27827
Updating the casing on the table headers/card headers/field labels throughout AAP with the correct casing and wording in the Workflow Approvals section.

### DIFF
--- a/frontend/common/columns.tsx
+++ b/frontend/common/columns.tsx
@@ -20,7 +20,7 @@ export function useIdColumn<T extends { id: number }>(isHidden: boolean = true) 
   const { t } = useTranslation();
   const column = useMemo<ITableColumn<T>>(
     () => ({
-      header: t('Id'),
+      header: t('ID'),
       cell: (team) => team.id,
       minWidth: 0,
       table: isHidden ? ColumnTableOption.hidden : undefined,


frontend/awx/common/awx-toolbar-filters.tsx
```diff
@@ -77,7 +77,7 @@ export function useCreatedByToolbarFilter() {
   return useMemo<IToolbarFilter>(
     () => ({
       key: 'created-by',
-      label: t('Created by'),
+      label: t('Created'),
```
prat98, line 80: I think this needs to stay as `Created by` cause `Created` is another filter. One get's a user the other gets the date the resource was created

frontend/awx/resources/projects/Projects.cy.tsx
prat98, line 1: This should also change for reason stated above

frontend/awx/resources/templates/TemplatesList.cy.tsx
prat98, line 1: Same with this change, should stay as `Created by`

prat98 (approved):
LGTM!

PR #2823, opened 2024-07-29
Collections tests unskip

MilanPospisil:


cypress/e2e/hub/collections-list.cy.ts
```diff
@@ -223,3 +218,9 @@ describe.skip('Collections List', () => {
     cy.deleteCollection(collectionName, namespace.name, repository);
   });
 });
+
+function actionClick(item: string, action: string) {
```
ZitaNemeckova, line 222: Why not use `clickTableRowAction` ?

ZitaNemeckova (approved):

cypress/e2e/hub/collections-list.cy.ts
MilanPospisil: It does not work here, not sure why, but we must now fully focus the project to make it finished.

PR #2824, opened 2024-07-29
[AAP-27745] Clear the resource field when the node_ type changes, not when node_type is dirty

lgalis:
Clear the resource field when the node_ type changes, not when node_type is dirty


marshmalien:
This fix should address this flaky test: https://issues.redhat.com/browse/AAP-27494

The test would fail after a few seconds because the wizard resource select would clear, the wizard would click "Next" and subsequently error out due to the cleared required input select. This fix removes the buggy clearing behavior that resulted in flaky test runs.

marshmalien (approved):

nixocio (approved):

PR #2834, opened 2024-07-29
[AAP-27827] Updates to Rule Audit section table headers/field labels casing

tiyiprh:
https://issues.redhat.com/browse/AAP-27827
Updating the casing on the table headers/card headers/field labels throughout AAP with the correct casing and wording in the Rule Audit section.


lgalis (changes requested):
Need to update the column here : https://github.com/ansible/ansible-ui/blob/main/frontend/eda/rule-audit/RuleAudit.cy.tsx#L90

The test failing on the E2E was fixed as far as I know, a rebase should fix that.

lgalis (approved):

PR #2848, opened 2024-07-31
Skipped test containing logout in executionEnvironments.cy.ts

vidyanambiar:
Multiple PRs in the AWX test runs are showing failures caused by random 401 errors.

Looking at the test replays it appears that the Cypress session that we create before all the tests run is not being properly restored in a few cases (intermittently).
Example: https://cloud.cypress.io/projects/77sud6/runs/8127/overview?roarHideRunsWithDiffGroupsAndTags=1

There is one test in `executionEnvironments.cy.ts` that includes a logout and login. I wonder if there is a race condition where the tests running in parallel rely on the "Awx" session being present at the same time as this logout/login being triggered. 

This PR skips that test to see if it helps the 401 issue.

### DIFF
--- a/cypress/e2e/awx/resources/executionEnvironments.cy.ts
+++ b/cypress/e2e/awx/resources/executionEnvironments.cy.ts
@@ -153,7 +153,8 @@ describe('Execution Environments', () => {
       cy.clickModalButton('Close');
     });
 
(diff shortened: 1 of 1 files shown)


nixocio (approved):

prat98 (approved):
LGTM! 🚀

PR #2870, opened 2024-08-05
AAP-28425 : show None for Proxy URL, add a link to Repositories labels

ZitaNemeckova:
- Repositories are links to that repo detail page
- Proxy URL says None instead of being hidden

Part of https://issues.redhat.com/browse/AAP-28425

Before:
Screenshot 2024-08-05 at 15 34 46

After:
Screenshot 2024-08-05 at 15 31 24


himdel (approved):
Works for me :)

PR #2874, opened 2024-08-05
EE: Use in Controller

himdel:
make sure the "Use in Controller" action is available in EE list row kebabs, EE detail header, and EE detail Images tab list row kebabs, and leads to Automation Controller > Infrastructure > EEs > Add

Issue: AAP-26404

This makes the feature ready for the hackathon,
but does NOT close AAP-26404, because there's still a TODO of moving this to downstream only, and making it possible to inject actions here. (And tests.)
(For now, the feature also appears upstream, but only does a console.error about a missing route.)

(Does close AAH-2515)

---


ZitaNemeckova (approved):
Tested downstream. Works as expected. LGTM 👍

PR #2876, opened 2024-08-05
Hub notification link

appuk:
https://issues.redhat.com/browse/AAP-28024


himdel:
@appuk there's also an `ExternalLink` component that already does this.

But .. also, IIUC the "tech preview" vs "full experience" bits will be removed again, and that link should now go back to loading `HubRoute.Approvals`, probably without a new tab? (Not sure if now, or pending the Approvals ..approval re feature parity :))

himdel:
Closing, reverted #2759, removing the need to open notifications links in new window

PR #2879, opened 2024-08-05
[AAP-28713] Hub collections documentation

AlexSCorey:
This addresses 2 things:
1. First commit (98fb0a35db6cf078aa7399e7f150bf54f67e949d) addresses AAP-28713.
2. The second commit (5d076b96d47d4289e3ba837f90f87438eb5ef444) removes the drawer close button because it can't be reopened without navigating away from the view and returning, or refreshing the page.  Also, this button does not seem to be present on the old UI.  Its a POC and I put it in a separate commit to easily remove that work if we decide we want to keep it.


marshmalien (approved):
lgtm!

PR #2889, opened 2024-08-06
Hub roles

vidyanambiar:


frontend/hub/access/roles/RolePage/HubRoleForm.tsx
```diff
@@ -0,0 +1,177 @@
...
+import { useGet } from '../../../../common/crud/useGet';
+import { usePatchRequest } from '../../../../common/crud/usePatchRequest';
+import { useIsValidRoleName } from '../hooks/useIsValidRoleName';
+
+export function CreateRole(props: { breadcrumbLabelForPreviousPage?: string }) {
+  const { t } = useTranslation();
+  const navigate = useNavigate();
+  const pageNavigate = usePageNavigate();
+
+  useInvalidateCacheOnUnmount();
```
nixocio, line 32: Do we need this one?

frontend/hub/access/roles/RolePage/HubRoleForm.tsx
```diff
@@ -0,0 +1,177 @@
...
+  const pageNavigate = usePageNavigate();
+  const params = useParams<{ id?: string }>();
+  const id = Number(params.id);
+  const { data: Role } = useGet<HubRbacRole>(hubAPI`/_ui/v2/role_definitions/${id.toString()}/`);
+
+  useInvalidateCacheOnUnmount();
+
+  const postRequest = usePostRequest<Partial<HubRbacRole>, HubRbacRole>();
+  const patchRequest = usePatchRequest<Partial<HubRbacRole>, HubRbacRole>();
+
```
nixocio, line 79: When attempting to edit a role already existent, but editable.  Just a few field should be editable, basically name and description. But generates an error.


image

frontend/hub/access/roles/RolePage/HubRoleDetails.tsx
```diff
@@ -0,0 +1,73 @@
...
+                  })
+                )
+              }
+            />
+          </PageDetail>
+          <PageDetail label={t('Modified')}>
+            <DateTimeCell value={role?.modified} />
+          </PageDetail>
+          <PageDetail label={t('Editable')}>
+            {role?.managed ? t('Built-in') : t('Editable')}
```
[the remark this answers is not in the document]
vidyanambiar: @nixocio This is how we have it on all the Roles UIs including AWX/EDA and the tests are wired to use them too. 
image
We can probably revisit this and consider changing it if there is time at the end after the RBAC implementation. WDYT?

frontend/hub/access/roles/RolePage/HubRoleForm.tsx
vidyanambiar: Honestly, I'm not sure. I see that both Keith and Mike used it in their implementations of the AWX and EDA roles. @keithjgrant do you know why this is needed?

frontend/hub/access/roles/RolePage/HubRoleDetails.tsx
```diff
@@ -0,0 +1,73 @@
...
+          </PageDetail>
+        </PageDetails>
+        <PageDetails disableScroll={true} numberOfColumns={'single'}>
+          <PageDetail label={t('Permissions')}>
+            <HubRolePermissions role={role} />
```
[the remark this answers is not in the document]
vidyanambiar: image
The permission chips seem to align with the column above. But they just use standard components ExpandableRowContent and DescriptionList for creating the expanded row for AWX/EDA/Hub. Are you recommending moving the chips closer to the content type?

frontend/hub/access/roles/RolePage/HubRoleForm.tsx
vidyanambiar: There is an API bug [AAH-3348](https://issues.redhat.com/browse/AAH-3348). If this role is built-in it should not be editable. However the API is returning the wrong value of the `managed` flag so the UI thinks it is editable.

frontend/hub/access/roles/RolePage/HubRoleForm.tsx
keithjgrant: I copied that from the EDA form, so I'm not 100% sure. My guess is it ensures that SWR reloads a fresh copy of the Role object after it redirects to the details page when saving.

vidyanambiar: Thanks @keithjgrant ! @nixocio I'm inclined to keep this as it's been working on the other roles UIs.

frontend/hub/access/roles/RolePage/HubRoleDetails.tsx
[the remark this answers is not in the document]
nixocio: That was on details page, but if they are standards components we can revisit that one later.

nixocio (approved):

PR #2901, opened 2024-08-07
Redirect to ?next path on login

keithjgrant:
If a `?next` query param is provided, redirect to it upon successful login (for oath)


mabashian (approved):

mabashian:
@keithjgrant and I walked through this and confirmed that the flow is working.  The only gotcha was that the UI's proxy isn't letting `/o` requests through.  Once we tweaked the proxy we were able to follow the entire flow through from creating an application to authorizing and generating a token.

PR #2913, opened 2024-08-08
[AAP-28859] Fix Subscription Wizard Popup Bug

prat98:
This PR is for AAP-28859 which aims to fix the subscription wizard popup bug. It removes `platformVersion` from awxConfig and introduces a new context `docsVersion` which downstream can use to pass the correct version. Upstream, we will leave docsVersion as `undefined` as it is not needed for upstream docs. The related downstream PR can be found here: https://github.com/ansible/aap-ui/pull/473


keithjgrant (approved):
lgtm. good call on renaming the docs hook

PR #2924, opened 2024-08-08
Hub User Roles

vidyanambiar:
[AAP-28446](https://issues.redhat.com/browse/AAP-28446)

Content for User -> Roles tab
- Display role assignments for the user.
- System-level roles do not have a content-type, hence the Resource Name column will be empty.
- In the `Add roles` wizard, if the Resource Type selected is "System" then the step to select resources will be hidden.

Since there is no upstream implementation of users pages, this can be tested with the following steps:
1. Get the user ID you want to test with using `<server>/api/galaxy/_ui/v2/users/`
2. Access `https://localhost:4102/access/users/<user_id_from_above>/roles`

Note: The step number on the wizard does not update when a step is hidden. I see this in other wizards too. This can be a separate investigation from this PR.

https://github.com/user-attachments/assets/0b7aa3cf-6e78-4138-9094-551d793d6248


nixocio (approved):
I got this code locally. It works as expected.

Great job @vidyanambiar

PR #2935, opened 2024-08-09
[AAP-28442] Add initial roles tab for teams

nixocio:
Add initial roles tab for teams

To test this.

1. Create a group using the old ui, get the group id
2.  https://localhost:4102/access/teams/<group_id>/roles


vidyanambiar (approved):
Works great ❇️

PR #2936, opened 2024-08-09
[AAP-29041] Use useGetJobOutputUrl to resolve page not found error

daphnemaeve:
Jira issue: https://issues.redhat.com/browse/AAP-29041

Resolves page not found error in downstream, which was being caused by different links being needed for upstream vs downstream.

### DIFF
--- a/frontend/awx/resources/templates/components/Sparkline.tsx
+++ b/frontend/awx/resources/templates/components/Sparkline.tsx
@@ -13,6 +13,14 @@ const Wrapper = styled.div`
   flex-wrap: wrap;
 `;
 export const Sparkline = ({ jobs }: { jobs: SummaryFieldRecentJob[] | undefined }) => {
+  const JOB_TYPE_URL_SEGMENTS: { [char: string]: string } = {
+    job: 'playbook',
+    project_update: 'project',
+    system_job: 'management',
+    inventory_update: 'inventory',
+    ad_hoc_command: 'command',
+    workflow_job: 'workflow',
+  };
   const { t } = useTranslation();
   const getJobOutputUrl = useGetJobOutputUrl();
   const generateTooltip = (job: SummaryFieldRecentJob) => (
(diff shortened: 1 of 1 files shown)


keithjgrant (approved):

frontend/awx/resources/templates/components/Sparkline.tsx
```diff
@@ -40,7 +43,11 @@ export const Sparkline = ({ jobs }: { jobs: SummaryFieldRecentJob[] | undefined
     <Tooltip position="top" content={generateTooltip(job)} key={job.id}>
       <Link
         aria-label={t(`View job ${job.id}`)}
-        to={`/jobs/${JOB_TYPE_URL_SEGMENTS[job.type]}/${job.id}/output`}
+        to={
+          getJobOutputUrl(job as UnifiedJob)
+            ? getJobOutputUrl(job as UnifiedJob)
+            : `/jobs/${JOB_TYPE_URL_SEGMENTS[job.type]}/${job.id}/output`
```
prat98, line 49: Maybe you can try to use the null coalescing operator here. That way the function to get the url isn't called twice.
```suggestion
          getJobOutputUrl(job as UnifiedJob) ?? `/jobs/${JOB_TYPE_URL_SEGMENTS[job.type]}/${job.id}/output`
```

prat98: if `getJobOutputUrl` is `undefined` or `null` it'll default to using  `/jobs/${JOB_TYPE_URL_SEGMENTS[job.type]}/${job.id}/output`. If `getJobOutputUrl` returns a non null value, then it will use that.

vidyanambiar (commented):
lgtm 👍 I agree with @prat98 's comment above.

frontend/awx/resources/templates/components/Sparkline.tsx
daphnemaeve: I agree and like that better, but it gets a component test failure in InventoryHostDetails. I can try to make a change or two to see if I can fix that problem.

daphnemaeve: For some reason there is a test failure when using the null coalescing operator, but was able to reduce the number of calls by defining jobOutputUrl and then using its value for the ternary operation. Checked with @prat98 and resolving this change :)

prat98 (approved):
LGTM! 🚀

PR #2940, opened 2024-08-12
Fix access UI filters

vidyanambiar:
The team/user access list filters are currently equality filters and need to support the `__icontains` query.

### DIFF
--- a/frontend/common/access/components/TeamAccess.tsx
+++ b/frontend/common/access/components/TeamAccess.tsx
@@ -38,7 +38,7 @@ export function TeamAccess(props: {
           label: t('Team name'),
         },
       }}
-      toolbarNameColumnFiltersValues={{ label: t('Team name'), query: 'team__name' }}
+      toolbarNameColumnFiltersValues={{ label: t('Team name'), query: 'team__name__icontains' }}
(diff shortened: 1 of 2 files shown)


frontend/common/access/components/TeamAccess.tsx
```diff
@@ -38,7 +38,7 @@ export function TeamAccess(props: {
           label: t('Team name'),
         },
       }}
-      toolbarNameColumnFiltersValues={{ label: t('Team name'), query: 'team__name' }}
+      toolbarNameColumnFiltersValues={{ label: t('Team name'), query: 'team__name__icontains' }}
```
nixocio, line 41: @vidyanambiar, do we need to verify if the other APIs support icontains as well, or is it a feature that's intrinsic to all of them?

nixocio (approved):
@vidyanambiar I got the code locally, and works for hub.

frontend/common/access/components/TeamAccess.tsx
vidyanambiar: Since this comes from DAB, it applies to all the services. I did a quick check on AWX/EDA resources as well and it seems to work as expected. 👍

PR #2948, opened 2024-08-13
Collections dependencies tab e2e tests.

kialam:


MilanPospisil:
Thanks for huge test coverage Kia! I have only one suggestion for improvement - what about creating collection once in before and use it in all its? Because you are intercepting data and mocking them, collection can still be the same. Creating collection is time consuming.

MilanPospisil (approved):
LGTM :+1:

keithjgrant (approved):

PR #2953, opened 2024-08-13
AWX E2E using GitHub matrix vs Cypress parallelization

jamestalton:
Updates AWX E2E tests to use a GitHub matrix instead of Cypress parallelization of tests.

Cypress parallelization is only supported if paying for and reporting to the Cypress dashboard. The Cypress costs for PRs have gotten out of control and we need to find another solution. This is the solution.

Nightly E2E runs will still report to the dashboard and we will be able to use that to determine flaky tests and such. In addition, if the label `dashboard` is added to a PR, it will run using Cypress parallelization and report to the dashboard.

There are really good benefits from using the GitHub Matrix
- The test times are lower then the current cypress PR runs because there are more runners.
-  If any test runner fails, you can restart just that one test runner, vs Cypress rerunning all tests.
- It is easy to see which tests are taking the longest.

The Github Matrix has been setup to be dynamic, generated from the directories in the e2e/awx directory. This allows us to easily split up long running tests into their own runner by moving them to their own directory.

Screenshot 2024-08-15 at 11 52 26 AM


akus062381 (dismissed):
Looks good!

nixocio (approved):

PR #2954, opened 2024-08-13
Undeprecate collection

MilanPospisil:
Undeprecate collection was missing and forgotten operation.

This PR adds it, also it allows to show deprecated collections in namespace-collections.

It adds tests.

However, framework does not offer dynamic labels, so I used text deprecate/undeprecate and the action then sends the request based on current state of collection. If its deprecated, it will send undeprecate request and vica versa.


himdel:
Not so sure about this one, can we just add Deprecate and Undeprecate as 2 separate actions, and hide the one that does not apply?

PR #2969, opened 2024-08-14
[AAP-29018] Update source-event mappings options lists

lgalis:
Changes for the source-event stream mappings

Rulebook Activation form:

-  The 'Manage event stream' link is disabled if a rulebook is not selected or is there are no event streams
-  The labels for the event stream in the Create Activation form are closable, removing the closed mapping

- Screenshot from 2024-08-14 14-03-14

Mapping dialog

-  The Add event stream mapping is hidden if there are no more sources or event streams to select
-  When the dialog open, a blank mapping is displayed if there are no mappings previously created

Screenshot from 2024-08-15 06-12-00

### DIFF
--- a/frontend/eda/rulebook-activations/RulebookActivationForm.tsx
+++ b/frontend/eda/rulebook-activations/RulebookActivationForm.tsx
@@ -248,7 +248,9 @@ export function RulebookActivationInputs() {
                 />
               )
             }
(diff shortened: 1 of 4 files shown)


tiyiprh (approved):
Looks good to me so far! I know there are still updates to be made

PR #2981, opened 2024-08-15
[UXD][AAP-28389] AWX - fix to the popovers on templates

tiyiprh:
https://issues.redhat.com/browse/AAP-28389
Fix to the popovers on templates to match IBM's guidelines:
-punctuation should be outside of quotation mark
-use " " instead of ' '


prat98 (approved):
LGTM!

PR #2991, opened 2024-08-15
fixes to inconsistencies in hub buttons

tiyiprh:
Part of this PR: https://github.com/ansible/ansible-ui/pull/2669
Noticed some additional inconsistencies in my review of Hub after the above PR was merged:

Collections:
-the bulk action order was different from the row level
Execution Environment:
-the row level action order was different than on the details page


himdel (approved):
Ah, consistent ordering, thanks! :)

PR #3004, opened 2024-08-16
Unskip tests that were skipped due to 401 error

akus062381:


cypress/e2e/awx/access/organizations/organizations.cy.ts
```diff
@@ -61,7 +61,7 @@ describe('Organizations: Edit and Delete', function () {
   });
 
   //Skipping due to https://issues.redhat.com/browse/AAP-28597
```
nixocio, line 63: @akus062381 Do we want to remove those comment as well?

nixocio (approved):

nixocio (approved):
if CI is green. :merge:

PR #3016, opened 2024-08-19
Add Collection Import Log e2e tests

kialam:
As part of testing the error state of the CollectionImportLog component, I removed the `Alert` component from the JSX as it never shows up.


AlexSCorey (approved):
I pulled this down and manually tested the UI along with running the e2e tests.  Everything worked as expected.  Well done!

PR #3018, opened 2024-08-19
[AAP-29149] Collection version switch bugfix

prat98:
This PR is for AAP-29149, it fixes the issue with the collection version automatically switching when a new one is created. In addition, the user is notified with an alert that allows them to refresh the page if there is a new collection version available.

Test Steps:

1. Create a new collection under the `community` repo on hub (don't have to deal with approvals after upload when uploading to community repo). Use this tar file to make the collection version
[redhat-ansible_on_clouds-2.5.0.tar.gz](https://github.com/user-attachments/files/16682921/redhat-ansible_on_clouds-2.5.0.tar.gz)

2. Navigate to the collection named `ansible_on_clouds` (the default version may be different from the version you uploaded) 
3. In another tab, upload another version of the collection also to the `community` repo. Use this tar file to make the collection version  
[redhat-ansible_on_clouds-2.6.0.tar.gz](https://github.com/user-attachments/files/16682936/redhat-ansible_on_clouds-2.6.0.tar.gz)

4. Go back to the the first collection details tab wait for a notification to pop-up which will allow you to navigate to the newly uploaded collection. (In addition, note that the version has not automatically changed even with there being a new version available). You should be able to click on the link in the notification and it should take you to the newly uploaded version

### DIFF
--- a/cypress/e2e/hub/collections-detail.cy.ts
+++ b/cypress/e2e/hub/collections-detail.cy.ts
@@ -175,7 +175,7 @@ describe('Collections Details', () => {
     });
   });
 
-  it.skip('can sign a selected version of a collection', () => {
+  it('can sign a selected version of a collection', () => {
     cy.uploadCollection(collectionName, namespace.name).then(() => {
       cy.galaxykit(
         'collection move',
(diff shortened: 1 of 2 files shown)


prat98:
> Some of the test failures may be valid, it seems this is no longer autoreloading the chosen version at all?
> 
> (sorry for the close+open, misclicked :))

No worries yeah I just caught the issue with the tests, looking into a fix right now

keithjgrant (approved):

PR #3020, opened 2024-08-20
Milan hub tests remove move to

MilanPospisil:
Gets rid of need to approve or move collections by uploading directly into validated repository. The tests are now faster, possibly more stable and we dont need to implement galaxykit move to using API requests :-)

### DIFF
--- a/cypress/support/hub-commands.ts
+++ b/cypress/support/hub-commands.ts
@@ -306,7 +306,7 @@ Cypress.Commands.add(
 
 Cypress.Commands.add(
   'uploadCollection',
-  (collection: string, namespace: string, version: string, repository? : string) => {
+  (collection: string, namespace: string, version: string, repository?: string) => {
     cy.galaxykit(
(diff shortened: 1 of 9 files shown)


cypress/support/hub-commands.ts
```diff
@@ -319,7 +319,7 @@ Cypress.Commands.add(
         formData.append('file', Cypress.Blob.binaryStringToBlob(fileData), filePath);
 
         cy.hubPostRequest({
-          url: hubAPI`/v3/plugin/ansible/content/staging/collections/artifacts/`,
+          url: hubAPI`/v3/plugin/ansible/content/${repository || 'validated'}/collections/artifacts/`,
```
himdel, line 322: Slightly worried about `staging` not being the default since that's where collections go when uploaded the normal way.
But .. I suppose most tests just care about having an approved collection, and this achieves that :+1:

himdel (approved):

PR #3032, opened 2024-08-21
fix cred tests

appuk:
Fix flakey Credential type tab tests
https://issues.redhat.com/browse/AAP-26674


AlexSCorey (approved):
Pulled down and tested locally against upstream.  Works as expected.

PR #3034, opened 2024-08-21
Select Dialog Fixes

jamestalton:
There is a timing issue with browse dialogs that use dynamic filters.
It should not impact customers, but it does impact fast tests.
Long term we should look for a fix, but for now, replacing test selection of items to not use the browse dialogs.


AlexSCorey (changes requested):
Lets remove the commented code

kialam (dismissed):

PR #3036, opened 2024-08-21
Reconfigure EDA Access Commands

akus062381:
Move EDA access commands to a separate file.


vidyanambiar (approved):

lgalis (approved):

vidyanambiar:
The only failures are on the EDA E2E tests and with the EDA server not expected to work for the time being, I think this PR is ready for merge.

PR #3038, opened 2024-08-21
[AAP-29664] Navigation after collection deletion

AlexSCorey:
This addresses AAP-29664.  After deleting a collection the user will be navigated to the collections list every time.  

We did see a case where if deleting only a version of a collection the user might want to remain on the details view for that collection, but decided that since the details will would likely be exactly the same after deleting a version the user could become confused as they'd expect to see different information.

Additionally the approach of deleting an items results in the user being navigated to a list view is an established pattern in the new UI so this bring in consistency.


kialam (dismissed):
Pulled this down and tested locally. Works when deleting from collections detail and namespace -> collections tab views. Nice work!

kialam (approved):

PR #3040, opened 2024-08-22
Remove tech preview

ZitaNemeckova:
As a part of https://issues.redhat.com/browse/AAP-29059

Removing remaining code that mentions "tech preview".


himdel (approved):
Ah, looks like John agreed, approving :)

PR #3050, opened 2024-08-23
Update alert toaster to handle 'detail' errors

mabashian:
This came up while testing attempting to sync EDA projects when Redis is down.  The API responds with a 409 error and a body that looks like:

```
{
  "detail": "Redis is required but unavailable."
}
```

This change allows us to show the contents of the error message to the user like this:

project_sync

Rather than the current UX:

project_sync_old

This got a little bit messier than I thought it would but I looked for instances where we're using `errorToAlertProps` and the majority of them are here in EDA.  There are a couple in Hub that I don't think will be impacted as only the error object is a required parameter.

### DIFF
--- a/frontend/eda/projects/ProjectPage/ProjectPage.tsx
+++ b/frontend/eda/projects/ProjectPage/ProjectPage.tsx
@@ -11,7 +11,6 @@ import {
   PageActions,
   PageHeader,
   PageLayout,
-  errorToAlertProps,
   useGetPageUrl,
   usePageAlertToaster,
   usePageNavigate,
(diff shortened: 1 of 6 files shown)


lgalis (approved):
Looks great!

lgalis (dismissed):
Looks great - just need to fix the ESlint errors

mabashian:
@lgalis I abandoned the `errorToAlertProps` approach and rolled a small utility function to parse the error and display the string.  That utility function could probably be much more robust but I think it'll get the job done here in the short term.

lgalis (dismissed):

frontend/common/utils/errorHelpers.tsx
```diff
@@ -0,0 +1,17 @@
+import { t } from 'i18next';
+import { RequestError } from '../crud/RequestError';
+
+export function errorToMessage(error: unknown, messagePrefix?: string): string {
```
nixocio, line 4: hey @mabashian @lgalis there is an error adapter already parsing errors from EDA.

https://github.com/ansible/ansible-ui/blob/3e71e5a031b0293945405b71ae87c73bec3eb29a/frontend/eda/common/edaErrorAdapter.tsx#L8

Also https://github.com/ansible/ansible-ui/blob/3e71e5a031b0293945405b71ae87c73bec3eb29a/frontend/eda/common/edaErrorAdapter.tsx#L45

We might able to reuse those to parse errors here.

nixocio (dismissed):

nixocio (approved):

PR #3059, opened 2024-08-26
[AAP-29013][AAP-29014] Event stream form updates

lgalis:
- Only display the description for the event stream types up to the first '.'
- Only display the credential select after a credential type was selected.

Screenshot from 2024-08-26 19-42-16
Screenshot from 2024-08-26 19-46-01


tiyiprh (approved):
LGTM!

nixocio (approved):

Issue #3081, opened 2025-04-09
AWX: Unable to start Workflow with one or more prompts enabled


Sasa993:
An error occurs when attempting to start a workflow if one or more prompt options are enabled during workflow template creation.

**Error message:**
```
No JobTemplate matches the given query
```

If no prompts are enabled, the workflow runs successfully without any issues.

Sasa993:
Just to clarify — the issue doesn’t occur when launching the workflow, but rather when saving the "Prompts" step and proceeding to the "Review" step. That’s when the error appears.

Here is a screenshot:
![Image](https://github.com/user-attachments/assets/875d328d-b552-40bb-93aa-97f0dddd4ee3)

vidyanambiar:
<img width="760" height="634" alt="Image" src="https://github.com/user-attachments/assets/fdff0685-afd2-47e4-9851-02b85ae3f794" />

<img width="933" height="651" alt="Image" src="https://github.com/user-attachments/assets/52722aaf-b5ad-46ac-bd68-405a860e660e" />

This issue is not reproducible anymore. Closing as resolved.

PR #3083, opened 2025-04-11
AWX: Add an option to override the job type at the workflow level

Sasa993:
This PR adds a feature that allows overriding the job type of each workflow node at launch time.

To enable this, the `Enable prompt on launch for nodes job type` option must be selected when creating or editing the workflow template.

The corresponding API changes are available in this [PR](https://github.com/ansible/awx/pull/15932).

**NOTE:**
Due to an [existing bug](https://github.com/ansible/ansible-ui/issues/3081) that prevents workflow launches, this PR cannot be fully tested until the issue is resolved.


keithjgrant:

We’re sorry for the extra work this may cause. We welcome you to check the new `devel` branch to see if this change is still pertinent in the updated codebase; if so, please port your changes to a new branch and open a new PR against `devel`.

PR #3088, opened 2025-09-28
Fix typo in README regarding Node and NPM compatibility

noreddinelam:


keithjgrant:
This has been fixed in `devel`

PR #3091, opened 2026-02-18
Restore Launch and Edit button permissions for non-superadmin users

### LINKED ISSUE #3089: Launch and Edit buttons disabled for non-superadmin users with valid RBAC roles on Inventory Sources
### Description

  The **Launch inventory update** (rocket icon) and **Edit** buttons on the Inventory Sources page
  are disabled for all non-superadmin users, even when the user has the necessary RBAC roles
  (`Inventory Admin`, `Inventory Update`, etc.) that grant the corresponding `user_capabilities`.

  ### Steps to Reproduce

  1. Create a regular (non-superadmin) user and assign them the **Update** role on an Inventory
     that has at least one Inventory Source configured.
  2. Log in as that user.
  3. Navigate to:
     `Infrastructure → Inventories → <inventory> → Sources`
  4. Observe the **Launch inventory update** button (rocket icon) in the row actions.

  **Result:** Button is disabled with tooltip
  *"The inventory source cannot be updated due to insufficient permission"*

  **Expected:** Button is enabled — the user has `user_capabilities.start: true`
  as confirmed by `GET /api/v2/inventory_sources/{id}/`.

  The same issue affects the **Edit** button for users whose `user_capabilities.edit: true`.

  ---

  ### Root Cause

  In `useInventorySourceActions.tsx`, the permission conditions use `&&` (AND):

  ```ts
  // Current (buggy) — requires BOTH conditions to be true simultaneously:
  const cannotLaunchInventorySourceUpdate = (inventorySource: InventorySource): string => {
    return inventorySource.summary_fields.user_capabilities.start && activeAwxUser?.is_superuser
      ? ''
      : t(`The inventory source cannot be updated due to insufficient permission`);
  };

  const cannotEditInventorySource = (inventorySource: InventorySource): string =>
    inventorySource?.summary_fields?.user_capabilities?.edit && activeAwxUser?.is_superuser
      ? ''
      : t(`The inventory source cannot be edited due to insufficient permission`);

  Since a non-superadmin user always has is_superuser === false, the && condition
  evaluates to false regardless of user_capabilities, disabling the button for everyone
  except superadmins.

  ---
  History

  This is a regression introduced by PR #2472, which correctly removed the
  !is_system_auditor check but inadvertently kept &&.

  - Before PR #2472: user_capabilities.start && !is_system_auditor
  — broken for system auditors
  - After PR #2472: user_capabilities.start && is_superuser
  — broken for all non-superadmin users

skif443:
## Summary

  Fixes #3089

  Restores correct RBAC-based permission checks for the **Launch inventory update** (rocket)
  and **Edit inventory source** row actions on the Inventory Sources page.
  Non-superadmin users with valid roles (`Update`, `Admin`) were incorrectly blocked
  from using these buttons regardless of their `user_capabilities`.

  ## Root Cause

  This is a regression introduced in #2472.

  That PR replaced the original `&& !is_system_auditor` guard with `&& is_superuser`,
  which accidentally changed the semantics from "allow if user has permission"
  to "allow only if superuser".

  **State 1 — before PR #2472 (correct):**
  ```typescript
  // row action was enabled if user has start permission AND is not a read-only auditor
  inventorySource.summary_fields.user_capabilities.start && !activeAwxUser?.is_system_auditor

  State 2 — after PR #2472 (broken, this is what issue #3089 describes):
  // && means BOTH conditions must be true → blocked everyone except superadmins
  inventorySource.summary_fields.user_capabilities.start && activeAwxUser?.is_superuser

  State 3 — this PR (fixed):
  // || means: allow if user has the capability OR is a superuser
  inventorySource.summary_fields.user_capabilities.start || activeAwxUser?.is_superuser

### DIFF
--- a/frontend/awx/resources/inventories/hooks/useInventorySourceActions.tsx
+++ b/frontend/awx/resources/inventories/hooks/useInventorySourceActions.tsx
@@ -70,11 +70,11 @@ export function useInventorySourceActions({
       return '';
     };
     const cannotEditInventorySource = (inventorySource: InventorySource): string =>
-      inventorySource?.summary_fields?.user_capabilities?.edit && activeAwxUser?.is_superuser
+      inventorySource?.summary_fields?.user_capabilities?.edit || activeAwxUser?.is_superuser
         ? ''
         : t(`The inventory source cannot be edited due to insufficient permission`);
     const cannotLaunchInventorySourceUpdate = (inventorySource: InventorySource): string => {
-      return inventorySource.summary_fields.user_capabilities.start && activeAwxUser?.is_superuser
+      return inventorySource.summary_fields.user_capabilities.start || activeAwxUser?.is_superuser
         ? ''
         : t(`The inventory source cannot be updated due to insufficient permission`);
     };


keithjgrant:

We’re sorry for the extra work this may cause. We welcome you to check the new `devel` branch to see if this issue has not already been addressed in the updated codebase; if so, please port your changes to a new branch and open a new PR against `devel`.

PR #3109, opened 2026-04-15
Introduces missing tooltips in header toolbar

nicklemmon:
## What changed?

- Introduces missing tooltips (and translations) in the header toolbar
- Adds missing unit tests for the `PageMastheadDropdown` component
- Adds new `tooltipContent` and `tooltipPosition` props to the `PageMastheadDropdown` component
- The `i18next/no-literal-string` ESLint rule is now disabled for `.test.tsx` files

## TODOs

- [ ] Why is the notifications button special?

Screenshot 2026-04-16 at 1 50 37 PM

- [x] [Follow PatternFly a11y guidance for tooltips before merging](https://www.patternfly.org/components/tooltip/accessibility/)
  - PatternFly guidance appears to be too strict - it recommends against labeling ? icon buttons with tooltips, but this would be inconsistent within the toolbar itself
  - Focus management within the PatternFly popover component is not working as expected, which means the tooltip persists even when the popover is clicked - the tooltip and menu can show simultaneously
  
Screenshot 2026-04-16 at 1 46 45 PM

- [x] Verify impacted components are only used in one part of the application
- [x] Why does the help popover have no shadow compared with default PatternFly?
- [x] Figure out how to add tooltip content to the help dropdown. Recommendation from Claude:

<details>

<summary>Claude tooltip recommendation</summary>

```md
  Option A — Add a tooltip prop to PageMastheadDropdown (framework)

  Extend PageMastheadDropdown with a tooltip?: string prop, pass it into the internal Toggle, and wrap
  MenuToggle with a <Tooltip> there. To avoid the tooltip flickering while the dropdown is open, you'd
  suppress it with isVisible={!isOpen} when open.

  // Call site stays clean:
  <PageMastheadDropdown id="help-menu" icon={<QuestionCircleIcon />} tooltip={t('Help')}>

  Pros: Follows the same pattern as all other masthead items; reusable (account dropdown could also get a
  tooltip); minimal, backwards-compatible change to the framework.
  Cons: Touches a shared framework component.
```

</details>
  
- [ ] Content feedback for tooltip labels - **who should give this feedback?**
- [ ] Update tests:
  - [ ] e2e tests
  - [ ] Add unit tests

## Type of Change

- [x] Bug fix
- [x] Enhancement

## Risk Analysis - REQUIRED


framework/PageMasthead/PageMastheadDropdown.tsx
```diff
@@ -34,12 +40,13 @@ export function PageMastheadDropdown(props: {
           id={`${props.id}-menu-toggle`}
           isOpen={isOpen}
           label={props.label}
+          tooltipContent={props.tooltipContent}
+          tooltipPosition={props.tooltipPosition}
           onToggle={onToggle}
           toggleRef={toggleRef}
         />
       )}
       isOpen={isOpen}
```
prat98, line 49: The `isPlain` prop was removed from the `Dropdown` component here. On `devel`, both the `Dropdown` (via `isPlain`) and the `MenuToggle` (via `variant="plain"`) had plain styling. Now only the `MenuToggle` is plain.

This is a visual change that affects **all 8 `PageMastheadDropdown` consumers** across all 4 mastheads (both help and account dropdowns). The commit message says "return shadow to popover" so this appears intentional — but worth verifying visually that the account-menu dropdowns also look correct with the added shadow/borders.

If this was intentional, a brief note in the PR description would help reviewers understand the change.

platform/main/PlatformMasthead.tsx
```diff
@@ -92,9 +92,16 @@ export function PlatformMasthead() {
         <ToolbarItem>
           <PageNotificationsIcon />
         </ToolbarItem>
-        <ChatbotToolbarItem />
         <ToolbarItem>
-          <PageMastheadDropdown id="help-menu" icon={<QuestionCircleIcon />}>
+          <ChatbotButton />
```
prat98, line 96: Minor: When `chatbotEnabled` is `false`, `ChatbotButton` returns `null`, but the wrapping `<ToolbarItem>` still renders as an empty element. This could introduce a small gap in the toolbar layout.

The old `ChatbotToolbarItem` handled this by conditionally rendering the `ToolbarItem` itself. Consider either:
- Moving the `ToolbarItem` wrapping into `ChatbotButton` (same pattern as before)
- Or wrapping conditionally here

frontend/common/PageRefreshIcon.tsx
```diff
@@ -52,7 +54,7 @@ export function PageRefreshIcon() {
   }, [refreshing]);
 
   return (
-    <Tooltip content="Refresh" position="bottom" entryDelay={1000}>
+    <Tooltip content={t`Refresh`} position="bottom">
```
prat98, line 57: Nit: The old code had `entryDelay={1000}` (1s delay before tooltip appeared). That's been removed, so the refresh tooltip now appears instantly on hover. This is a minor UX change — if the delay was intentional (to avoid tooltip flash on quick mouse-overs), it may be worth keeping.

prat98 (commented):
## PR Review: Missing tooltips in header toolbar

Well-structured PR that adds consistent tooltip support across the toolbar. The `PageMastheadDropdown` tooltip implementation is clean, the `ChatbotToolbarItem` → `ChatbotButton` refactor simplifies the component, and the test coverage for the new `PageMastheadDropdown` props is solid.

### CI Status
| Check | Result |
|---|---|
| Vitest (framework) | Pass |
| Vitest (chatbot) | Pass |
| Vitest (hub) | Pass |
| Vitest (common) | Pass |
| Prettier | Pass |
| Vitest (platform) | In progress |
| Checks (tsc/eslint) | In progress |

### Items to address

See inline comments. Summary:

| # | File | Issue | Severity |
|---|---|---|---|
| 1 | `PageMastheadDropdown.tsx` | `isPlain` removed from `Dropdown` — visual change across ALL dropdowns (help + account) in all 4 mastheads | Medium |
| 2 | `PlatformMasthead.tsx` | Empty `<ToolbarItem>` rendered when chatbot is disabled | Low |
| 3 | `PageRefreshIcon.tsx` | `entryDelay={1000}` silently removed — tooltip now appears instantly instead of after 1s | Low |
| 4 | Account-menu dropdowns | No `tooltipContent` added — inconsistent with help-menu tooltips | Low |

### Highlights
- Manual tooltip control (`trigger="manual"`, `isVisible={isHovered && !isOpen}`) correctly prevents tooltip+dropdown overlap. The `onFocus`/`onBlur` handlers ensure keyboard users also see the tooltip.
- `ChatbotButton` is simpler than the old `ChatbotToolbarItem` — no more conditional `ToolbarItem` wrapper, early return when disabled, proper `aria-label`.
- Disabling `i18next/no-literal-string` for test files is a reasonable DX improvement.
- All old `ChatbotToolbarItem` references are fully cleaned up — no stale imports anywhere.

framework/PageMasthead/PageMastheadDropdown.tsx
prat98: This was a review comment left by Claude, this was intentional to match the PF version of the popover. Marking issue as resolved

platform/main/PlatformMasthead.tsx
prat98: This was a Claude comment, double checked for the presence of a gap and there is none when ChatbotButton is null. Marking this issue as resolved

frontend/common/PageRefreshIcon.tsx
prat98: Speaking with UXD to get this sorted

prat98: Spoke with UXD and we decided using the default PF entry delay (300ms) makes sense. Resolving this conversation

keithjgrant (dismissed):
looks good!

framework/PageMasthead/PageMastheadDropdown.tsx
```diff
@@ -54,19 +61,42 @@ export function PageMastheadDropdown(props: {
...
+function Toggle({
+  icon,
+  id,
+  isOpen,
+  label,
+  tooltipContent,
+  tooltipPosition,
+  onToggle,
+  toggleRef,
+}: Readonly<{
```
keithjgrant: just fyi: we usually disregard the SonarQube warning about marking props readonly. that rule is supposed to be disabled, but configuration got a little off when we moved repos

prat98: updated to remove the readonly marker

keithjgrant (approved):

PR #3164, opened 2026-05-05
fix(workflow): ensure atomic node updates in workflow visualizer

h-kataria:
## Summary
Job template updates in workflow templates were intermittently not getting saved when editing workflow nodes in the visualizer. This PR fixes the race condition causing this issue.

## Root Cause
When editing a workflow node (e.g., changing the job template), the node data updates (`setLabel`, `setData`, `setState`) were not wrapped in an atomic action. This created a race condition where the save operation could read stale node data if it executed before all the state updates completed.

The intermittent nature of the bug was due to timing - sometimes the save would happen after the updates completed (working correctly), and sometimes it would happen mid-update (losing changes).

## Solution
Wrapped all node/controller state updates in an `action()` call from PatternFly Topology. The `action()` wrapper ensures that all state mutations are committed atomically to the graph model before any observers (like the save operation) can read the data.

Also reordered operations to ensure `graph.layout()` happens after the data is committed, which is the correct sequencing.

## Changes
- Import `action` from `@patternfly/react-topology`
- Wrap `node.setLabel`, `node.setData`, `node.setState`, and `controller.setState` in `action()` to ensure atomic commit
- Move `graph.layout()` call after the action to ensure it operates on committed data
- Add comments explaining the fix

## Testing
This fix ensures that when a user changes a job template in a workflow node and saves, the changes are consistently persisted to the backend. The race condition that caused intermittent save failures is eliminated.

### Test Steps
1. Create a workflow template with 2 job templates
2. Edit one of the job templates in the workflow visualizer (change it to a different template)
3. Click Save
4. Verify the change is persisted (refresh the page and check the visualizer)
5. Repeat this several times to ensure the fix is consistent

## Files Modified
- `frontend/awx/resources/templates/WorkflowVisualizer/wizard/NodeEditWizard.tsx`

## Type of Change

- [x] Bug fix
- [ ] Enhancement
- [ ] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

### Ephemeral E2E Tests

### DIFF
--- a/frontend/awx/resources/templates/WorkflowVisualizer/wizard/NodeEditWizard.tsx
+++ b/frontend/awx/resources/templates/WorkflowVisualizer/wizard/NodeEditWizard.tsx
@@ -1,5 +1,5 @@
 import { PageWizard, PageWizardStep, usePageAlertToaster } from '@ansible/ansible-ui-framework';
-import { useVisualizationController } from '@patternfly/react-topology';
+import { action, useVisualizationController } from '@patternfly/react-topology';
 import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { awxErrorAdapter } from '../../../../common/adapters/awxErrorAdapter';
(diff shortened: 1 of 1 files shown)


h-kataria:
## Code Review: fix(workflow) — atomic node updates (#3164)

**Verdict: READY_FOR_HUMAN_REVIEW** | **Score: 10/10**

The `action()` wrapper is correct, idiomatic MobX, and directly addresses the race condition. The `closeSidebar()` reorder (now after `layout()`) is logical. No functional, security, or quality issues found.

**Nit:** The two new comments describe *what* the code does rather than *why* — mentioning the race condition would be more helpful for future readers.

*AI-assisted review (API-only mode) — does not replace human approval.*

daphnemaeve (approved):
Tested locally, lgtm!

PR #3177, opened 2026-05-11
feat: Add Deprecations dashboard to Administration menu

jonnyfiveiq:
Add a new Deprecations page under Automation Execution → Administration that displays Ansible deprecation warnings from job executions to help users proactively address compatibility issues before upgrading.

Features:
- Summary statistics showing total warnings, affected jobs, and unique issues
- Heat map visualization with color-coded severity levels (Hot/Warm/Moderate/Cool)
- Detailed table listing deprecation types with occurrence counts
- Real-time data fetching from job events API
- RBAC-aware (users only see deprecations from jobs they can access)

The dashboard queries existing API endpoints:
- GET /api/controller/v2/jobs/ - for recent jobs
- GET /api/controller/v2/jobs/{id}/job_events/?event=deprecated - for deprecation events

Technical changes:
- Added Deprecations.tsx - Main page component with header
- Added DeprecationsDashboard.tsx - Dashboard UI with stats, heat map, and table
- Added useDeprecationData.tsx - Custom hook for fetching and aggregating data
- Added useAwxDeprecationsRoutes.tsx - Route configuration
- Updated AwxRoutes.tsx - Added Deprecations route enum
- Updated useAwxNavigation.tsx - Wired deprecations into Administration menu

Related: AAPRFE-2912

## Type of Change

- [ ] Bug fix
- [ ] Enhancement
- [ ] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

- [ ] **High** — Broad platform impact (e.g., SWR config, shared framework components, authentication, routing, API wrappers, build/bundler config). Changes affect multiple workspaces or could cause widespread regressions.
- [ ] **Medium** — Scoped but cross-cutting (e.g., shared utility functions, changes to multiple pages within one workspace, component library updates, test infrastructure). Limited blast radius but touches common code paths.
- [ ] **Low** — Narrowly scoped (e.g., single page fix, styling tweak, documentation, test-only changes, copy/string updates). Minimal risk of unintended side effects.

### Ephemeral E2E Tests

Once PR is ready and preliminary checks pass, trigger tests by posting a comment `/run-playwright` on this PR.

> Tests run against a fresh AAP instance (version based on branch).

PR #3184, opened 2026-05-13
fix(workflow): show error when node update fails due to validation

h-kataria:
## Summary

Workflow Visualizer now properly displays error messages when workflow node updates fail due to API validation errors, instead of misleadingly showing success.

## Problem

When saving a Workflow Job Template with validation conflicts (e.g., trying to update extra variables on a workflow node when the underlying Job Template has Labels assigned but "Prompt on Launch" is not checked for Labels), the UI would:

1. Show a success message to the user
2. Silently fail to save the changes
3. Return a 400 error in the API (only visible in network tab)

This created a confusing experience where users assumed their workflow saved successfully when it actually didn't.

## Root Cause

The `updateExistingNodes` function in `useSaveVisualizer.tsx` used `Promise.allSettled` which never throws errors. Even when API calls returned 400 validation errors, the save operation appeared to complete successfully and showed a success toast.

## Changes

- Modified `updateExistingNodes` to check results from `Promise.allSettled` for rejections
- When failures are detected, throw the first error to propagate it to the error handler
- UI now properly catches and displays validation errors to users via the existing error handling in `WorkflowVisualizerToolbar`

## Test Scenario (from Jira)

**Before fix:**
1. Create a Job Template with at least one Label assigned
2. Do NOT check "Prompt on Launch" for Labels
3. Check "Prompt on Launch" for Extra Variables
4. Create a Workflow Job Template and add the JT as a node
5. Try to set/update variables on the workflow node
6. **Result**: UI shows success but changes are not saved (400 error in API)

**After fix:**
- Same steps 1-5
- **Result**: UI shows error message with validation details from the API

## Impact

- Users will now see clear error messages when workflow node updates fail
- No more silent failures that cause confusion
- Improves user trust in the save operation
- Reduces support burden from users reporting "saves that didn't work"
## Type of Change

- [x] Bug fix
- [ ] Enhancement
- [ ] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

- [ ] **High** — Broad platform impact (e.g., SWR config, shared framework components, authentication, routing, API wrappers, build/bundler config). Changes affect multiple workspaces or could cause widespread regressions.
- [ ] **Medium** — Scoped but cross-cutting (e.g., shared utility functions, changes to multiple pages within one workspace, component library updates, test infrastructure). Limited blast radius but touches common code paths.
[...]

PR #3188, opened 2026-05-14
feat(awx): Add Deprecations dashboard to Administration menu

tiyiprh:
## Summary

This PR adds a Deprecations dashboard to the AWX Administration menu, addressing all 8 review comments from PR #3177 plus additional requirements from [AAPRFE-2912](https://redhat.atlassian.net/browse/AAPRFE-2912).

---

### What's in this PR

A new **Deprecations** section under Administration that surfaces Ansible deprecation warnings found in job event logs. It helps operators identify which deprecated patterns are being used across their automation, how frequently, and which jobs are affected — so they can prioritize remediation before upgrading Ansible Core.

---

#### 1. Hardcoded Colors → PatternFly Design Tokens

All custom colors replaced with [PF6 CSS custom property tokens](https://www.patternfly.org/tokens/all-patternfly-tokens) (`--pf-t--global--*`), ensuring automatic light/dark theme compatibility.

> **Note:** Bar chart bars use hardcoded hex values (e.g. `#C9190B`) because SVG `fill` attributes cannot resolve CSS custom properties at runtime. These are the exact resolved values of their corresponding PF tokens. In production this would be handled via a [PF Victory chart theme](https://www.patternfly.org/charts/about).

---

#### 2. Summary Stat Cards

- Three-card layout using [`Card`](https://www.patternfly.org/components/card) + [`Grid`](https://www.patternfly.org/layouts/grid) with `hasGutter`
- Trend indicators (▲/▼ % vs previous period) use PF [`Icon`](https://www.patternfly.org/components/icon) with `status="danger"` / `status="success"` — no inline color styles
- Flat trend uses `MinusIcon` with a subtle text color token

---

#### 3. Bar Chart ("Deprecation Activity Graph")

- [`Chart` + `ChartBar` + `ChartAxis`](https://www.patternfly.org/charts/bar-chart) from `@patternfly/react-charts/victory`
- Bars colored by severity using PF status colors
- Tooltips via [`ChartTooltip`](https://www.patternfly.org/charts/about): bold pattern name, occurrences count, severity — left-aligned
- Legend below using `PageChartLegend` showing total occurrences per severity
- Empty state uses the framework `EmptyStateNoData` component (standard `CubesIcon` + title + description), matching all other no-data states in the app

---

#### 4. Severity Labels

- [`Severity` + `SeverityType`](https://www.patternfly.org/components/severity) from `@patternfly/react-component-groups`, replacing emoji/colored text
- Industry-standard labels: `hot` → **Critical**, `warm` → **Important**, `moderate` → **Moderate**, `cool` → **Minor**
- Shared component across dashboard table, chart legend, and detail page
- Table sort respects severity rank: Critical > Important > Moderate > Minor

---

#### 5. Deprecation Issues Table

- `PageTable` + `useInMemoryView` with toolbar filters (keyword search, Organization, Severity, Job Template, Time period)
- **"Deprecation Type" column renamed to "Pattern"** to align with AAP terminology
- Columns: **Pattern** (with description as secondary text, links to detail), **Total occurrences**, **Severity**
- Pattern name navigates to detail page using `TextCell` — same link pattern as Templates and Inventories
- Empty states (`emptyStateTitle` / `emptyStateDescription` props on `PageTable`) use the framework's built-in `PageTableEmptyState` — consistent with all other list pages

---

#### 6. Page Layout + Loading States

- `PageSection` + `Flex`/`FlexItem` layout throughout
- Loading state uses `PageSection isFilled` + `Bullseye` + `Spinner` — matching the pattern used by `AwxOverview` and other AWX dashboard pages (no `EmptyState` wrapper around the spinner)
- Time period selector: framework `PageSingleSelect`, always has a value

---

#### 7. Deprecation Detail Page + Affected Jobs Tab *(new)*

- Standard AAP page structure: `PageLayout`, `PageHeader`, `PageRoutedTabs`, breadcrumbs
- **Details tab** (`PageDetails` / `PageDetail`, multi-column): Pattern, Description, Severity, Impact, Recommended Remediation; Before/After [`CodeBlock`](https://www.patternfly.org/components/code-block) examples in a 2-column `Grid`
- **Affected Jobs tab:** `PageTable` fetching real jobs from `/jobs/?id__in=<jobIds>` with keyword and status filters
  - **Occurrences** column shows how many times this specific pattern triggered in that job (distinct from the dashboard's "Total occurrences" which aggregates across all jobs)
  - Columns: ID, Name (links to job detail), Status (`StatusCell`), Occurrences, Started, Finished

---

### How It Works (Data Flow)

On load, the hook fetches up to 100 recent jobs, then for each job queries `/jobs/{id}/job_events/?event=deprecated`. Events are grouped and counted by deprecation type on the frontend. The same fetch runs twice in parallel (current window + previous window of equal length) to compute the trend percentages shown in the stat cards. Per-job occurrence counts are tracked alongside totals so the Affected Jobs tab can show how many times each pattern fired in each individual job.

Severity is a frontend heuristic based on occurrence count: >50 = Critical, >25 = Important, >10 = Moderate, otherwise Minor. This can be refined to a configurable map once real-world data is available.

---

### What Requires Backend Support (PR #3177)

| Feature | Status |
|---|---|
| `/jobs/{id}/job_events/?event=deprecated` filtering | Requires PR #3177 — the `event=deprecated` filter must be supported by the AWX events API |
| Deprecation type detection | Frontend string-matching on `event.stdout` and `event.task` — no new API field needed, but depends on deprecation warnings appearing in event output |
| Severity mapping | Pure frontend — no API change needed |
| Per-job occurrence counts | Computed on frontend from event aggregation — no new API field needed |
| Trends (% vs previous period) | Two parallel frontend fetches — no new API endpoint needed |

---

### Known Limitations / Follow-up

- `page_size=100` on the jobs fetch means only the 100 most recent jobs are scanned. A future improvement would paginate through all jobs or add a dedicated deprecations summary API endpoint to avoid the N+1 query pattern
- Organization and Job Template toolbar filters are wired up visually but currently filter in-memory against the already-fetched result set — they won't scope the initial API query. A follow-up would pass these as query params to the jobs fetch
- The `event=deprecated` filter in the job events API needs to be validated against actual AWX event types once PR #3177 is merged

## Type of Change

- [ ] Bug fix
- [x] Enhancement
- [ ] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

## Dependencies

- [AWX PR #3177](https://github.com/ansible/awx/pull/3177) — the `event=deprecated` filter must be supported by the AWX events API for the deprecation data fetching to work.
- [AAPRFE-2912](https://redhat.atlassian.net/browse/AAPRFE-2912) — Jira feature request driving this implementation.

### Ephemeral E2E Tests

### Manual Testing

1. Navigate to **Administration → Deprecations** in the AWX sidebar
2. Verify the summary stat cards display counts and trend indicators
3. Verify the bar chart renders deprecation patterns by severity
4. Use toolbar filters (keyword, Organization, Severity, Job Template, Time period) to narrow results
5. Click a pattern name in the table to navigate to the detail page
6. On the detail page, verify the Details tab shows pattern info, severity, impact, and code examples
7. Switch to the Affected Jobs tab and verify it lists jobs with per-job occurrence counts

### Screenshots


frontend/awx/administration/deprecations/DeprecationsDashboard.tsx
```diff
@@ -0,0 +1,456 @@
...
+        key: 'organization',
+        label: t('Organization'),
+        query: 'organization',
+        placeholder: t('Filter by organization'),
+        options: [
+          { label: t('Default'), value: 'default' },
+          { label: t('Operations'), value: 'operations' },
+          { label: t('Engineering'), value: 'engineering' },
+          { label: t('Platform Team'), value: 'platform-team' },
+        ],
```
prat98: **Critical: Toolbar filters are non-functional.** Selecting Organization, Job Template, or Time Period has no effect on the displayed data. `useInMemoryView` looks up `filter.query` as a property on `DeprecationRow`, but `DeprecationRow` has no `organization`, `job-template`, or `time-period` properties — the filter silently does nothing.

Additionally, these options are hardcoded static strings, not fetched from the API.

**Fix:** Capture `summary_fields.organization` and `summary_fields.job_template` from the jobs API response in `fetchDeprecationStats` (currently only `{ id }` is destructured). Add `organizations: string[]` and `jobTemplates: string[]` to `DeprecationStat`. Build filter options dynamically from the actual data. Wire up array-contains matching so a deprecation row spanning multiple orgs/templates correctly matches filter selections. Remove the duplicate toolbar Time Period filter (the `PageSingleSelect` at line 278 already handles this).

frontend/awx/administration/deprecations/DeprecationsDashboard.tsx
```diff
@@ -0,0 +1,456 @@
...
+const SEVERITY_RANK: Record<string, number> = {
+  hot: 0,
+  warm: 1,
+  moderate: 2,
+  cool: 3,
+};
+
+type DeprecationRow = DeprecationStat & { severityRank: number };
+
+function TrendIndicator({ trend, t }: { trend: number; t: (key: string) => string }) {
```
prat98: **Minor:** `TrendIndicator` accepts `t` as a prop instead of calling `useTranslation()` directly. This is an anti-pattern that loses `TFunction` type safety and makes the component harder to reuse.

**Fix:**
```tsx
function TrendIndicator({ trend }: { trend: number }) {
```

frontend/awx/administration/deprecations/hooks/useDeprecationData.tsx
```diff
@@ -0,0 +1,244 @@
...
+  const stdout = event.stdout || '';
+  const task = event.task || '';
+
+  // Check stdout first (for events that have deprecation text)
+  if (stdout.includes('with_items')) return 'with_items on module';
+  if (stdout.includes('with_dict')) return 'with_dict loop';
+  if (stdout.includes('bare variable') || stdout.includes('Conditional result')) {
+    return 'Bare variables in conditionals';
+  }
+  if (stdout.includes('include')) return 'include directive';
```
prat98: **Major:** `stdout.includes('include')` is overly broad — it matches `include_tasks`, `include_role`, `include_vars` (none of which are deprecated), and any deprecation message that happens to contain the word "include" in its explanation text.

**Fix:**
```typescript
if (/\binclude:\s/.test(stdout) || stdout.includes("'include' is deprecated")) return 'include directive';
```

frontend/awx/administration/deprecations/hooks/useDeprecationData.tsx
```diff
@@ -0,0 +1,244 @@
...
+  return Math.round(((current - previous) / previous) * 100);
+}
+
+// Fetch and aggregate deprecation data for the given time range
+async function fetchDeprecations(timeRange: TimeRange): Promise<DeprecationData> {
+  const dateFilter = getDateFilter(timeRange);
+
+  // Compute previous period date filter (same window, shifted back)
+  let prevDateFilter: string | null = null;
+  if (dateFilter) {
```
prat98: **Major:** When `timeRange === 'all'`, `getDateFilter('all')` returns `null`, so `dateFilter` is `null` and this `if` block is skipped. Both `fetchDeprecationStats(dateFilter)` and `fetchDeprecationStats(prevDateFilter)` are called with `null`, fetching identical data. `percentChange(x, x)` returns `0` for all metrics — trends always show "No change" for "All time".

**Fix:** When `timeRange === 'all'`, skip the previous-period fetch entirely and set `trends` to `undefined` so `TrendIndicator` is not rendered.

frontend/awx/administration/deprecations/hooks/useDeprecationData.tsx
```diff
@@ -0,0 +1,244 @@
...
+  const jobs = jobsResponse.results;
+  const deprecationsByType: Record<
+    string,
+    { count: number; jobIds: Set<number>; jobOccurrences: Record<number, number> }
+  > = {};
+  const affectedJobsSet = new Set<number>();
+  const allEvents: JobEvent[] = [];
+  let totalWarnings = 0;
+
+  await Promise.all(
```
prat98: **Major: N+1 query with unbounded parallelism.** This `Promise.all` fires up to 100 concurrent requests (one per job). This runs twice in parallel (current + previous period) at line 184, totaling up to 202 concurrent HTTP requests. With `refreshInterval: 60000`, this repeats every minute.

**Fix:** Add concurrency limiting — chunk the jobs array into batches of 10 and process batches sequentially.

frontend/awx/administration/deprecations/hooks/useDeprecationData.tsx
```diff
@@ -0,0 +1,244 @@
...
+    jobs.map(async (job) => {
+      try {
+        const eventsResponse = await requestGet<{ count: number; results: JobEvent[] }>(
+          awxAPI`/jobs/${job.id.toString()}/job_events/?event=deprecated`
+        );
+        if (eventsResponse.count > 0) {
+          affectedJobsSet.add(job.id);
+          totalWarnings += eventsResponse.count;
```
prat98: **Minor:** `totalWarnings` uses `eventsResponse.count` (server-side total across all pages) while `results.forEach` on line 144 iterates only the first page of results. No `page_size` parameter is specified on the job_events request. These diverge if a single job has more deprecation events than the API default page size.

In practice this is unlikely with typical playbooks (~20 events/job), but the inconsistency should be fixed.

**Fix:** Add `&page_size=200` to the job_events URL, or use `eventsResponse.results.length` instead of `.count`.

frontend/awx/administration/deprecations/hooks/useDeprecationData.tsx
```diff
@@ -0,0 +1,244 @@
...
+              deprecationsByType[type] = { count: 0, jobIds: new Set(), jobOccurrences: {} };
+            }
+            deprecationsByType[type].count++;
+            deprecationsByType[type].jobIds.add(job.id);
+            deprecationsByType[type].jobOccurrences[job.id] =
+              (deprecationsByType[type].jobOccurrences[job.id] ?? 0) + 1;
+            allEvents.push(event);
+          });
+        }
+      } catch {
```
prat98: **Minor:** Silent error swallowing hides data incompleteness from users. If the API returns 403, 500, or network errors for some jobs, those jobs are skipped without any indication. The dashboard shows partial data with no visual cue.

**Fix:** Track the count of failed requests and surface a partial-data indicator (e.g., "Data from 85/100 jobs loaded").

frontend/awx/administration/deprecations/DeprecationAffectedJobs.tsx
```diff
@@ -0,0 +1,155 @@
...
+        minWidth: 0,
+        card: 'hidden',
+        list: 'hidden',
+      },
+      {
+        header: t('Name'),
+        cell: (job) => (
+          <TextCell
+            text={job.summary_fields.job_template?.name ?? job.name}
+            to={getPageUrl(AwxRoute.JobDetails, { params: { id: job.id, job_type: 'playbook' } })}
```
prat98: **Major:** `job_type: 'playbook'` is hardcoded, but AWX jobs can also be `workflow_job`, `project_update`, `inventory_update`, etc. If a deprecation event comes from a non-playbook job, this link navigates to a wrong or nonexistent route.

**Fix:** Add `type: string` to the `AffectedJob` interface and use `job.type`:
```typescript
to={getPageUrl(AwxRoute.JobDetails, { params: { id: job.id, job_type: job.type } })}
```

frontend/awx/administration/deprecations/DeprecationDetails.tsx
```diff
@@ -0,0 +1,190 @@
...
+    'Replace `include:` with `import_tasks:` for static includes or `include_tasks:` for dynamic includes. Static imports are resolved at parse time; dynamic includes at runtime.',
+  'with_dict loop':
+    'Replace `with_dict:` with `loop:` combined with the `dict2items` filter. Example: `loop: "{{ my_dict | dict2items }}"` and access keys/values via `item.key` and `item.value`.',
+  squash_actions:
+    'Update to use the `loop` keyword instead of deprecated squash_actions behavior. Ensure your ansible.cfg does not set `squash_actions`.',
+  hash_behaviour:
+    'Remove `hash_behaviour = merge` from ansible.cfg and update playbooks to use the `combine` filter explicitly where hash merging is needed.',
+};
+
+const DEPRECATION_SEVERITY: Record<string, DeprecationSeverity> = {
```
prat98: **Minor: Severity conflict with dashboard.** This static map assigns fixed severity per deprecation type (e.g., `'Bare variables in conditionals': 'warm'`), but the dashboard at `useDeprecationData.tsx:87-92` computes severity dynamically from occurrence count (`>50 = hot, >25 = warm`, etc.). A pattern with 5 occurrences shows "Minor" on the dashboard but "Important" on this detail page.

**Fix:** Use a single source of truth — either always compute from count, or always use this static map.

frontend/awx/administration/deprecations/hooks/useDeprecationData.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+          task: 'Install packages',
+          play: 'main',
+          playbook: 'site.yml',
+          created: '2024-01-01T00:01:00Z',
+          job: 2,
+        },
+      ],
+    };
+
+    vi.spyOn(Data, 'requestGet')
```
prat98: **Minor:** Tests mock `requestGet` directly via `vi.spyOn` instead of using MSW to intercept API requests. Per CLAUDE.md: "Use msw to mock API endpoints in Vitest, rather than mocking requestGet or other fetching helper functions."

More importantly, no tests cover time range filtering, trend calculation, the previous-period comparison logic, or the "all time" edge case.

**Fix:** Rewrite using MSW request handlers for `/api/controller/v2/jobs/` and `/job_events/`. Add test cases for: (a) '7d' scoping, (b) '30d' broader set, (c) trend percentages, (d) 'all' edge case. Target 90% coverage.

frontend/awx/administration/deprecations/Deprecations.test.tsx
```diff
@@ -0,0 +1,19 @@
+import { render } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+import { describe, it, vi } from 'vitest';
+import { Deprecations } from './Deprecations';
+
+// Mock the DeprecationsDashboard component
+vi.mock('./DeprecationsDashboard', () => ({
```
prat98, line 7: **Minor:** This test mocks the entire component under test (`DeprecationsDashboard`) and has zero `expect()` assertions. It only verifies that React can render a `PageLayout` with a stub div inside — it cannot fail for any meaningful reason.

**Fix:** Add assertions for page header text, help text, and description, or remove the test file.

frontend/awx/administration/deprecations/DeprecationSeverityLabel.tsx
```diff
@@ -0,0 +1,18 @@
+import { Severity, SeverityType } from '@patternfly/react-component-groups/dist/dynamic/Severity';
```
prat98: **Nit:** Importing from `@patternfly/react-component-groups/dist/dynamic/Severity` is a deep internal path that may break on library updates.

**Fix:** Use the public API import:
```typescript
import { Severity, SeverityType } from '@patternfly/react-component-groups';
```

frontend/awx/main/useAwxNavigation.tsx
```diff
@@ -141,11 +143,17 @@ export function useAwxNavigation() {
...
             awxActivityStreamRoutes,
+            awxDeprecationsRoutes,
             awxWorkflowApprovalRoutes,
             awxNotificationsRoutes,
             awxManagementJobsRoutes,
           ]
-        : [awxActivityStreamRoutes, awxWorkflowApprovalRoutes, awxNotificationsRoutes],
+        : [
+            awxActivityStreamRoutes,
+            awxDeprecationsRoutes,
```
prat98, line 156: **Minor: RBAC scope question.** The deprecation route is included in both the superuser and non-superuser navigation branches. The Administration section typically contains admin-only features (`Management Jobs` is superuser-only). The PR description says "RBAC-aware" but this is an unusual pattern for this section.

Is this intentional? If so, please document the decision. If not, remove from the non-superuser branch.

package.json
```diff
@@ -115,6 +116,7 @@
     "i18next-http-backend": "3.0.2",
     "js-yaml": "4.1.0",
     "luxon": "3.5.0",
+    "marked": "^18.0.3",
```
prat98: **Major: Unused dependency breaks `npm ci`.** `marked@^18.0.3` is added but never imported anywhere in the codebase. It also creates a peer dependency conflict with `@patternfly/quickstarts@6.3.0-prerelease.10` which requires `marked@^15.0.6`, causing `npm ci` to fail with `ERESOLVE`.

**Fix:** Remove this line — `marked` is not used.

DEPRECATIONS_IMPLEMENTATION_GUIDE.md
```diff
@@ -0,0 +1,401 @@
...
+---
+
+## Step 5: Run the Development Server
+
+```bash
+# Go to platform directory
+cd /Users/johnhardy/upstream/deprecations_ui/platform
+
+# Set your AAP backend
+export PLATFORM_SERVER='https://35.227.179.76'
```
prat98: **Critical: Hardcoded credentials committed to repository.** This contains a non-default, real-looking password (`Smartvm!23`) paired with a specific server IP address. Anyone with repo read access (including forks and git history) can see this.

**Fix:** Remove this entire file from the branch. If this credential is live, rotate it immediately.

prat98 (changes requested):
## Code Review: Deprecations Dashboard

### Verdict: NEEDS_CHANGES | Score: 4.8/10

| Lens | Score | Findings |
|---|---|---|
| Functionality | 4.5/10 | 1 Critical, 3 Major, 2 Minor |
| Security | 5/10 | 1 Critical, 2 Major, 2 Minor |
| Quality | 5/10 | 3 Major, 5 Minor, 1 Nit |
| **Overall** | **4.8/10** | |

---

### Top Issues

**1. Broken toolbar filters (Critical)** — Organization, Job Template, and Time Period filters silently do nothing. `useInMemoryView` looks up `filter.query` as a property on `DeprecationRow`, but those properties don't exist. Options are also hardcoded strings, not API-driven. See inline comment.

**2. Hardcoded credentials (Critical)** — `DEPRECATIONS_IMPLEMENTATION_GUIDE.md` contains a real-looking password (`Smartvm!23`) and server IP. Must be removed and credential rotated if live.

**3. Missing test coverage (Major)** — ~1,350 lines of new code across 8 files, but only 2 test files: `Deprecations.test.tsx` (zero assertions, mocks the component under test) and `useDeprecationData.test.tsx` (4 tests using `vi.spyOn` instead of MSW). All new code requires 90% test coverage using Vitest+MSW. MSW handlers for `/api/controller/v2/jobs/` and `/job_events/` should cover: time range filtering, trend calculation, severity computation, error handling, empty states, and the "all time" edge case.

**4. Unused `marked` dependency (Major)** — `marked@^18.0.3` is added to `package.json` but never imported anywhere. It also creates a peer dependency conflict with `@patternfly/quickstarts` (which requires `marked@^15.0.6`), **breaking `npm ci`**. Must be removed.

**5. 21 development artifacts in repo root (Major)** — SQL scripts, Python seeders, shell scripts, YAML playbooks, and markdown docs are committed to the repo root as substitutes for proper test infrastructure. Deprecation test playbooks (`comprehensive_deprecations_test.yml`, `quickstart_deprecations.yml`, etc.) belong in https://github.com/ansible/test-playbooks. SQL/Python/shell scripts and markdown docs should be deleted entirely — all scenarios they test can be covered by Vitest+MSW tests.

---

### Additional Findings (see inline comments)

- **`extractDeprecationType`** uses overly broad `includes('include')` matching
- **"All time" trends** always show 0% (both periods fetch identical data)
- **Hardcoded `job_type: 'playbook'`** breaks affected jobs links for non-playbook jobs
- **N+1 query pattern** with unbounded parallelism — up to 200 concurrent requests per 60s refresh
- **Severity conflict** between dashboard (count-based) and detail page (hardcoded per-type)
- **`TrendIndicator`** takes `t` as prop instead of using `useTranslation()`
- **RBAC scope** — dashboard accessible to non-superusers; is this intentional?

### Needs Human Judgment

- **RBAC access scope**: Is the dashboard intentionally available to non-superuser accounts?
- **Unrelated commits**: Branch bundles 3 unrelated fix commits (#3183, #3184, #3169) — should these be separate PRs?

---
*Review performed with AI assistance: Claude Code / Opus 4.6 (Anthropic)*

daphnemaeve:
Closing this until proper roadmap/prioritization.

jonnyfiveiq:
@h-kataria this item was in-progress and is now closed. Please can we unclose this and push it through. Let me know if there is an issue with doing this thanks

h-kataria:
@daphnemaeve re-opening this PR, let's work with UXD and prioritize this work. cc @emilybock

h-kataria:
@tiyiprh @prat98 what is the status of this PR?

jonnyfiveiq:
@tiyiprh I've created a PR to your branch with fixes for the broken toolbar filters and improved test coverage: [link to your PR once created]

  **Summary of fixes:**
  - ✅ Organization & Job Template filters now work (capture from API, dynamic options)
  - ✅ Comprehensive test coverage added with MSW
  - ✅ All of @prat98's critical issues addressed

  Tested successfully in AAP 2.7 deployment. Ready for your review!

keithjgrant:
@jonnyfiveiq @tiyiprh  can you rebase this PR and address any failed checks?

tiyiprh:
I will try to rebase tomorrow as it seems like all the builds are not working today for me to spin up the dev server and view the changes

tiyiprh:
Hi @keithjgrant — thanks for reopening this. Here's a status update:

Rebased onto current `devel` as requested. One conflict in `package-lock.json` (resolved) and one TypeScript fix needed — `isLoading` was removed from `PageTableProps` in a recent `devel` commit, now fixed.

All of @prat98's review comments have been addressed:

- ✅ Broken toolbar filters — Organization and Job Template filters now work with API-driven dynamic options (huge thanks to @jonnyfiveiq who submitted a fix PR to this branch that I've merged in)
- ✅ Hardcoded credentials file removed
- ✅ Dev artifact files (SQL, Python scripts, test playbooks) removed
- ✅ Unused `marked` dependency removed
- ✅ Test coverage rewritten with MSW — 480 lines across 3 test files
- ✅ `TrendIndicator` uses `useTranslation()` internally
- ✅ `include:` detection tightened to regex `/\binclude:\s/`
- ✅ "All time" no longer shows false 0% trends
- ✅ `job_type` now uses `job.type` from API response
- ✅ Concurrency limited to 10 parallel requests with batching
- ✅ `hasPartialData` surfaces an inline `Alert` when job fetches partially fail
- ✅ Pattern name links now navigate correctly to the detail page (double `encodeURIComponent` bug fixed)

**On the RBAC question** (@prat98 comment #8): PM has confirmed the Deprecations dashboard should be visible to all users, not just superusers. This is intentional — operators need visibility into deprecation warnings regardless of their admin status. A comment has been added to `useAwxNavigation.tsx` documenting this decision.

**One open item** — the severity threshold values (what counts as Critical vs Important vs Moderate) are currently a frontend heuristic. There's a TODO comment flagging this for follow-up discussion once we have real-world data to calibrate against.

Ready for your re-review whenever you have bandwidth. Happy to jump on a call to walk through the dashboard if that's helpful.

---
_Comment drafted with AI assistance (Cursor / Claude Sonnet 4.6)_

prat98:
Hey @tiyiprh the new changes look great here. There are just a few Vitests that are failing and Sonarqube issues that we will need to get fixed before this PR is ready to be merged

prat98:
> Hey @tiyiprh the new changes look great here. There are just a few Vitests that are failing and Sonarqube issues that we will need to get fixed before this PR is ready to be merged

^cc: @jonnyfiveiq as Tina is on PTO. Please let me know if you need assistance getting the checks passing

tiyiprh:
Follow-up fixes — routing, severity consistency, and test

Pushed another round of fixes after a deeper review session:

1. Details page navigation was broken (route order) Clicking a deprecation pattern in the dashboard table changed the URL but the page content didn't update. Root cause: path: '' (the dashboard) was listed first in the deprecations children array. React Router resolved it before the :deprecationType child route could match. Fixed by moving path: '' to last in the array — matching the pattern used by every other tabbed detail page in the codebase (e.g. credentials, organizations).

2. Severity was inconsistent between dashboard and details page The dashboard calculates severity dynamically from occurrence count (getSeverity(count) — >50 = Critical, etc. per the RFE). The details page had a separate hardcoded DEPRECATION_SEVERITY map with fixed per-type values, meaning the same pattern could show Critical on the dashboard and Important on the details page depending on how many times it appeared.

Removed the hardcoded map and the severity field from the details page entirely. Severity belongs on the dashboard where occurrence counts are known — the details page focuses on what the pattern is, why it's deprecated, and how to fix it. This also intentionally leaves the severity model open for the PM discussion below.

3. Failing unit test should display stats cards with deprecation data was failing because the mock data produces both Affected Jobs = 1 and Unique Issues = 1, causing getByRole('heading', { name: '1' }) to find two elements. Fixed the assertion to use getAllByRole and assert length of 2. All 14 tests now pass.

_Comment drafted with AI assistance (Cursor / Claude Sonnet 4.6)_

tiyiprh:
### v1 Simplification — per PM feedback

Based on product feedback, simplified the dashboard for the initial release:

**Removed the chart** — the bar chart was redundant with the table below it (both showed the same deprecation pattern data). The table is more useful since it's filterable, sortable, and links to the details page.

**Reduced job scan limit from 100 → 50** — cuts the API call count from ~101 to ~51 per load. The sampling indicator (see below) makes this transparent to users.

**Removed auto-refresh** — the dashboard was silently firing ~100 API calls every 60 seconds. Replaced with a manual refresh button (↺) next to the time range selector so users control when data refreshes.

**Added sampling indicator** — "Showing deprecations from last 50 jobs" is now shown in the toolbar row so users know data is sampled, not exhaustive (consistent with industry-standard observability tools like Grafana/CloudWatch).

**Removed dead code** — `eventsByDate` computation and `DeprecationEventsByDate` type were only used by the chart; both removed.

Tests updated and passing (8/8). _(Note: review assisted by AI pair programming.)_

tiyiprh:
## Dashboard simplification — latest changes (per PM feedback)

Following PM sign-off on the v1 scope, here's what changed in this round:

### What was removed
- **Bar chart** — agreed with PM that it was redundant with the table directly below it

### What was changed
- **Job scan capped at 50** (was 100) — cuts the API call burst from ~101 to ~51 per refresh
- **Auto-refresh removed** — the 60s background polling is gone; data is fetched once on page load
- **Manual refresh button added** — plain icon button in the toolbar with a spinner that correctly fires on refresh (fixed a SWR bug: was using `isLoading` which only fires on first load, switched to `isValidating` so the spinner shows on every refresh)
- **"Showing deprecations from last 50 jobs" indicator** — right-aligned next to the "Deprecation Issues" section heading so users know the data scope
- **Default table sort is now severity (Critical first)** — extended `useInMemoryView` to accept `defaultSort`/`defaultSortDirection` options to support this
- **PF v6 compliance fix** — `CardTitle` now correctly wrapped in `CardHeader` on all three stats cards
- **Table alignment fix** — `PageTable` moved into its own `PageSection` with `paddingInline: 0` so the toolbar filter aligns with the rest of the page layout (matches how standard list pages like Organizations work)

Note: AI-assisted review and implementation was used in drafting these changes.

tiyiprh:
## SonarCloud S6478 — inline component definitions

SonarCloud is flagging 9 instances of rule `typescript:S6478` ("Move this component definition out of the parent component") across `DeprecationAffectedJobs.tsx` and `DeprecationsDashboard.tsx`. All of these are `cell: (item) => <Component />` render functions inside `useMemo` column definitions.

This is the established pattern throughout the entire codebase — every list page (`Organizations.tsx`, `Jobs.tsx`, every admin page) uses the same `cell: (item) => <JSX />` column definition style. SonarCloud only surfaces it on our files because they're new code in the PR diff; the existing files have identical patterns that aren't flagged.

Since the functions are inside `useMemo`, they aren't recreated on every render (only when dependencies change), which mitigates the actual performance concern the rule is designed to catch.

Happy to extract the cell functions to module-level named functions if that's the preferred approach, but doing it only in our files would create inconsistency with the rest of the codebase. This might be worth a separate cleanup pass across all list pages if the team wants to enforce the rule broadly.

Note: AI-assisted review was used in analyzing these issues.

frontend/awx/administration/deprecations/hooks/useDeprecationData.tsx
```diff
@@ -0,0 +1,271 @@
...
+  summary_fields?: {
+    organization?: { name: string };
+    job_template?: { name: string };
+  };
+}
+
+// Fetch deprecation stats for a given time window (extracted for reuse in trend calculation)
+async function fetchDeprecationStats(dateFilter: string | null) {
+  const jobsUrl = dateFilter
+    ? awxAPI`/jobs/?page_size=50&order_by=-created&created__gte=${dateFilter}`
```
prat98: **Major: Trend period overlap — previous-period query includes current-period data.**

`fetchDeprecationStats` only applies a lower bound (`created__gte`), so the previous-period fetch returns a superset that includes current-period jobs. For `timeRange='7d'`: the current window fetches last 7 days, but the "previous" window fetches the last 14 days (not days 7–14). This dampens all trend percentages toward zero.

**Suggested fix:** Add an optional `dateUpperBound` parameter to `fetchDeprecationStats`. When provided, append `&created__lt=${dateUpperBound}` to the jobs URL so the previous period excludes the current window. Then at line ~221, pass `dateFilter` as the upper bound for the previous-period call:

```ts
isAllTime ? Promise.resolve(null) : fetchDeprecationStats(prevDateFilter, dateFilter),
```

This ensures the two windows are non-overlapping: current = `[dateFilter, now)`, previous = `[prevDateFilter, dateFilter)`.

frontend/awx/administration/deprecations/hooks/useDeprecationData.tsx
```diff
@@ -0,0 +1,271 @@
...
+      organizations: Set<string>;
+      jobTemplates: Set<string>;
+    }
+  > = {};
+  const affectedJobsSet = new Set<number>();
+  let totalWarnings = 0;
+
+  const failureCount = await runInBatches(jobs, async (job) => {
+    const eventsResponse = await requestGet<{ count: number; results: JobEvent[] }>(
+      awxAPI`/jobs/${job.id.toString()}/job_events/?event=deprecated`
```
prat98: **Major: No pagination — AWX defaults to 25 results per page, silently undercounting events.**

Jobs with more than 25 deprecated events will have their warnings truncated. Your deprecation test playbook produces 51 events per job — only 25 would be counted. This affects `totalWarnings`, per-type counts, severity classification, and trend calculations.

A fixed `page_size=200` would help but still silently cap at 200. The proper fix is to paginate through all results:

```typescript
async function fetchAllJobEvents(jobId: number): Promise<JobEvent[]> {
  const allResults: JobEvent[] = [];
  let url: string | null = awxAPI`/jobs/${jobId.toString()}/job_events/?event=deprecated&page_size=200`;

  while (url) {
    const response = await requestGet<{ count: number; results: JobEvent[]; next: string | null }>(url);
    allResults.push(...response.results);
    url = response.next;
  }

  return allResults;
}
```

Then replace the single fetch in `fetchDeprecationStats`:

```typescript
const failureCount = await runInBatches(jobs, async (job) => {
  const events = await fetchAllJobEvents(job.id);
  if (events.length > 0) {
    affectedJobsSet.add(job.id);
    totalWarnings += events.length;
    events.forEach((event) => {
      // ... existing aggregation logic
    });
  }
});
```

The existing `runInBatches` concurrency limiter (CONCURRENCY_LIMIT=10) already prevents flooding the API — pagination just adds sequential requests per job within each batch slot.

frontend/awx/administration/deprecations/DeprecationAffectedJobs.tsx
```diff
@@ -0,0 +1,100 @@
...
+    job_template?: { name: string };
+  };
+}
+
+export function DeprecationAffectedJobs() {
+  const { t } = useTranslation();
+  const { deprecationType } = useParams<{ deprecationType: string }>();
+  const decodedType = decodeURIComponent(deprecationType ?? '');
+
+  const { data: deprecationData } = useDeprecationData();
```
prat98: **Major: Time range mismatch — detail page always uses '7d' regardless of dashboard selection.**

`useDeprecationData()` defaults to `'7d'`. If the user selects '30d' on the dashboard and then clicks through to a deprecation's Affected Jobs tab, they see data from only the last 7 days — a confusing mismatch where jobs they saw on the dashboard may not appear here.

**Suggested fix:** Pass the selected time range via URL search params when navigating to the detail page:

```ts
// In useDeprecationDashboardColumns.tsx — add timeRange to the detail URL:
to={getPageUrl(AwxRoute.DeprecationDetails, {
  params: { deprecationType: dep.type },
}) + `?timeRange=${timeRange}`}
```

Then read it here:
```ts
const [searchParams] = useSearchParams();
const timeRange = (searchParams.get('timeRange') as TimeRange) || '7d';
const { data: deprecationData } = useDeprecationData(timeRange);
```

prat98 (commented):
**Code Review Run 2: 8.0/10** (up from 4.8/10)

All prior Critical and Major findings have been addressed. Three new Major findings posted as inline comments:

1. **Trend period overlap** — previous-period query overlaps current period, dampening all trends toward zero
2. **Missing event pagination** — per-job event queries lack `page_size`, silently capping at 25 events per job
3. **Time range mismatch** — Affected Jobs tab always uses 7d regardless of dashboard selection

See inline comments for suggested fixes.

frontend/awx/main/useAwxNavigation.tsx
prat98: it is intentional

tiyiprh:
## Latest changes (AI-assisted)

Addressed the three review comments from @prat98 and added a time range context indicator on the Affected Jobs tab. Note: this work was reviewed and implemented with AI assistance.

### Fix 1 — Trend period overlap
`fetchDeprecationStats` now accepts an optional `upperBound` param that appends `created__lte` to the jobs URL. The previous-period call passes the current period's start date as the upper bound, so trend comparison covers `[prevStart, currentStart]` and no longer overlaps with the current window.

### Fix 2 — Job event pagination
All `/job_events/` requests now include `&page_size=200`. AWX defaults to 25 results per page, which was silently undercounting jobs with many deprecation events. 200 covers the realistic max for a single playbook run (the v1 data completeness tradeoff PM already accepted).

### Fix 3 — Time range mismatch on detail page
Three-part fix:
- Dashboard pattern links now append `?timeRange=${timeRange}` to the detail page URL
- `DeprecationAffectedJobs` reads `timeRange` from URL params and passes it to `useDeprecationData`, so the Affected Jobs tab scans the same window the user was viewing on the dashboard
- A **"Showing jobs from last N days"** label is shown in the toolbar immediately to the left of the view switcher (consistent with the dashboard's "Showing deprecations from last 50 jobs" pattern)

The `timeRange` param is passed through `useDeprecationDashboardColumns(timeRange)` so the hook keeps the column definition clean, and a new `toolbarRightContent` prop was added to `PageToolbar`/`PageTable` to inject content into the right-aligned toolbar group without the double `margin-inline-start: auto` issue.

### Tests
- `useDeprecationData.test.tsx`: 2 new tests — `page_size=200` on events requests, and previous-period upper bound (`created__lte`)
- `DeprecationsDashboard.test.tsx`: new test verifying `?timeRange=` in pattern link `href`
- `DeprecationAffectedJobs.test.tsx`: 2 new tests verifying the time range label renders correctly from URL param (default and custom)

All 84 tests pass across 9 test files.

tiyiprh:
**CI fix (AI-assisted):** Pushed a follow-up commit to resolve the failing `eslint` and `prettier` checks.

Both failures traced back to the same 2 files with prettier formatting issues introduced in the previous commit:

- `administration/deprecations/DeprecationAffectedJobs.test.tsx` — `<Route>` element was split across 4 lines; prettier wants it on 1 line
- `administration/deprecations/hooks/useDeprecationData.tsx` — ternary expression needed a different line-break position

The local working tree already had the correct formatting (prettier was passing locally) but the fixes weren't included in the last commit. This commit adds them — no logic changes, formatting only.

prat98:
Changes looking good pending e2e test results. Pipeline is currently broken but will be fixed and unblocked soon after this PR merges https://github.com/ansible/ansible-ui/pull/3365. Once the pipeline fix is merged, we will rebase and run E2E tests against the branch

prat98:
## E2E Test Failures — Unrelated to This PR

The 2 failing tests (`constructed-inventory-crud.spec.ts`) are a **pre-existing infrastructure issue** that has been failing consistently since July 16 across all branches.

### Why these failures are not caused by this PR

1. **No inventory code was touched.** This PR adds a new Deprecations dashboard under `frontend/awx/administration/deprecations/` — entirely new files with no modifications to existing inventory, job status, or WebSocket components.

2. **The only existing file edits are additive and backward-compatible:**
   - `AwxRoutes.tsx` — adds 4 new route enums (no removals or renames)
   - `useAwxNavigation.tsx` — adds a deprecations nav item under Administration
   - `framework/useInMemoryView.tsx` — adds optional `defaultSort`/`defaultSortDirection` params with fallback to prior behavior

3. **The failures predate this PR.** Currents history shows this test has been failing on every CI run since July 16 (introduced by #3365, a pipeline bundle bump), while this PR's changes were not yet in the codebase at that time. The last green run was July 14.

4. **The error confirms no overlap.** The tests fail because `getByTestId('last-job-status')` (rendered by `InventorySourceDetails.tsx`) is not found — a component this PR does not import, modify, or interact with.

prat98 (dismissed):
Tested changes in this PR manually. Was able to confirm the base functionality for v1 is working as expected. In addition the PR has extensive test coverage and the Playwright E2E test results are looking good! 🎉

prat98:
**Review iteration 3 — no new issues from latest commit**

The last commit (`28f918db` — "remove deprecations time filter") cleanly removes the time range selector, trend indicators, and associated logic. This resolves the three Major findings from review iteration 2 (trend period overlap, missing event pagination interaction, time range mismatch between dashboard and detail page).

No new issues introduced by this commit.

prat98 (approved):

PR #3200, opened 2026-05-19
fix(ci): support SonarCloud analysis for fork PRs

vidyanambiar:
## Summary

SonarCloud checks fail on PRs from forks because GitHub Actions
does not expose secrets (`SONAR_TOKEN`) to `pull_request` events
originated from forked repositories.

**Fix:** Add a new `workflow_run`-triggered workflow (`sonar_pr.yml`)
that runs SonarCloud analysis after the Platform PR workflow completes.
`workflow_run` always executes in the base repository's context, so it
has access to secrets even when the triggering PR came from a fork.
The push-to-devel scan stays inline in `platform-pull-request.yml`
since push events always have secret access.

## Type of Change

- [x] Enhancement

## Risk Analysis - REQUIRED

- [x] **Medium** — Changes CI workflow structure for SonarCloud
  scanning. Other CI jobs (tests, linting, success gate) are unchanged.

## Testing

This can only be fully tested after merge, since `workflow_run`
workflows must exist on the default branch to be triggered.

Post-merge verification:
1. Open a PR from a fork → verify `SonarQube PR Analysis` fires
   and SonarCloud posts results
2. Open a PR from an upstream branch → verify no regression
3. Push to main/devel → verify inline branch scan still works

Assisted-by: Claude Code / Opus 4.6 (Anthropic)

### DIFF
--- a/.github/workflows/sonar_pr.yml
+++ b/.github/workflows/sonar_pr.yml
@@ -53,8 +53,9 @@ jobs:
 
       - name: Checkout PR Code
         run: |
-          gh pr checkout ${{ steps.pr.outputs.number }}
+          gh pr checkout "$PR_NUMBER"
         env:
+          PR_NUMBER: ${{ steps.pr.outputs.number }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Download Vitest Coverage
--- a/.github/workflows/sonar_pr.yml
+++ b/.github/workflows/sonar_pr.yml
@@ -1,0 +1,77 @@
+# Generated with AI assistance: Claude Code (Anthropic)
+name: SonarQube PR Analysis
+
+on:
+  workflow_run:
+    workflows:
+      - Platform PR
+    types:
+      - completed
+
+permissions: read-all
+jobs:
+  sonarcloud:
+    name: SonarQube PR Analysis
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    if: >-
+      github.event.workflow_run.event == 'pull_request'
+      && (github.event.workflow_run.conclusion == 'success'
+          || github.event.workflow_run.conclusion == 'failure')
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Fetch PR Number
+        uses: actions/download-artifact@v4
+        with:
+          name: pr_number
+          path: .
+          github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/sonar_pr.yml
```diff
@@ -0,0 +1,77 @@
...
+name: SonarQube PR Analysis
+
+on:
+  workflow_run:
+    workflows:
+      - Platform PR
+    types:
+      - completed
+
+permissions: read-all
```
BrennanPaciorek: Is `read-all` necessary here?

DAB has SonarCloud working with only the following permissions ([link](https://github.com/ansible/django-ansible-base/blob/devel/.github/workflows/sonar-pr.yml#L33-L36)):
- pull-request:read
- content:read
- actions:read

.github/workflows/sonar_pr.yml
```diff
@@ -0,0 +1,77 @@
...
+          repo: ${{ github.event.repository.full_name }}
+          number: ${{ steps.pr.outputs.number }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Checkout PR Code
+        run: |
+          gh pr checkout ${{ steps.pr.outputs.number }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
BrennanPaciorek, line 59: I am unsure if we can be certain that `steps.pr.outputs.number` will always be a number, so to be safe, we may want to put it in an `env` block to ensure proper sanitation and prevent bash injection.

```suggestion
      - name: Checkout PR Code
        run: |
          gh pr checkout ${PR_NUMBER}
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          PR_NUMBER: ${{ steps.pr.outputs.number }}
```

BrennanPaciorek (dismissed):
This looks correct, though I cannot be certain until we see some testing.

Left some comments about permissions and potential injection mitigations.

.github/workflows/sonar_pr.yml
vidyanambiar: Thank you, yes `read-all` was broader than what was required. Updated in latest commit 👍

.github/workflows/sonar_pr.yml
vidyanambiar: @BrennanPaciorek Good catch! Updated in latest commit.

BrennanPaciorek (approved):

h-kataria:
LGTM

h-kataria (approved):

PR #3222, opened 2026-05-27
fix: replace direct i18next t import with useTranslation hook

nixocio:
## Summary
- Replaces all `import { t } from 'i18next'` with the `useTranslation()` hook from `react-i18next` across 28 files in framework, AWX, EDA, and platform workspaces
- Ensures consistent translation context throughout the app by using the React hook instead of direct module imports
- Passes `t` as a parameter to non-React helper functions that cannot use hooks directly
- Leaves `i18next.t.bind(i18next)` in Zustand stores (`useDomains.tsx`, `useUIFlags.tsx`) as legitimate non-React context usage

## Test plan
- [x] `npm run tsc` passes
- [x] `npm run eslint` passes
- [x] No remaining `import { t } from 'i18next'` in source files
- [ ] CI pipeline passes
- [ ] Verify translations still render correctly in the UI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/framework/PageDashboard/PageDashboardCountBar.tsx
+++ b/framework/PageDashboard/PageDashboardCountBar.tsx
@@ -1,6 +1,6 @@
 import { ChartPie } from '@patternfly/react-charts/victory';
 import { Title } from '@patternfly/react-core';
-import { t } from 'i18next';
+import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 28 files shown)


nixocio:
Recreating with a cleaner branch name — see replacement PR.

PR #3281, opened 2026-06-15
perf(awx): reduce unnecessary job list refetches on WS status changes

vidyanambiar:
## Summary

Add status-aware routing to the WebSocket handler in `JobsList.tsx` so that only structurally significant status changes trigger a full list refresh. Intermediate status updates fetch individual jobs or are skipped entirely for off-page jobs.

Builds on top of [PR #2517](https://github.com/ansible-automation-platform/aap-ui/pull/2517), which added a 5-second leading+trailing throttle. The throttle remains — this change adds smarter routing inside  the handler.

  ## Problem

The WebSocket handler in `JobsList.tsx` calls `throttledRefresh()` on every `job`, `workflow_job`, or `project_update` message regardless of status. Even with the 5-second throttle from PR #2517, every status transition still triggers a full list refetch with all active filters against the DB. On a busy system with many concurrent jobs, this produces unnecessary expensive API calls — especially for jobs not even visible on the current page.

Additionally:
  `JobsList.tsx:102` constructs the detail URL as:
`  requestGet<UnifiedJob>(awxAPI`/unified_jobs/${action.jobId.toString()}/`)`

  This endpoint does not exist in AWX's API. The fix is to use the page item's url field, which
  already contains the correct routed path (e.g., `/api/controller/v2/jobs/{id}/`).

  Recommended Fix

```
  case 'fetch': {
    const item = pageItemsRef.current?.find((i) => i.id === action.jobId);
    const url = item?.url ?? awxAPI`/unified_jobs/${action.jobId.toString()}/`;
    requestGet<UnifiedJob>(url).then(
      (updatedJob) => updateItemRef.current(updatedJob),
      () => throttledRefresh()
    );
    break;
  }

```
  Once the detail fetch returns 200 instead of 404, on-page intermediate events will update
  in-place without a full list refresh — which is where the larger performance win should come
  from. Re-capture after fixing to validate.

  ## Solution

  Categorize WebSocket `status_changed` events into three paths:

  | Status | Action | Rationale |
  |---|---|---|
  | `new`, `pending` | Throttled full list refresh | New job may appear in list |
  | `successful`, `failed`, `error`, `canceled` | Throttled full list refresh | Sort
  order, filters, finished/elapsed fields change |
  | `waiting`, `running` (intermediate) | Fetch single job via `/unified_jobs/{id}/` →
  `updateItem` if on page; skip if not | Progress update only, no structural list
  change |
  | Missing status/id | Throttled full list refresh | Defensive fallback |

  Uses `useRef` for `pageItems` and `updateItem` to keep the callback identity stable
  and avoid a re-render loop through the `useAwxWebSocketSubscription` hook.

  ## Type of Change

  - [ ] Bug fix
  - [ ] Enhancement
  - [ ] Tests
  - [ ] Documentation
  - [x] Performance improvement

## Risk Analysis - REQUIRED

- [ ] **High**
- [ ] **Medium**
- [x] **Low** — The change is narrowly scoped to a single file (JobsList.tsx), only
  modifies the WebSocket message handler logic within that one component, and doesn't
  touch shared framework components, API wrappers, routing, or anything
  cross-workspace.

  ## Testing

  Tested with 50 concurrent jobs (page_size=10), using the same stress test
  methodology from PR #2517:

  1. Created a project with [websocket_stress_test.yml](https://github.com/vidyanambiar/test-playbooks/blob/test-jobs-perf/websocket_stress_test.yml)
  2. Created a job template, set `localStorage.setItem('perPage', '10')` to ensure
  pagination
  3. Opened Jobs page in Chrome, opened Network tab, cleared it
  4. Launched 50 concurrent jobs via stress test script
  5. Waited 5 minutes, exported HAR
  6. Compared before/after HAR files

  ### Test Results

  Tested with 50 concurrent jobs, page_size=10, ~5 minute capture window:

  | Metric | Before | After | Improvement |
  |---|---|---|---|
  | List endpoint requests | 65 | 55 | **15% fewer** |
  | Detail endpoint requests | 0 | 1 | — |
  | Total unified_jobs requests | 65 | 56 | **14% fewer** |
  | List avg response time | 2,373 ms | 2,145 ms | 10% faster |
  | List max response time | 7,670 ms | 5,476 ms | **29% faster** |
  | Avg blocked time | 315 ms | 91 ms | **71% less** |
  | Max blocked time | 5,626 ms | 3,416 ms | **39% less** |
  | Total time in list requests | 154.2s | 118.0s | **23% less** |
  | Sub-throttle gaps (<5s) | 40 | 29 | **28% fewer** |
  | Data transferred | 118.1 KB | 99.9 KB | 15% less |
  | List requests per minute | 12.8 | 10.9 | **15% lower** |

  ### Key takeaways

  - **Off-page jobs now cost nothing** — with page_size=10 and 50 jobs, 40 jobs on
  other pages no longer trigger any API calls for intermediate status updates
  - **Blocked time dropped 71%** — less connection slot contention from fewer list
  requests
  - **Max response time dropped 29%** — peak latency improved from reduced request
  volume
  - Existing throttle from PR #2517 continues to coalesce full-refresh events on the
  pending/final status path
  - Bulk cancel, bulk delete, and direct user actions are unaffected — they use
  `unselectItemsAndRefresh` which bypasses the WebSocket handler

### DIFF
--- a/frontend/awx/views/jobs/JobsList.tsx
+++ b/frontend/awx/views/jobs/JobsList.tsx
@@ -16,7 +16,7 @@ import { useJobsFilters } from '../../views/jobs/hooks/useJobsFilters';
 
 type QueryParams = { [key: string]: string };
 
-type WebSocketMessage = {
+export type WebSocketMessage = {
   group_name?: string;
   type?: string;
   status?: string;
@@ -25,6 +25,33 @@ type WebSocketMessage = {
 
 const FINAL_STATUSES = new Set(['successful', 'failed', 'error', 'canceled']);
 const NEW_JOB_STATUSES = new Set(['new', 'pending']);
+
+export type WsAction = { type: 'refresh' } | { type: 'fetch'; jobId: number } | { type: 'skip' };
+
+export function getWsAction(
+  message: WebSocketMessage | undefined,
+  pageItemIds: number[]
+): WsAction {
+  if (message?.group_name !== 'jobs') return { type: 'skip' };
+
+  switch (message?.type) {
+    case 'job':
+    case 'workflow_job':
+    case 'project_update':
+      break;
+    default:
+      return { type: 'skip' };
+  }
+
+  const status = message?.status;
+  const jobId = message?.unified_job_id;
+
+  if (!status || !jobId) return { type: 'refresh' };
+  if (NEW_JOB_STATUSES.has(status) || FINAL_STATUSES.has(status)) return { type: 'refresh' };
+
+  if (pageItemIds.includes(jobId)) return { type: 'fetch', jobId };
+  return { type: 'skip' };
+}
 
 export function JobsList(props: {
   queryParams?: QueryParams;
@@ -63,37 +90,22 @@ export function JobsList(props: {
 
   const handleWebSocketMessage = useCallback(
     (message?: WebSocketMessage) => {
-      if (message?.group_name !== 'jobs') return;
+      const pageItemIds = (pageItemsRef.current ?? []).map((item) => item.id);
+      const action = getWsAction(message, pageItemIds);
 
-      switch (message?.type) {
-        case 'job':
-        case 'workflow_job':
-        case 'project_update':
+      switch (action.type) {
+        case 'refresh':
+          throttledRefresh();
           break;
-        default:
-          return;
+        case 'fetch':
+          void requestGet<UnifiedJob>(awxAPI`/unified_jobs/${action.jobId.toString()}/`).then(
+            (updatedJob) => updateItemRef.current(updatedJob),
+            () => throttledRefresh()
+          );
+          break;
+        case 'skip':
+          break;
       }
-
-      const status = message?.status;
-      const jobId = message?.unified_job_id;
-
-      if (!status || !jobId) {
-        throttledRefresh();
-        return;
-      }
-
-      if (NEW_JOB_STATUSES.has(status) || FINAL_STATUSES.has(status)) {
-        throttledRefresh();
-        return;
-      }
-
-      const existingJob = (pageItemsRef.current ?? []).find((item) => item.id === jobId);
-      if (!existingJob) return;
-
-      void requestGet<UnifiedJob>(awxAPI`/unified_jobs/${jobId.toString()}/`).then(
-        (updatedJob) => updateItemRef.current(updatedJob),
-        () => throttledRefresh()
-      );
     },
     [throttledRefresh]
   );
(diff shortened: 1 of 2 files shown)


vidyanambiar:
## Code Review: perf(awx): reduce unnecessary job list refetches on WS status changes

**Score**: 9.5/10 (Functionality 10, Security 10, Quality 8.5)

No critical or major findings. Three minor test improvements suggested:

1. **[JobsList.test.tsx]** Add an explicit test for `system_job` type → verifies the `skip` behavior is intentional (matches pre-existing behavior)
2. **[JobsList.test.tsx]** Replace `setTimeout(500)` in the "skip off-page job" test with a synchronous assertion or fake timers — the skip path triggers no async work
3. **[JobsList.test.tsx]** Simplify the `vi.mock` return for `useAwxWebSocketSubscription` — the component doesn't use the return value

**Observations**: Well-structured extraction of `getWsAction` as a pure function for testability. Defensive defaults (missing fields → refresh, failed detail fetch → refresh) prevent silent data staleness. Ref pattern correctly stabilizes callback identity.

**Needs Human Judgment**: Should `system_job` WS events trigger list updates? Currently skipped (unchanged from before this PR).

*Review performed in API-only mode — codebase-wide verification commands were skipped.*

---
*Generated with AI assistance: Claude Code / Opus 4.6 (Anthropic)*

frontend/awx/views/jobs/JobsList.tsx
```diff
@@ -45,17 +83,27 @@ export function JobsList(props: {
   );
   useEffect(() => () => throttledRefresh.cancel(), [throttledRefresh]);
 
+  const pageItemsRef = useRef(pageItems);
```
keithjgrant, line 87: why are these stored to refs?

vidyanambiar: The refs break a dependency chain that would otherwise cause a re-render loop. `useAwxWebSocketSubscription` includes the `onMessage` callback in a `useEffect` dependency array, so if `pageItems`/`updateItem` were direct dependencies of `handleWebSocketMessage`, every data update would create a new callback, re-trigger the WS effect, call the handler with the stale last message, trigger another data update, and loop.

The refs let the callback always read fresh values via `.current` while keeping its own identity stable (only `throttledRefresh` in the dep array).

Added a one-line comment above the refs to make this clearer for future readers.

keithjgrant (dismissed):

keithjgrant (approved):

PR #3305, opened 2026-06-19
Prevent crash launching WJT with no org

demonpig:
## Summary

Fixes a crash that occurred when a non-superuser (team-level execute access only) launched a Workflow Job Template with no organization set, where a survey or other prompt forced the launch wizard open.

The UI threw `TypeError: can't access property "id", f[0] is undefined` instead of creating the workflow job. The backend launch API itself succeeds — this was a UI-only bug.

**Root cause:** `useLabelPayload` unconditionally called `GET /organizations/` before processing any labels and dereferenced `results[0].id` without a length guard. For users with no org membership the API returns `{ results: [] }`, causing the crash. Additionally, `TemplateLaunchWizard` was calling `createLabelPayload` unconditionally even when `ask_labels_on_launch` is false, so the org-fetch path was reached on every wizard submit regardless of whether labels were prompted.

## Type of Change

- [x] Bug fix
- [ ] Enhancement
- [x] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

> Changes are isolated to the WJT/JT launch wizard and template form submit handlers. The only behavioral change for users unaffected by the bug is that `GET /organizations/` is no longer called on every wizard submit when `ask_labels_on_launch` is false.

### Manual Testing

**Repro scenario (requires a non-superuser with team-level execute access):**

1. Log in to the AAP web UI as **admin** (`admin` / `admin`).
2. Create a **Workflow Job Template** named `NoOrgTestWJT` with **Organization** set to none (leave blank / null).
3. Open the **Workflow Visualizer**, add a single node that launches **Demo Job Template**, and save.
4. On the **Survey** tab, add one survey question (e.g. text, variable `test_var`, required) and enable the survey.
5. Create a **User** named `NoOrgUser` (User Type: normal; do not assign to any organization).
6. Create a **Team** named `DefaultTeam` in the **Default** organization.
7. Add `NoOrgUser` to team `DefaultTeam`.
8. On `NoOrgTestWJT`, open **Team Access**, add role **Workflow Job Template Execute** (or equivalent execute role) for `DefaultTeam`.
9. Log out and log in as **NoOrgUser** (`NoOrgUser123!` on the lab; Gateway login at `/api/gateway/v1/login/`).
10. Navigate to **Automation Execution → Templates**, select **NoOrgTestWJT**, and click **Launch**.
11. Complete the **Survey** step, proceed through **Review**, and submit the launch.

**Before:** UI crashes with `Failure to launch — TypeError: can't access property "id", f[0] is undefined`.

**After:** The workflow job is created successfully and the user is navigated to the job output page.

**Workaround that existed before this fix:** Assign an organization to the WJT.

### Unit Tests Added

New Vitest coverage added across four files:

- **`useLabelPayload.test.tsx`** — org fetch not called when no labels need creation; descriptive error thrown when org results are empty and a new label must be created; label created successfully when org is available.
- **`TemplateLaunchWizard.test.tsx`** — `createLabelPayload` skipped when `ask_labels_on_launch` is false; called when true.
- **`WorkflowJobTemplateForm.test.tsx`** — org fetch skipped when no labels are added; `results[0]` access guarded.
- **`TemplateForm.test.tsx`** — same as above.

### DIFF
--- a/frontend/awx/resources/inventories/inventorySources/InventorySourceDetails.test.tsx
+++ b/frontend/awx/resources/inventories/inventorySources/InventorySourceDetails.test.tsx
@@ -2,6 +2,7 @@ import { render, screen, waitFor } from '@testing-library/react';
 import { http, HttpResponse } from 'msw';
 import { setupServer } from 'msw/node';
 import { MemoryRouter, Route, Routes } from 'react-router-dom';
+import { SWRConfig } from 'swr';
 import { afterAll, afterEach, beforeAll, describe, expect, it, vi } from 'vitest';
 import { InventorySourceDetails, LastJobTooltip } from './InventorySourceDetails';
 
@@ -89,14 +90,16 @@ afterAll(() => server.close());
 
 function renderInventorySourceDetails(inventorySourceId?: string) {
   return render(
-    <MemoryRouter initialEntries={['/inventories/inventory/1/sources/1/details']}>
-      <Routes>
-        <Route
-          path="/inventories/:inventory_type/:id/sources/:source_id/details"
-          element={<InventorySourceDetails inventorySourceId={inventorySourceId} />}
-        />
-      </Routes>
-    </MemoryRouter>
+    <SWRConfig value={{ provider: () => new Map() }}>
+      <MemoryRouter initialEntries={['/inventories/inventory/1/sources/1/details']}>
+        <Routes>
+          <Route
+            path="/inventories/:inventory_type/:id/sources/:source_id/details"
+            element={<InventorySourceDetails inventorySourceId={inventorySourceId} />}
+          />
+        </Routes>
+      </MemoryRouter>
+    </SWRConfig>
   );
 }
 
--- a/frontend/awx/resources/templates/TemplateForm.test.tsx
+++ b/frontend/awx/resources/templates/TemplateForm.test.tsx
@@ -110,6 +110,10 @@ const server = setupServer(
   http.get(awxAPI`/job_templates/42/`, () => HttpResponse.json(mockJobTemplate)),
   http.get(awxAPI`/job_templates/42/instance_groups/`, () =>
     HttpResponse.json({ count: 0, results: [] })
+  ),
+  http.get(
+    ({ request }) => request.url.includes('/webhook_key/'),
+    () => HttpResponse.json({ webhook_key: 'test-key' })
   )
 );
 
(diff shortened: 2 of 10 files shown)


demonpig:
I have to rework this PR and will get back to it later.

demonpig:
This one is going to need more work.

acruzgon:
@demonpig — We rebased this branch on latest `devel`, fixed the CI failures (TypeScript narrowing + an unrelated SWR cache issue in `InventorySourceDetails.test.tsx`), and all checks are now green.

We tried to reproduce the bug manually on a **2.6-stable-RPM-A** build (Controller 4.7.14) but the error didn't surface there — the launch succeeded without crashing. Do you have an instance where you can still reproduce the original `"Cannot read properties of undefined (reading 'id')"` error? We'd like to do a live before/after confirmation against that environment.

We were able to confirm the bug by running the dev server locally from `devel` (without your fix) — the crash triggers as expected for a no-org user launching a survey-enabled WJT. With your fix applied, the launch succeeds cleanly.

acruzgon:
Confirmed on @demonpig's lab (AAP 2.6, https://10.0.110.156):

**Bug reproduction** ✅ — Logged in as `NoOrgUser`, launched WJT `NoOrgTestWJT` (id 349, no org, survey enabled), filled the survey, and clicked Finish:

```
❌ Failure to launch — Cannot read properties of undefined (reading 'id')
```

**Fix verification** ✅ — Running the dev server locally from this branch (with the fix), same scenario launches the workflow job successfully and navigates to the output page.

| Environment | Result |
|-------------|--------|
| Max's lab (2.6 latest, stock UI) | ❌ Crash: `Cannot read properties of undefined (reading 'id')` |
| Local dev server (`devel` branch, no fix) | ❌ Same crash |
| Local dev server (`AAP-79301` branch, with fix) | ✅ Job launched, navigated to output |

Root cause confirmed: `useLabelPayload` unconditionally fetches `/organizations/` and dereferences `results[0].id` without a length guard. For `NoOrgUser`, that API returns `{ count: 0, results: [] }`.

acruzgon (approved):
Approved. Bug confirmed on live lab (https://10.0.110.156) as NoOrgUser — crash reproduces with stock UI. Fix verified locally — launch succeeds cleanly. All CI checks green.

PR #3315, opened 2026-06-25
Add regression tests for SWR error retry configuration

marshmalien:
## Summary

  Adds tests that would have caught SWR infinite retry loop on 5xx API errors

  **Vitest (unit)** — verifies `PageSettingsProvider` wires the three SWRConfig properties that prevent retry storms:
  - `onErrorRetry` — caps retries instead of looping forever on 5xx
  - `revalidateOnFocus` — prevents refetch bursts when switching tabs
  - `dedupingInterval` — prevents duplicate concurrent requests

  **Playwright (integration)** — proves the retry cap works in a real browser:
  - Intercepts the `unified_jobs` API to return 503
  - Waits for retries to settle, then confirms no more requests arrive
  - Passes on `devel` (retries stop), fails on `stable-2.7` (retries continue)

## Type of Change

- [ ] Bug fix
- [ ] Enhancement
- [x] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

### Ephemeral E2E Tests


vidyanambiar:
---
  Code Review: Add regression tests for SWR error retry configuration

  Verdict: READY_FOR_HUMAN_REVIEW

  Solid, low-risk test-only PR. The mock-capture pattern is well-implemented — testing what PageSettingsProvider passes to
  SWRConfig is better than the old useSWRConfig() approach that tested through SWR's internal config merging.

  What works well

  - The spy wrapper captures props while still rendering the real SWRConfig, so other tests remain unaffected
  - Correct Vitest patterns: createElement in the hoisted vi.mock factory, Readonly props, proper cleanup
  - Clean consolidation of the duplicate revalidateOnFocus test

  Suggestions (non-blocking)

  1. onErrorRetry assertion depth — The test asserts typeof config.onErrorRetry === 'function', which would catch someone
  removing the prop but not someone accidentally passing a no-op () => {} (which would silently fall back to SWR's default
  infinite retry). A stronger assertion would verify identity against the exported handler or test a behavioral property (e.g.,
  call it with retryCount: 3 and verify no revalidation). The existing SWR Error Retry Logic tests cover the handler in
  isolation, so the gap is narrow.
  2. dedupingInterval value — > 0 would pass for any positive number. Asserting against SWR_DEDUPING_INTERVAL_MS directly would
  be tighter.

  ---

h-kataria (approved):
LGTM

PR #3319, opened 2026-06-26
[AAP-80791][AAP-80796] - Correct cursor pagination and value type coercion in useQueryPlatformOptions

demonpig:
## Summary

Fixes two independent bugs in `useQueryPlatformOptions` that made the **User** and **Application** toolbar filter dropdowns on the API Tokens page (and any Platform page backed by this hook) unusable.

**Bug 1 — Page crash on selection:** Selecting any user from the filter dropdown crashed the entire page with `TypeError: U.includes is not a function`. The hook returned numeric IDs with a TypeScript-only `as string` cast that has no runtime effect, so the number flowed into `usePlatformView`'s `.includes(',')` call and threw. Fixed by using `String()` for a real runtime conversion.

**Bug 2 — HTTP 400 on "Load more":** Paginating the dropdown sent a malformed URL (`?order_by=username&order_by=testuser09`) that the Gateway's `OrderByBackend` rejected because duplicate `order_by` keys collapse to the last value, which is a cursor string rather than a valid field name. Fixed by using the correct Django cursor filter (`username__gt=testuser09`), matching the existing AWX pattern in `PageFormSingleSelectAwxResource`.

Both fixes are in the shared hook; all 8 call sites across 7 files are healed automatically. Two new tests cover the previously untested cursor pagination branch and the value type coercion.

## Type of Change

- [x] Bug fix
- [ ] Enhancement
- [x] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

### Ephemeral E2E Tests

### Manual Testing

1. Seed enough users to exceed the Gateway's default page size (or use `page_size=10`).
2. Navigate to **Access → API Tokens**.
3. Open the **User** filter dropdown — first page loads without error.
4. Select any user — page should filter without crashing (fixes Bug 1).
5. Click **Load more** — next page of users loads without an "Error loading options" message (fixes Bug 2).
6. Repeat steps 3–5 for the **Application** filter dropdown.

### DIFF
--- a/platform/common/useQueryPlatformOptions.test.tsx
+++ b/platform/common/useQueryPlatformOptions.test.tsx
@@ -5,6 +5,14 @@ import { setupServer } from 'msw/node';
 import { gatewayAPI } from '../utils/gateway-api-utils';
 import { useQueryPlatformOptions } from './useQueryPlatformOptions';
 
+const multiItemResponse = {
+  count: 20,
+  results: [
+    { id: 10, username: 'testuser09' },
+    { id: 11, username: 'testuser10' },
+  ],
+};
+
 const mockResponse = {
   count: 1,
   next: null,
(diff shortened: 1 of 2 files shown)


demonpig:
@keithjgrant im not able to see the konflux issue being reported. Is there anything specific that needs to be changed?

daphnemaeve:
Tested locally, lgtm. Running playwright to verify

vidyanambiar (approved):

PR #3321, opened 2026-06-26
JobList - reduced network traffic

keithjgrant:
## Summary

Significantly reduces network traffic on the Jobs list, mostly to the unified_jobs endpoint. Reworks how websocket messages are handled:

- When a job starts, only that job is fetched from unified_jobs. Any list filters are included so filtered jobs don't appear on list.
- When job status changes, status is updated in memory without fetching from API
- When job enters `running` status, the Started time is approximated at the time the message is received
- When a job finishes, and the duration is calculated based on started/finished times.
- SWR refresh interval (default 60s) will periodically refresh the entire page so list data should not become too stale or out of sync

NOTE: when a new job is added to the page, this attempts to maintain sorting on the page to match the API sorting, but there is potential for the order of jobs listed to change. In my testing, this is pretty stable, but if any shifting order is jarring, we may need to revert to a full load of the page when a new job comes in.

## Type of Change

- [x] Bug fix
- [ ] Enhancement
- [ ] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

### DIFF
--- a/frontend/awx/access/credentials/hooks/useCredentialToolbarActions.test.tsx
+++ b/frontend/awx/access/credentials/hooks/useCredentialToolbarActions.test.tsx
@@ -41,6 +41,8 @@ describe('useCredentialToolbarActions', () => {
     keyFn: (item: { id: number }) => item.id,
     limitFiltersToOneOrOperation: true as const,
     updateItem: vi.fn(),
+    upsertItem: vi.fn(),
+    listUrl: '',
   };
 
   const mockPageNavigate = vi.fn();
--- a/frontend/awx/analytics/automation-dashboard/AutomationDashboard.test.tsx
+++ b/frontend/awx/analytics/automation-dashboard/AutomationDashboard.test.tsx
@@ -135,6 +135,8 @@ const mockMainTableView: IAutomationDashboardView['mainTableView'] = {
   unselectItemsAndRefresh: vi.fn(),
   limitFiltersToOneOrOperation: true,
   updateItem: vi.fn(),
+  upsertItem: vi.fn(),
+  listUrl: '',
 } as unknown as IAutomationDashboardView['mainTableView'];
 
 const mockView: IAutomationDashboardView = {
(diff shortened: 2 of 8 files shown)


vidyanambiar:
## Code Review: JobList - reduced network traffic

**Score**: 10/10 (Functionality 10, Security 10, Quality 10)

No critical or major findings. Well-designed optimization with clean architecture.

### Summary

The three-tier WebSocket action model (patch in-memory → fetch single → skip) is a clean pattern that significantly reduces network traffic. All code paths were traced and verified correct.

### Verified

- **URL construction**: `new URL(listUrl, 'http://localhost')` safely parses relative URLs; `searchParams.set('id', ...)` preserves active filters for the single-item fetch
- **Patch scope**: Only `status`, `started`, `finished` fields can be patched — no authorization-relevant fields
- **`upsertItem`**: Correctly prevents duplicates via `findIndex`, inserts in sort order when sort is active
- **`compareByField`**: Null/undefined returns 0 (neutral sort position), read-only access (no prototype pollution)
- **`count_disabled=1`**: Good optimization avoiding expensive `COUNT(*)` on single-item fetch
- **Backward compatible**: `updateItem` correctly retained for other consumers (Hosts, Schedules, Dashboard)
- **Self-correcting**: 60s SWR refresh bounds all client-side approximations (timestamps, page size, sort order)
- **WS trust boundary**: Same-origin authenticated channel, appropriate for display-state updates

### Nit-Level Suggestions (optional)

1. **Guard `started` overwrite**: If a duplicate `running` WS message arrives (reconnection), `patch.started = new Date().toISOString()` overwrites the server-provided timestamp. Consider: `if (data.started && existing.started) delete data.started;` in the patch handler.
2. **Test sorted insertion**: The `upsertItem` "prepend" test doesn't configure sort columns, so the `newItems.sort(...)` branch is unexercised. The `compareByField` unit tests compensate partially.
3. **`compareByField` location**: If a second consumer arises, move to `/frontend/common/`.

### Needs Human Judgment

- Sort order stability on busy clusters with many concurrent job launches
- 60s SWR refresh interval as implicit consistency bound for all approximations

---

*AI-assisted review: Claude Code / Opus 4.6 (Anthropic)*

vidyanambiar (approved):
Tested locally and it works well. We can merge it if the playwright run looks good 👍

PR #3122, opened 2026-04-20
[AAP-42092] Use case-insensitive filters for all toolbar text searches

amasolov:
## Summary

- Switch all user-typed toolbar text filter lookups from `__contains` to `__icontains` across 13 files
- Fixes case-sensitive search on the Users page (username, first name, last name, email) and all other toolbar text filters (role names, task names, authenticator mapping names)
- Restores case-insensitive search behaviour matching AAP 2.4

## Type of Change

- [x] Bug fix
- [ ] Enhancement
- [ ] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

- [ ] **High** — Broad platform impact (e.g., SWR config, shared framework components, authentication, routing, API wrappers, build/bundler config). Changes affect multiple workspaces or could cause widespread regressions.
- [x] **Medium** — Scoped but cross-cutting (e.g., shared utility functions, changes to multiple pages within one workspace, component library updates, test infrastructure). Limited blast radius but touches common code paths.
- [ ] **Low** — Narrowly scoped (e.g., single page fix, styling tweak, documentation, test-only changes, copy/string updates). Minimal risk of unintended side effects.

### Manual Testing

- Navigate to Access > Users, search for a username in different case (e.g. uppercase when stored as lowercase). Verify the user is found.
- Repeat for email, first name, and last name filters on the Users page.
- Navigate to roles lists (AWX, EDA, Hub) and search for a role name in different case. Verify the role is found.
- Navigate to Hub > Tasks and search for a task name in different case.

Closes #3121

PR #3322, opened 2026-06-28
[AAP-80860] Fix EDA role wizard resource type dropdown and search

amasolov:
## Summary

- Fix empty "Resource type" dropdown in EDA role wizard by parsing OPTIONS choices as tuples (matching AWX implementation)
- Switch EDA resource selection toolbar filter to `name__icontains` for case-insensitive search (missed in #3122)

## Type of Change

- [x] Bug fix
- [ ] Enhancement
- [ ] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

### Manual Testing

- Navigate to EDA > Users, select a user, click "Add roles"
- Verify the "Resource type" dropdown now lists EDA resource types (Credentials, Projects, Rulebook Activations, Decision Environments, etc.)
- Select a resource type and verify the resource list loads
- Search for a resource by name using different case and verify it is found

Relates to #3122

### DIFF
--- a/frontend/eda/access/common/EdaRolesWizardSteps/EdaSelectResourceTypeStep.tsx
+++ b/frontend/eda/access/common/EdaRolesWizardSteps/EdaSelectResourceTypeStep.tsx
@@ -4,10 +4,7 @@ import { useOptions } from '@ansible/common-ui/crud/useOptions';
 import { useTranslation } from 'react-i18next';
 import { edaAPI } from '../../../common/eda-utils';
 
-interface ContentTypeOption {
-  value: string;
-  display_name: string;
-}
+type ContentTypeOption = [string, string];
 
 export function EdaSelectResourceTypeStep() {
   const { t } = useTranslation();
@@ -28,13 +25,13 @@ export function EdaSelectResourceTypeStep() {
       name="resourceType"
       options={options
         .filter(
-          (option) =>
-            option.value.startsWith('eda.') &&
+          ([value, _]) =>
+            value?.startsWith('eda.') &&
             !['extravar', 'auditrule', 'rulebookprocess', 'rulebook'].some(function (v) {
-              return option.value.endsWith(v);
+              return value.endsWith(v);
             })
         )
-        .map(({ value, display_name }) => ({
+        .map(([value, display_name]) => ({
           value,
           label: display_name,
         }))}
--- a/frontend/eda/access/common/EdaRolesWizardSteps/EdaSelectResourcesStep.tsx
+++ b/frontend/eda/access/common/EdaRolesWizardSteps/EdaSelectResourcesStep.tsx
@@ -80,7 +80,7 @@ export function EdaSelectResourcesStep(props: { userOrTeamName: string }) {
         key: 'name',
         label: t('Name'),
         type: ToolbarFilterType.MultiText,
-        query: 'name',
+        query: 'name__icontains',
         comparison: 'contains',
       },
     ],


amasolov:
After further investigation using Playwright against a live AAP 2.7 instance, the "Resource type" dropdown in the unified "Assign roles" wizard (`/access/users/{id}/roles`) populates correctly with all 18 resource types from Automation Execution, Automation Decisions, Automation Content, and System.

The unified wizard uses `GET /api/gateway/v1/service-index/role-types/` (a gateway-level API) to populate the dropdown, and does not call `OPTIONS /api/eda/v1/role_definitions/` at all. The `EdaSelectResourceTypeStep.tsx` component (modified in this PR) does not appear to be rendered in the AAP 2.7 unified UI.

This suggests the original report may have been against a standalone EDA development server or older codebase where that component was still active. Leaving this PR open for team discussion, as the fix may still be valid for standalone EDA UI contexts.

jon-nfc:
@amasolov 

This PR is blocked by: https://github.com/ansible/eda-server-operator/pull/331

> This suggests the original report may have been against a standalone EDA development server or older codebase where that component was still active. Leaving this PR open for team discussion, as the fix may still be valid for standalone EDA UI contexts.

That is correct and why I included the site header in my report (https://github.com/ansible/ansible-ui/pull/3122#issuecomment-4825377777).

The version of the EDA server being used is https://github.com/ansible/eda-server/commit/b4030df6aea508dd0d1d228150f3c2879bbd1598, due to a bug somewhere after this commit because login does not work. see bug report https://github.com/ansible/eda-server/issues/1504.

**Currently no newer version of EDA Server can be used until this bug is addressed**. There is a PR open, https://github.com/ansible/eda-server-operator/pull/331. However no one appears to be interested in addressing.

As such no newer version of EDA server can be used to confirm if this PR is relevant/"still an issue" for EDA server, as the eda-server cant be logged on to.

marshmalien:
@amasolov Thanks for putting this together! 
I looked at the eda-server backend and it looks like most EDA endpoints don't actually support `name__icontains` as a filter lookup. When the UI sends `?name__icontains=foo`, the backend silently ignores it and returns everything unfiltered. 
I tested locally `https://localhost:4100/api/eda/v1/credential-types/?name__icontains=basic&order_by=name&page=1&page_size=10` and got back all 28 credential types instead of just the match. Is this query supported in the standalone eda builds?

Also the test mock in EdaSelectResourceTypeStep.test.tsx still uses the old object format ({value, display_name}) instead of tuples. Worth updating to ['eda.project', 'Project'] and adding an assertion that filtered options actually render, otherwise the test can't catch a regression.

amasolov:
Closing this PR after verifying both changes against eda-server/django-ansible-base source and a live AAP 2.7 instance (EDA 1.3.2). Both changes turn out to be incorrect, and the real bugs are server-side.

**1. The tuple change doesn't fix the empty dropdown — the server sends no choices at all.**

On live AAP 2.7, `OPTIONS /api/eda/v1/role_definitions/` returns for `content_type`:

```json
{"type": "field", "label": "Content type"}
```

No `choices` key, in either GET or POST actions. Root cause: django-ansible-base commit `894af11` (AAP-48392, July 2025) changed `content_type` on `RoleDefinitionSerializer` from `ContentTypeField(ChoiceLikeMixin)` — a `ChoiceField` subclass — to `serializers.SlugRelatedField`. DRF's `SimpleMetadata` explicitly excludes RelatedFields from choices rendering, and eda-server's `EDAMetadata` doesn't override that. So:

- **Pre-July-2025 DAB**: choices returned as `{value, display_name}` objects (standard DRF ChoiceField rendering) — the *existing* object parsing was correct, and tuple destructuring throws `object is not iterable`, which is exactly the current Vitest (eda) failure against the devel fixture.
- **Post-July-2025 DAB** (incl. AAP 2.7 and the env in the original report): no choices at all — dropdown is empty under either parsing.

The AWX component parses tuples because AWX ships its own `Metadata` class that renders choices as `(value, label)` pairs; EDA uses DRF-standard rendering. Same-looking component, different wire contract — the AWX precedent doesn't transfer.

Filed against django-ansible-base: ansible/django-ansible-base#1060 (restore choices, or clients move to `GET /role_metadata/`).

**2. `name__icontains` is a filtering regression, confirming @marshmalien's finding.**

`name__icontains` appears nowhere in the eda-server tree. Every FilterSet declares a plain `name` filter with a server-side `i`-prefixed `lookup_expr` (`istartswith` for eda-credentials/projects/credential-types/decision-environments/event-streams, `icontains` for activations/rulebooks/activation-instances; audit-rules has no filterset at all). With no django-filter STRICTNESS configured, the unknown param is silently dropped. Reproduced on AAP 2.7:

- `?name__icontains=basic` → 28 of 28 credential types (unfiltered — matches @marshmalien's repro exactly)
- `?name=basic` (lowercase) → 1 result, "Basic Event Stream" — already case-insensitive

Same in standalone builds — it's the filter definition, not routing.

There *is* a genuine UX bug underneath: the UI labels the filter "contains" but six of nine endpoints are `istartswith` server-side — `?name=auth` returns 0 results even though "OAuth2 Event Stream" exists. Filed against eda-server: ansible/eda-server#1624.

@jon-nfc this also explains your environment without needing a working newer eda-server login — the choices removal shipped in DAB July 2025, so any build after that has the empty dropdown regardless of UI parsing.

PR #3417, opened 2026-07-28
batch dependabot non-major production dependencies

keithjgrant:
## Summary

Only dev-dependencies were being batched in dependabot config. This adds another batch for production dependencies.

Major cross-cutting dependencies like React, Vite, patternfly are excluded from the batch.

## Type of Change

- [ ] Bug fix
- [ ] Enhancement
- [ ] Tests
- [ ] Documentation
- [x] Other (please specify)

## Risk Analysis - REQUIRED

- [x] **Medium** - This PR moves a lot of non-production code libraries from `dependencies` to `devDependencies`


.github/dependabot.yml
```diff
@@ -11,14 +11,27 @@ updates:
...
+        update-types:
+          - 'minor'
+          - 'patch'
+        exclude-patterns:
+          - 'react'
+          - 'react-dom'
+          - 'react-router-dom'
+          - 'typescript'
+          - 'vite'
+          - 'webpack'
```
nicklemmon: Aren't these dev dependencies? They should be if they aren't listed that way.

nicklemmon (dismissed):

nicklemmon (approved):

PR #3435, opened 2026-08-03
AAP-85062 XLAB Fix Export CSV button position

kcagran:
## Summary
[AAP-85062]  Fix Export CSV button position

This pull request refactors the layout of the `DashboardTableToolbarRow` component to improve responsiveness and consistency across different screen sizes. The changes primarily reorganize the usage of PatternFly's `Flex` and `FlexItem` components, adjusting alignment, direction, and spacing to provide a better user experience and cleaner code.

**Layout and Responsiveness Improvements:**

* Simplified the top-level `Flex` container to always use a row direction, added `flexFlow: 'row'`, and introduced consistent `rowGap` and `columnGap` for improved spacing and alignment.
* Refactored nested `Flex` structures to use responsive direction and alignment properties, ensuring that controls (such as the hourly rate input and toggle switch) are consistently aligned on larger screens (`2xl`) and stacked on smaller screens. [[1]](diffhunk://#diff-c5bf117bdc1765adbd49308947e402a2746a7c64c3da9037bcb9b4f4381e2f42L108-R124) [[2]](diffhunk://#diff-c5bf117bdc1765adbd49308947e402a2746a7c64c3da9037bcb9b4f4381e2f42L149-R158)
* Moved the export button into its own `FlexItem` with right alignment, ensuring it remains visually distinct and properly positioned regardless of screen size.
* Removed redundant or unnecessary `Flex` wrappers, reducing nesting and simplifying the component structure for easier maintenance.

These changes collectively enhance the maintainability of the component and provide a more consistent and responsive UI layout.

## Type of Change

- [x] Bug fix
- [ ] Enhancement
- [ ] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

### Ephemeral E2E Tests

### Screenshots
image

### DIFF
--- a/framework/PageDashboard/PageDashboardCard.test.tsx
+++ b/framework/PageDashboard/PageDashboardCard.test.tsx
@@ -1,0 +1,66 @@
+import { render, screen } from '@testing-library/react';
+import { userEvent } from '@testing-library/user-event';
+import { MemoryRouter } from 'react-router-dom';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { PageDashboardCard } from './PageDashboardCard';
+
+function renderCard(props: Parameters<typeof PageDashboardCard>[0]) {
+  return render(
+    <MemoryRouter>
+      <PageDashboardCard {...props} />
+    </MemoryRouter>
+  );
+}
+
+describe('PageDashboardCard', () => {
+  beforeEach(() => {
+    localStorage.clear();
+  });
+
+  it('should not render the link when linkText is not provided', () => {
+    renderCard({ title: 'Inventories' });
+
+    expect(screen.queryByTestId('card-link-text')).not.toBeInTheDocument();
+    expect(screen.queryByRole('link')).not.toBeInTheDocument();
+  });
+
+  it('should render the link when linkText and to are provided', () => {
+    renderCard({ title: 'Inventories', linkText: 'View all', to: '/inventories' });
+
+    const link = screen.getByRole('link', { name: 'View all' });
+    expect(link).toBeInTheDocument();
+    expect(link).toHaveAttribute('href', '/inventories');
+  });
+
+  it('should render the header and link when only linkText is provided without a title', () => {
+    renderCard({ linkText: 'View all', to: '/inventories' });
+
+    expect(screen.getByRole('link', { name: 'View all' })).toBeInTheDocument();
+  });
+
+  it('should not render a header when neither title nor linkText are provided', () => {
+    renderCard({ children: <div>Body</div> });
+
+    expect(screen.queryByTestId('card-title')).not.toBeInTheDocument();
+    expect(screen.queryByTestId('card-link-text')).not.toBeInTheDocument();
+  });
+
+  it('should render children in the card body', () => {
+    renderCard({ title: 'Inventories', children: <div>Card content</div> });
+
+    expect(screen.getByText('Card content')).toBeInTheDocument();
+  });
+
+  it('should call onClick when the footer action button is clicked', async () => {
+    const user = userEvent.setup();
+    const onClick = vi.fn();
+    renderCard({
+      title: 'Inventories',
+      footerActionButton: { title: 'Add inventory', onClick },
+    });
+
+    await user.click(screen.getByRole('button', { name: 'Add inventory' }));
+
+    expect(onClick).toHaveBeenCalledTimes(1);
+  });
+});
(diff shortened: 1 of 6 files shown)


tiyiprh:
_Claude assisted_

Thanks for picking this up! After reviewing the PR in the UI, I think there's a cleaner approach that would better match the expected layout from [AAP-85062](https://redhat.atlassian.net/browse/AAP-85062) — sorry for the extra direction here, but wanted to share what I found while testing. I'll update the issue accordingly as well.
cc @marshmalien let me know if you agree!

**What's working**
- Export is no longer tied to the switch flex, so it won't drop below the switch when the sidebar expands

**What I noticed while reviewing**

The current flex refactor still doesn't fully solve the layout issue. Export sits beside the input/switch group rather than being anchored top-right, and the switch looks slightly pushed down at wider breakpoints because of `alignSelfFlexEnd` at `2xl`.

**Suggested approach — two-row layout**

After comparing against the design intent, a better fix would be:

**Row 1 — Card header**
- Title: "Cost calculation" (left)
- Export as CSV button (right, inline with title)

**Row 2 — Toolbar beneath header**
- Hourly rate | Monthly AAP cost | Switch — all on one line
- Switch vertically centered with the input fields

**Suggested changes**

1. **`DashboardMainTableCard.tsx`** — add `title` and move Export into `headerControls` on `PageDashboardCard`:

```tsx
<PageDashboardCard
  title={t('Cost calculation')}
  headerControls={
    <DashboardExportButton
      exportType="csv"
      title={t('Export as CSV')}
      icon={ExportIcon}
      isDisabled={...}
      onExport={exportCsv}
    />
  }
  ...
>

**Comment by marshmalien** [comment]
@tiyiprh I agree! The two-row layout (header with title + export, toolbar with inputs + switch) is cleaner and more maintainable.

**Comment by kcagran** [comment]
@tiyiprh @marshmalien Done. If there's anything else, let me know.

**Comment by tiyiprh** [comment]
_Cursor assisted_

Thanks for the update — the two-row layout is much closer to what we had in mind! After reviewing the latest changes in the UI, a few small polish items stood out. Sorry for the extra round of feedback!

**What's working**
- Export is in the card header, separate from the switch — no more dropping below the switch when the sidebar expands
- "Cost calculation" title + toolbar row underneath matches the intended structure

**Remaining polish**

1. **Export button alignment** — The Export button sits slightly inset from the right edge of the card, not fully flush right. Looks like the empty link slot in `PageDashboardCard`'s header (the `linkText` FlexItem renders even when there's no link). Could we get Export aligned to the true top-right corner?

2. **Switch vertical alignment** — The switch looks slightly low relative to the two input fields. Could we restore `alignItemsCenter` on the toolbar `Flex` so the switch is vertically centered with the inputs?

3. **Compact mode on this card** — This card uses `isCompact` on `PageDashboardCard`, which other dashboard cards (Successful jobs, Top 5 projects, etc.) don't. Now that we've added a title header, the compact styling makes "Cost calculation" feel a bit different from the other card titles. Could we drop `isCompact` here so the header matches the rest of the dashboard?

4. **Inner KPI cards sizing** (optional / follow-up) — The four value cards inside this card (Cost of manual automation, etc.) use the same size as the top-row dashboard cards. Since they're nested inside a parent card, they might read better slightly smaller — but that feels like a separate design pass. Happy to track that in another PR if easier to keep this one focused.

Let me know if any of this needs clarification — thank you!

cc @marshmalien

**Comment by marshmalien** [comment]
@tiyiprh 
1. Export button alignment
Good catch. The linkText FlexItem always renders even when there's no linkText passed, and since PF6 Flex adds `margin-inline-end (24px)` to every child except the last, the export button ends up inset because the empty FlexItem sits right after it. Fix:
```
// PageDashboardCard

{props.linkText && (
  <FlexItem>
    <Content data-cy="card-link-text" data-testid="card-link-text" component="small">
      <Link to={props.to as string}>{props.linkText}</Link>
    </Content>
  </FlexItem>
)}
```

  2. Switch vertical alignment
I think this one's already handled. The switch FlexItem has `alignSelf={{ default: 'alignSelfCenter' }}` on it. I checked locally and didn't notice a visual difference. 

  3. Dropping isCompact
Makes sense. Dropping isCompact will bump the padding from 16px to 24px, which matches the other careds. Easy one-prop removal.

  4. Inner KPI card sizing
  +1 on deferring to a separate PR to keep this one focused.

tiyiprh:
> 4. Inner KPI card sizing
>    +1 on deferring to a separate PR to keep this one focused.

Created issue here: https://redhat.atlassian.net/browse/AAP-85859

SherinV:
This will go in after the 8/17 GA - https://redhat.atlassian.net/browse/AAP-85062?focusedCommentId=17879465
It can merge to devel, but backport will have to be after 8/17 GA @marshmalien @kcagran

keithjgrant:
@kcagran Can you rebase this to pick up some test fixes that have merged recently?

keithjgrant (approved):

PR #3469, opened 2026-08-12
fix(deps): update to msw 2.15.0

acruzgon:
## Summary

Hand upgrade msw to resolve Vitest test failures in dependabot bulk PR #3464.

**Root cause:** MSW 2.15.0 depends on `rettime` which is ESM-only. When Vitest's server tried to load it, `setupServer` silently failed to intercept HTTP requests, causing all tests to hit the real network (`ECONNREFUSED ::1:3000`).

**Fix:** Add `msw` and `rettime` to `server.deps.inline` in all workspace Vitest configs so they get bundled instead of loaded via `require()`.

## Type of Change

- [x] Bug fix

## Risk Analysis - REQUIRED

### DIFF
--- a/framework/vite.config.ts
+++ b/framework/vite.config.ts
@@ -52,7 +52,7 @@ export default defineConfig({
     setupFiles: ['vitest.setup.ts'],
     server: {
       deps: {
-        inline: ['@patternfly/react-styles'],
+        inline: ['@patternfly/react-styles', 'msw', 'rettime'],
       },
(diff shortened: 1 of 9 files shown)


acruzgon:
Closing — fix doesn't work (rettime ESM issue is deeper than a deps.inline fix). Will take the inverse approach: extract the 15 working packages into a separate PR and leave msw for manual investigation.

PR #3503, opened 2026-08-25
Shrink CLAUDE.md to a skill trigger table

nixocio:
## Summary

`AGENTS.md` is a symlink to `CLAUDE.md`. Cursor was loading a ~760-line instruction dump on every session. This replaces `CLAUDE.md` with a short trigger table (org pack shape) and moves leftover guidance into skills.

### Changes

- **Short `CLAUDE.md`:** trigger table, MCP notes, lint/test done-bar. `AGENTS.md` still points at it.
- **New skills:** `frontend-playwright-e2e` (write, run, and debug Playwright E2E in one skill; never print secrets), `frontend-overlay` (framework wrappers, API helpers, Playwright paths).
- **Existing skills:** drop "see CLAUDE.md" pointers; i18n rules live in `coding_standards`; E2E triggers load the Playwright skills instead of `testing_guidelines`.
- Existing skill filenames are unchanged. No Syntara wrappers, Storybook-as-SoT, or React 19.

## Type of Change

- [ ] Bug fix
- [ ] Enhancement
- [ ] Tests
- [x] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

## Dependencies

None. Independent of application code. Will conflict with any open PR that rewrites `CLAUDE.md` (rebase after that lands).

### Manual Testing

Docs and agent-config only. Confirmed `AGENTS.md` still symlinks to `CLAUDE.md` (~34 lines).


.claude/skills/frontend-overlay/SKILL.md
```diff
@@ -0,0 +1,108 @@
...
+  Use when implementing or reviewing UI in this monorepo.
+---
+
+# Overlay — ansible-ui
+
+## Stack
+
+- React version: 18.3.x (not 19 — no `ref`-as-prop, no `use(Context)`)
+- PatternFly version: 6.5.x
+- Node: 20.x+, NPM 8.x+
```
nicklemmon: Do we need to reference version numbers here that are legible from the `package.json`?

(We might in this case! I'm genuinely not sure)

.claude/skills/frontend-overlay/SKILL.md
```diff
@@ -0,0 +1,108 @@
...
+- API helpers: `awxAPI` / `edaAPI` / `hubAPI` / `gatewayAPI` (not a generated typed client)
+- Mock API handlers: MSW in Vitest. Playwright also has a mock project
+- E2E: `playwright/` (`playwright.config.ts`, `commands/`, `tests/`, `utils/`)
+- Storybook command and port: N/A — no Storybook
+- Check command: `npm test` (eslint + tsc + prettier + vitest). There is no `npm run check`
+- Test command: `npm run vitest` (unit); Playwright from `playwright/`
+- Instruction files: `CLAUDE.md` (symlink `AGENTS.md`)
+
+## Wrappers (use these, not raw PatternFly)
+
```
nicklemmon, line 36: We should point the agent the location of the global components (the framework package)

.claude/skills/frontend-overlay/SKILL.md
```diff
@@ -0,0 +1,108 @@
...
+| Empty (no data / no filter / error) | framework empty states | |
+| Confirmation | framework dialog / PF Modal | Reversible vs destructive |
+| Error with retry | workspace error adapter | See coding_standards |
+| Forms | `AwxPageForm` / `EdaPageForm` / `HubPageForm` / `PlatformPageForm` | Never raw `PageForm` |
+| Toast / alert helper | framework alerts | object form `{ title, description? }` |
+
+## API
+
+- Call the backend with workspace tagged templates + SWR / CRUD hooks (`useGet`, `usePostRequest`, …)
+- Forbidden: hardcoded `/api/...` paths; mocking `requestGet` instead of MSW
```
nicklemmon: Is this forbidden by an ESLint rule already? If not, we should follow-up with one

.claude/skills/frontend-overlay/SKILL.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Permissions
+
+- Hook names: workspace RBAC helpers in coding_standards
+- Disabled-with-tooltip: existing page action patterns
+- Nav / route guards: workspace routing
+
+## Icons
+
+- PatternFly icons / existing framework icons. No Syntara `RhUi*` / `Nx*` prefix
```
nicklemmon: I don't think we need to reference Syntara here!

Are the `RhUi` icons forbidden? If so, we should ad a follow-up ESLint rule

.claude/skills/coding_standards.md
```diff
@@ -579,3 +579,11 @@ export function MyComponent(props: Readonly<MyComponentProps>) { ... }
 ```
 
 This is enforced by SonarCloud rule S6759.
+
+## 18. File naming
+
+- Test files: `*.spec.ts` (Playwright) or `*.test.ts` / `*.test.tsx` (Vitest)
+- Component files: PascalCase (e.g. `UserTable.tsx`)
+- Utility files: camelCase (e.g. `apiHelpers.ts`)
```
nicklemmon, line 587: For follow-up: I want to explore more consistent file name casing across both major projects in the near future. I propose `kebab-case` for *every* file type. This should help agents grep more efficiently.

In a monorepo like Syntara, this would also align well with co-located python files.

.claude/skills/testing_guidelines.md
```diff
@@ -411,3 +409,212 @@ import { hubAPI } from '../../common/api/formatPath';
 // npm run vitest -- --watch   — watch mode
 // npm run vitest-preview      — vitest UI
 ```
+
+---
+
+## Playwright Testing
+
+Moved here from `CLAUDE.md` / `AGENTS.md`. Default to Vitest for unit/component work.
```
nicklemmon: > Moved here from `CLAUDE.md` / `AGENTS.md`

Unnecessary phrase

.claude/skills/testing_guidelines.md
```diff
@@ -411,3 +409,212 @@ import { hubAPI } from '../../common/api/formatPath';
...
+
+```bash
+export PLATFORM_SERVER='https://localhost:443'
+# Standalone services if needed
+export AWX_SERVER='https://localhost:8043'
+export EDA_SERVER='http://localhost:8000'
+export HUB_SERVER='http://localhost:5001'
+```
+
+Prerequisites: Node.js 20.x+, NPM 8.x+. Live UI is typically `https://localhost:4100`.
```
nicklemmon: NPM 8?? Pretty old! Is that true for this project?

CLAUDE.md
```diff
@@ -1,759 +1,33 @@
...
-```
-
-#### Allowed String Comparisons
-
-These types of strings are **safe** to use in logic (they won't be translated):
-
-- **API response values**: `status === 'successful'`, `type === 'job_template'`, `kind === 'playbook'`
-- **Route paths**: `pathname === '/organizations'`, `path.includes('/settings')`
-- **Internal constants**: `mode === 'edit'`, `view === 'list'`
-- **Technical identifiers**: `file.endsWith('.yaml')`, `name.startsWith('demo_')`
```
nicklemmon, line 447: Does this need to be captured somewhere? An ESLint rule?

nicklemmon (changes requested):
@nixocio Left some feedback.

General feedback: were all of the removals from `CLAUDE.md` captured in another way?

.claude/skills/frontend-overlay/SKILL.md
nixocio: Good call — dropped the hardcoded minor versions and now point to `package.json` for exact React/PF/Node. Kept only the behavioral "React 18, not 19" note since it drives which APIs are safe to use.

.claude/skills/frontend-overlay/SKILL.md
nixocio: Added a pointer under Wrappers: global/shared components live in the `framework/` package — search there first before raw PatternFly or a new component.

.claude/skills/frontend-overlay/SKILL.md
nixocio: It is enforced — a custom ESLint rule already blocks hardcoded `/api/...` paths (documented in coding_standards). I annotated the line to say so. The "mock via MSW, not requestGet" half is not lint-enforced; happy to file a follow-up for a rule there.

.claude/skills/frontend-overlay/SKILL.md
nixocio: Removed the Syntara reference — now just "PatternFly icons or existing framework icons in `framework/`". The `RhUi*`/`Nx*` prefixes are Syntara-only and do not exist in this repo, so no rule is needed here.

.claude/skills/coding_standards.md
nixocio: Agreed — I like the kebab-case-everywhere direction for grep-ability plus python co-location in a monorepo. That is a cross-repo convention change though, so out of scope for this docs PR. Worth its own follow-up issue; current casing (PascalCase components / camelCase utils) stays documented for now.

.claude/skills/testing_guidelines.md
nixocio: Removed — that whole Playwright section moved into the new `frontend-playwright-e2e` skill, and the "Moved here from CLAUDE.md / AGENTS.md" phrasing is gone.

.claude/skills/testing_guidelines.md
nixocio: Fixed — `engines` only pins `node >=20` (no npm version at all), so I dropped the "NPM 8.x+" claim; it now reads "Node.js 20+". This content also moved into the new `frontend-playwright-e2e` skill.

CLAUDE.md
nixocio: This guidance is captured in coding_standards §16 (Internationalization): "Never compare translated strings in logic — compare raw API values, enums, routes, or other non-translated IDs." On enforcement: the `i18next/no-literal-string` ESLint rule covers the JSX side (user-facing strings must use `t()`), but the "do not compare translated strings in logic" nuance is not separately lintable, so it stays a documented guideline.

nicklemmon (dismissed):

nicklemmon (approved):

PR #3507, opened 2026-08-26
Add knip as an advisory unused-code report on pull requests

nixocio:
## Summary
- Add knip and run it on every PR so unused files and unused dependencies show up in CI.
- The knip job is advisory (`continue-on-error`) and uploads a `knip-report` artifact. It does not block merge.
- Webpack/Vite packages knip cannot see (string loaders, monaco plugin) stay ignored so the report is real leftovers.
- When the report is small enough, drop `continue-on-error` on the knip step to make it a merge gate.

## Type of Change

- [ ] Bug fix
- [x] Enhancement
- [ ] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

## Test plan
- [ ] PR Checks includes a `knip` job
- [ ] knip logs unused files / unused deps (or is empty after cleanup)
- [ ] `knip-report` artifact is present
- [ ] A knip finding does not fail **All checks passed**
- [ ] `npm run knip` locally matches the same report


package.json
```diff
@@ -185,6 +187,7 @@
     "html-webpack-plugin": "5.6.3",
     "husky": "9.1.7",
     "klona": "2.0.6",
+    "knip": "^6.32.2",
```
nicklemmon: Looks like a new version was just released - update to `6.34.0`

package.json
```diff
@@ -46,6 +46,8 @@
     "tsc": "nx run-many --target=tsc",
     "prettier": "nx run-many --target=prettier",
     "prettier:fix": "nx run-many --target=prettier:fix",
+    "knip": "knip",
+    "knip:files": "knip --include files",
```
nicklemmon, line 50: What is this used for exactly? Doesn't seem to be run in the CI workflow so could probably be removed.

nicklemmon (dismissed):

package.json
nixocio: Updated to ^6.34.0 in package.json and package-lock.json (commit f9c6903f0).

package.json
nixocio: It is used by the new PR workflow knip job via npm run knip; its report and summary remain advisory.

.github/workflows/pull-request.yml
```diff
@@ -58,6 +58,61 @@ jobs:
...
       - run: npx nx affected -t ${{ matrix.script }} --base=origin/main --head=HEAD --verbose
 
+  knip:
+    name: knip
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    needs: packages
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
```
nicklemmon: Sorry, should have caught this earlier - these should use pinned SHAs/versions.

Worth running these changes through `zizmor` and `actionlint` as well.

nicklemmon (changes requested):

.github/workflows/pull-request.yml
nixocio: Pinned the four actions in the new Knip job to immutable v4 SHAs, and set read-only contents permissions plus `persist-credentials: false` for checkout. `actionlint` passes; I also ran zizmor. Its remaining findings are pre-existing elsewhere in this workflow.

nicklemmon (approved):

PR #3521, opened 2026-08-31
fix(e2e): prevent 3-minute timeout in credential Team/User Access tests

acruzgon:
## Summary

Fix credential Team and User Access Playwright tests that timed out (180s) in CI when the unfiltered credentials list returned HTTP 500. The error empty state hides the **Create credential** button, causing every test that starts with `Credential.ui.create` to hang.

- Create credentials, teams, and users via API in `beforeEach`/`afterEach` (matching the pattern in `host-groups.spec.ts`, `eda-user-access.spec.ts`, etc.)
- Navigate to credential details by id using client-side routing (`Credential.ui.open`) instead of the fragile list → filter → click path
- Add `Credential.api.create` / `Credential.api.delete` to `playwright/utils/credential.ts`
- Add fail-fast guard to `Credential.ui.create` so any remaining callers get an immediate error instead of a 3-minute timeout

## Type of Change

- [ ] Bug fix
- [ ] Enhancement
- [x] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

### Ephemeral E2E Tests

`/run-playwright` once CI is green.

### External Server E2E Runs

All 5 Team and User Access tests pass locally against a live 4.8.6 controller (~17-23s each, down from 180s timeout).

### Manual Testing

```bash
cd playwright
npx playwright test tests/integration/automation-execution/infrastructure/credentials/credentials.spec.ts \
  --project 'live chromium' --grep 'Credentials - Team and User Access' --retries=0
```

### DIFF
--- a/playwright/tests/integration/automation-execution/infrastructure/credentials/credentials.spec.ts
+++ b/playwright/tests/integration/automation-execution/infrastructure/credentials/credentials.spec.ts
@@ -1,11 +1,15 @@
+import { Credential as CredentialType } from '@ansible/awx-ui/interfaces/Credential';
+import { PlatformTeam } from '@ansible/platform-ui/interfaces/PlatformTeam';
+import { PlatformUser } from '@ansible/platform-ui/interfaces/PlatformUser';
 import { expect, test } from '@playwright/test';
+import { gatewayAPI } from '../../../../../commands/apiClient';
 import { clickPageAction } from '../../../../../commands/clickPageAction';
 import { clickTableRow } from '../../../../../commands/clickTableRow';
 import { createE2EName } from '../../../../../commands/createE2EName';
 import { filterTable } from '../../../../../commands/filterTable';
 import { navigateTo } from '../../../../../commands/navigateTo';
 import { setupAfter, setupBefore } from '../../../../../commands/setup';
-import { Credential } from '@ansible/playwright/utils';
+import { Credential, Team, User } from '@ansible/playwright/utils';
 
 test.beforeEach(setupBefore({ path: '/execution/infrastructure/credentials' }));
 test.afterEach(setupAfter);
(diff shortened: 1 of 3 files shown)


acruzgon:
image
Credentials - Team and User Acces tests all passing now

acruzgon:
image

Credentials - Team and User Access tests all passing now!

keithjgrant (dismissed):

keithjgrant (approved):

nixocio:
Reviewed against `.claude/skills/frontend-playwright-e2e/SKILL.md` and `.claude/skills/pr_review.md` (scope = this diff only).

The root-cause change is right: Team/User Access was hanging on `Credential.ui.create` because GET `/credentials/` 500 hides **Create credential**. API create + `Credential.ui.open` (skip the list) matches other live specs (`Resource.api.create`). Fail-fast on `Error loading credentials` is better than a 180s timeout bump. Currents on this PR already has `credentials.spec.ts` **22/22**.

Non-blocking, skill alignment:

1. **Table rows** — Playwright skill §3 says never `getByRole('row').filter({ hasText })` (pagination). This spec already had that; the PR still uses it after `filterTable` (remove-role / manage-roles). The same tests already use `getByRole('row', { name: new RegExp(...) })` in the assign wizard — prefer that or `clickTableRow` / `getTableRow` for the remove path too.

2. **`organization: 1`** — `Team.api.create(page, { organization: 1 })` and `Credential.api.create` defaulting `organization: 1` assume Default is id 1. Other live tests create an org and pass `organization.id` (see the jobs SCM PR). Safer if this starts failing on a non-id-1 gateway org.

3. **`history.pushState` + `popstate`** — new nav helper (pr_review “re-invented pattern”). Fine while the heading wait is the source of truth. If `open()` flakes, don’t copy this per resource; add one router helper or use `page.goto` with a tight wait.

Trigger note: `/run-playwright` must have **no trailing space** (this PR’s comment had one; it still ran).

PR #3522, opened 2026-08-31
Stabilize live inventory and credential-type Playwright assertions

nixocio:
## Summary

- Wait for constructed-inventory sync jobs, then assert `inventory-sources-with-active-failures` instead of `last-job-status` (that detail does not render until inventory sources refresh).
- Assert inventory source Details `inventory-path` (label is Inventory path, not Inventory file).
- Wait for all group rows before bulk delete, then wait for an empty table.
- Clear selected Monaco contents before typing replacement text so all shared-helper callers avoid native-edit-context concatenation.
- Extract the reusable constructed-inventory sync/capture/poll sequence into `playwright/commands`.

Does not address unrelated RBAC wizard, credentials, jobs, or EDA timeouts.

## Type of Change

- [x] Bug fix
- [ ] Enhancement
- [x] Tests
- [ ] Documentation
- [ ] Other

## Risk Analysis - REQUIRED

- [x] **Medium** — Scoped but cross-cutting
- [ ] **Low** — Narrowly scoped

The production code is unchanged. The shared Playwright helper change is one line but affects 25 callers across 15 files.

### Local aap-dev

Live Chromium with retries disabled:

- constructed-inventory spec: 7/7 passed (1.3m)
- credential-type JSON target: 4/4 passed including setup/teardown (18.1s)
- regular-inventory edit representative shared-helper consumer: 4/4 passed including setup/teardown (29.1s)

### Ephemeral E2E

Trigger with `/run-playwright` and classify failures against the changed specs; unrelated suite failures are outside this PR.

### DIFF
--- a/playwright/commands/syncConstructedInventoryAndWait.ts
+++ b/playwright/commands/syncConstructedInventoryAndWait.ts
@@ -1,0 +1,27 @@
+import { expect, Page } from '@playwright/test';
+import { waitForJobStatus } from './waitForJobStatus';
+
+export async function syncConstructedInventoryAndWait(
+  page: Page,
+  desiredStatus: 'successful' | 'failed'
+) {
+  const syncResponsePromise = page.waitForResponse(
+    (response) =>
+      response.url().includes('/inventory_sources/') &&
+      response.url().includes('/update/') &&
+      response.request().method() === 'POST'
+  );
+  await page.getByRole('button', { name: 'Sync inventory' }).click();
+  const syncResponse = await syncResponsePromise;
+  expect(syncResponse.status()).toBe(202);
+  const inventoryUpdate = (await syncResponse.json()) as { id: number };
+  await waitForJobStatus(
+    {
+      jobType: 'inventory_updates',
+      jobId: inventoryUpdate.id,
+      desiredStatus: desiredStatus === 'failed' ? ['failed', 'error'] : 'successful',
+      timeout: 60000,
+    },
+    page
+  );
+}
--- a/playwright/tests/integration/automation-execution/infrastructure/inventories/constructed/constructed-inventory-crud.spec.ts
+++ b/playwright/tests/integration/automation-execution/infrastructure/inventories/constructed/constructed-inventory-crud.spec.ts
@@ -7,6 +7,7 @@ import { fillMonacoEditor } from '../../../../../../commands/fillMonacoEditor';
 import { filterTableByText } from '../../../../../../commands/filterTableByText';
 import { navigateTo } from '../../../../../../commands/navigateTo';
 import { setupAfter, setupBefore } from '../../../../../../commands/setup';
+import { syncConstructedInventoryAndWait } from '../../../../../../commands/syncConstructedInventoryAndWait';
 
 test.beforeEach(setupBefore({ path: '/execution/infrastructure/inventories' }));
 test.afterEach(setupAfter);
@@ -122,19 +123,11 @@ test.describe('Constructed Inventory', () => {
         ).toBeVisible();
         await expect(page.getByTestId('description')).toHaveText(description);
 
-        // Sync inventory
-        const syncResponsePromise = page.waitForResponse(
-          (response) =>
-            response.url().includes('/inventory_sources/') &&
-            response.url().includes('/update/') &&
-            response.request().method() === 'POST'
-        );
-        await page.getByRole('button', { name: 'Sync inventory' }).click();
-        await syncResponsePromise;
-
-        // Wait for sync to complete and verify success status
-        await expect(page.getByTestId('last-job-status')).toContainText('Success', {
-          timeout: 60000,
+        await syncConstructedInventoryAndWait(page, 'successful');
+        // Reload so Inventory sources with active failures reflects the finished update.
+        await page.reload();
+        await expect(page.getByTestId('inventory-sources-with-active-failures')).toHaveText('0', {
+          timeout: 15000,
         });
       } finally {
         // Cleanup - Organization.api.deleteByName handles dependent inventories
@@ -195,19 +188,11 @@ test.describe('Constructed Inventory', () => {
           page.getByRole('heading', { name: constructedInventoryName, exact: true })
         ).toBeVisible({ timeout: 10000 });
 
-        // Trigger sync
-        const syncResponsePromise = page.waitForResponse(
-          (response) =>
-            response.url().includes('/inventory_sources/') &&
-            response.url().includes('/update/') &&
-            response.request().method() === 'POST'
-        );
-        await page.getByRole('button', { name: 'Sync inventory' }).click();
-        await syncResponsePromise;
-
-        // Wait for sync to complete and verify failed status
-        await expect(page.getByTestId('last-job-status')).toContainText('Failed', {
-          timeout: 30000,
+        await syncConstructedInventoryAndWait(page, 'failed');
+        // Reload so Inventory sources with active failures reflects the finished update.
+        await page.reload();
+        await expect(page.getByTestId('inventory-sources-with-active-failures')).toHaveText('1', {
+          timeout: 15000,
         });
       } finally {
         // Cleanup - Organization.api.deleteByName handles dependent inventories
--- a/playwright/tests/integration/automation-execution/infrastructure/inventories/tabs/inventory-groups.spec.ts
+++ b/playwright/tests/integration/automation-execution/infrastructure/inventories/tabs/inventory-groups.spec.ts
@@ -98,13 +98,24 @@ test.describe('Inventory Groups - List View', () => {
 
   test('can bulk delete multiple groups', { tag: ['@not_mock'] }, async ({ page }) => {
     const inventoryName = await Inventory.ui.create(page);
-    await InventoryGroup.ui.createGroup(page, { inventoryName });
-    await InventoryGroup.ui.createGroup(page, { inventoryName });
-    await InventoryGroup.ui.createGroup(page, { inventoryName });
+    const groupNames = [
+      await InventoryGroup.ui.createGroup(page, { inventoryName }),
+      await InventoryGroup.ui.createGroup(page, { inventoryName }),
+      await InventoryGroup.ui.createGroup(page, { inventoryName }),
+    ];
 
     await navigateTo(page, 'Automation Execution', 'Infrastructure', 'Inventories');
     await clickTableRow({ text: inventoryName }, page);
     await page.getByRole('tab', { name: 'Groups' }).click();
+
+    await clearTableFilters(page);
+    const groupRows = groupNames.map((groupName) =>
+      page.getByRole('row').filter({ hasText: groupName })
+    );
+
+    for (const groupRow of groupRows) {
+      await expect(groupRow).toBeVisible();
+    }
 
     await page.getByRole('checkbox', { name: 'Select all' }).check();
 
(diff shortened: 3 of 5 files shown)


playwright/tests/integration/automation-execution/infrastructure/credential-types/credential-types.spec.ts
```diff
@@ -284,11 +283,14 @@ test.describe('Credential Types', { tag: ['@not_mock', '@local_debug'] }, () =>
           fields: [{ id: 'api_key', type: 'string', label: 'API Key' }],
         });
 
-        await fillMonacoEditor(
-          page,
-          inputConfig,
-          page.locator('.monaco-editor').first().getByRole('textbox')
-        );
+        await page
```
acruzgon: I'd rather keep `fillMonacoEditor` here than swap it for the dropzone input.

`#code-editor-dropzone-input` is a hidden file upload, not the editor, so this is a different path than what the test is meant to cover. It's also only changed in these two specs — create still goes through `fillMonacoEditor`, and so does everything else.

If select-all/type isn't replacing the editor contents, that's a helper bug. Let's fix `fillMonacoEditor` so every caller gets it, instead of special-casing dropzone in a couple of tests.

acruzgon (changes requested):

playwright/tests/integration/automation-execution/infrastructure/inventories/constructed/constructed-inventory-crud.spec.ts
```diff
@@ -1,12 +1,47 @@
...
 import { clickPageAction } from '../../../../../../commands/clickPageAction';
 import { confirmAndAssertDeletion } from '../../../../../../commands/confirmAndAssertDeletion';
 import { createE2EName } from '../../../../../../commands/createE2EName';
-import { fillMonacoEditor } from '../../../../../../commands/fillMonacoEditor';
 import { filterTableByText } from '../../../../../../commands/filterTableByText';
 import { navigateTo } from '../../../../../../commands/navigateTo';
 import { setupAfter, setupBefore } from '../../../../../../commands/setup';
+import { waitForJobStatus } from '../../../../../../commands/waitForJobStatus';
+
+async function syncConstructedInventoryAndWait(page: Page, desiredStatus: 'successful' | 'failed') {
```
acruzgon: can we put syncConstructedInventoryAndWait in playwright/commands instead of defining it in the spec?

acruzgon (changes requested):

playwright/tests/integration/automation-execution/infrastructure/inventories/constructed/constructed-inventory-crud.spec.ts
nixocio: Done in fb203f974. The reusable sync/capture/poll sequence now lives in playwright/commands/syncConstructedInventoryAndWait.ts. The reload remains in this spec because it specifically refreshes the visible active-failures count.

playwright/tests/integration/automation-execution/infrastructure/credential-types/credential-types.spec.ts
nixocio: Done in the combined PR. Both affected flows use fillMonacoEditor, and the shared helper now clears selected content before typing. Local aap-dev validation passed for constructed inventory (7/7), the credential JSON target (4/4), and a representative regular-inventory helper consumer (4/4), with retries disabled.

nixocio:
Rebased onto current `devel` and fixed the Monaco 0.56 regression found in the previous Currents run. The editor replacement remains explicit, while the EDA test now asserts rendered `.view-lines` content instead of the empty native edit-context shim. The constructed sync helper now captures the exact POST first and asserts `202` afterward for an actionable failure rather than a response-wait timeout.

Local aap-dev validation, live Chromium, no retries:
- EDA credential type create/edit regression: 4/4 passed
- constructed inventories: all 4 test bodies passed, including successful and expected-failed sync
- inventory groups changed bulk-delete test passed
- inventory source changed edit/details test: 4/4 passed
- AWX credential-type JSON and YAML Monaco tests passed

Fresh `/run-playwright` requested on `3bea44220`.

playwright/commands/syncConstructedInventoryAndWait.ts
```diff
@@ -0,0 +1,28 @@
...
+  const syncResponsePromise = page.waitForResponse(
+    (response) =>
+      response.url().includes('/inventory_sources/') &&
+      response.url().includes('/update/') &&
+      response.request().method() === 'POST'
+  );
+  await page.getByRole('button', { name: 'Sync inventory' }).click();
+  const syncResponse = await syncResponsePromise;
+  expect(syncResponse.status()).toBe(202);
+  const inventoryUpdate = (await syncResponse.json()) as { id: number };
```
lgalis, line 17: throwOnFailure: desiredStatus !== 'failed' only recognizes the exact status 'failed'. If AWX ever returns 'error' or 'canceled' instead (a
  plausible terminal state for a bad constructed-inventory YAML), the helper won't throw early — it'll silently poll until the full 60s timeout, then fail with a generic timeout message instead of surfacing the
  real status. Fix: widen the desired-status set to ['failed', 'error', 'canceled'] and drop the throwOnFailure override

nixocio: Addressed in `d8c4b0aec`. The expected-failure path now accepts `failed`, `error`, or `canceled`, and the `throwOnFailure` override is removed. Successful syncs retain the default fail-fast behavior. Scoped Prettier and Playwright TypeScript checks pass.

nixocio:
Rebased onto current `devel` (`d995e18f7`). The last Konflux E2E failure was **5 unrelated flaky timeouts** — all in EDA `automation-decisions` (rulebook-activations edit/auto-restart, EDA credentials) and access-management flows (org team-management, EE team access) — pages timing out on `toBeVisible()` for the list heading, i.e. slow EDA responses. None of them are in the specs this PR touches (credential-types, constructed-inventory, the Monaco/sync helpers), and all of this PR's own specs passed (506/532 green). Re-triggering a fresh run.

lgalis (dismissed):
Approving the inventory/sync/testid work.

**Finding 1** (`fillMonacoEditor` still using `keyboard.type()` after Backspace) is not a merge blocker for this PR. It will be fixed in #3528, which replaces `type()` with clipboard paste so Monaco auto-pairing does not corrupt JSON/YAML.

playwright/commands/syncConstructedInventoryAndWait.ts
```diff
@@ -0,0 +1,27 @@
...
+  await waitForJobStatus(
+    {
+      jobType: 'inventory_updates',
+      jobId: inventoryUpdate.id,
+      desiredStatus: desiredStatus === 'failed' ? ['failed', 'error', 'canceled'] : 'successful',
```
lgalis: **Low (non-blocking):** `desiredStatus === 'failed'` waits for `['failed', 'error', 'canceled']`. A canceled job (infra/worker) satisfies the helper, then the spec asserts `inventory-sources-with-active-failures === '1'`. If a cancel increments that count, the “strict YAML should fail” test can pass without the constructed plugin actually failing.

Prefer waiting for `failed` / `error` and leaving `canceled` as a fail-fast unexpected terminal state.

playwright/tests/integration/automation-execution/infrastructure/inventories/tabs/inventory-groups.spec.ts
```diff
@@ -106,6 +106,9 @@ test.describe('Inventory Groups - List View', () => {
     await clickTableRow({ text: inventoryName }, page);
     await page.getByRole('tab', { name: 'Groups' }).click();
 
+    const groupRows = page.getByRole('row', { name: /E2E group/ });
```
lgalis: **Low (non-blocking):** Waiting for three rows before Select all is the right idea. Counting `getByRole('row', { name: /E2E group/ })` is the pagination-fragile pattern we usually avoid in favor of `getTableRow()` / the names returned from `createGroup`.

Unlikely to flake here (dedicated inventory, three rows), so not a merge blocker.

lgalis (commented):
Non-blocking follow-up from the review: two low findings inline. Neither should hold merge.

- `syncConstructedInventoryAndWait`: treating `canceled` as an expected failure can hide a canceled job as a passing “strict YAML failed” test.
- Inventory groups bulk-delete: `getByRole('row', { name: /E2E group/ })` is weaker than `getTableRow()`, but fine for this dedicated 3-row table.

playwright/commands/syncConstructedInventoryAndWait.ts
nixocio: Addressed in 48b561439: expected invalid-YAML terminal states are now failed/error only; canceled remains unexpected and fails fast.

playwright/tests/integration/automation-execution/infrastructure/inventories/tabs/inventory-groups.spec.ts
nixocio: Addressed in 48b561439: retain the three generated names, verify each through getTableRow(), then clear filters before Select all.

acruzgon (dismissed):
## Code Review: Stabilize live inventory and credential-type Playwright assertions

**Score**: 9.8/10 (Functionality 10, Security 10, Quality 9.5)

No critical or major findings. All prior review comments have been addressed:
- `fillMonacoEditor` fixed with Backspace after Ctrl+A instead of dropzone workaround
- `syncConstructedInventoryAndWait` extracted to `playwright/commands/`
- Terminal state handling tightened to `failed`/`error` only (no `canceled`)
- `getTableRow()` used for group verification before bulk delete

One minor observation: `#injectors .view-lines` in credential-types-crud relies on Monaco internal DOM structure — worth a comment for future maintainability.

lgalis (dismissed):
1. Low — bulk-delete waits while filtered, then Select-all without waiting for the unfiltered rows
getTableRow() always filterTables by name. The new pre-delete loop therefore never shows all three groups at once. After the last name, clearTableFilters only clicks “Clear all filters” if it is visible — it does not wait for the table to reload — and the next line is Select all:

for (const groupName of groupNames) {
  await expect(await getTableRow(page, groupName)).toBeVisible();
}
await clearTableFilters(page);
await page.getByRole('checkbox', { name: 'Select all' }).check();
If Select all runs while the last name filter is still applied, bulk delete can remove one group. The later toHaveCount(0) loop would then fail (a flake, not a false pass).

After clearTableFilters, wait for the three names on the unfiltered table (do not call getTableRow again, since that re-filters):

await clearTableFilters(page);
for (const groupName of groupNames) {
  await expect(page.getByRole('row').filter({ hasText: groupName })).toBeVisible();
}
Same leftover-filter issue after delete: the last getTableRow leaves a name filter on, so the empty-state regex will usually match No results found, not an unfiltered empty inventory. Restore clearTableFilters after the count-0 loop if that empty-state check is meant to prove the inventory is empty.

Not a merge blocker for a dedicated three-row table, but it undercuts the stabilization this hunk is trying to do.

2. Nit — #injectors .view-lines couples to Monaco internals
The switch from textbox[name="Editor content"] / not.toBeEmpty() to #injectors .view-lines + extra_vars: is the right assertion for Monaco 0.56 (the native-edit-context textbox stays empty). It depends on Monaco’s .view-lines class. A one-line comment would help the next person not “fix” it back to the textbox.

acruzgon (approved):
## Code Review: Stabilize live inventory and credential-type Playwright assertions (Iteration 2)

**Score**: 9.8/10 (Functionality 10, Security 10, Quality 9.5)

Re-review after squash and scope reduction. No critical or major findings.

**Changes since prior review:**
- Squashed to single commit — clean history
- `fillMonacoEditor` Backspace fix moved to #3528
- `ScheduleEditWizard.test.tsx` async fix already merged to devel
- `inventory-groups.spec.ts` reworked: `clearTableFilters()` + DOM-only `getByRole('row').filter()` locators instead of `getTableRow()` — addresses the filter-side-effect race

**Prior items resolved:**
- ✅ `syncConstructedInventoryAndWait` in `playwright/commands/` — correct
- ✅ Terminal states `failed`/`error` only — `canceled` is fail-fast
- ✅ Filter race in bulk-delete fixed with DOM-only locators

One minor note: `#injectors .view-lines` in credential-types-crud.spec.ts depends on Monaco internal DOM — a brief comment would help maintainability.

PR #3523, opened 2026-08-31
fix(e2e): fix Source Control Update job test timeout

acruzgon:
## Summary

Fix the "can launch a Source Control Update job" test that timed out (180s) in CI. `Project.ui.create` defaults its SCM URL to `https://github.com/ansible/ansible-ui` — syncing this entire monorepo on ephemeral CI is too slow. Switched to `Project.api.create` which defaults to `https://github.com/ansible/ansible-tower-samples` (tiny repo), bringing the test from 180s timeout down to ~35s.

## Type of Change

- [ ] Bug fix
- [ ] Enhancement
- [x] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

### Ephemeral E2E Tests

`/run-playwright` once CI is green.

### External Server E2E Runs

Test passes locally in ~35s against a live 4.8.6 controller.

### Manual Testing

```bash
cd playwright
npx playwright test tests/integration/automation-execution/jobs/jobs.spec.ts \
  --project 'live chromium' \
  --grep 'can launch a Source Control Update job' \
  --max-failures=1 --retries=0
```


nixocio:
Reviewed against `.claude/skills/frontend-playwright-e2e/SKILL.md` and `.claude/skills/pr_review.md`. Diff is one test in `jobs.spec.ts` — scope matches the PR body.

Root cause is right, not a timeout bump: `Project.ui.create` defaults SCM to `https://github.com/ansible/ansible-ui` (this monorepo); `Project.api.create` defaults to `ansible-tower-samples` and `Project.api.sync` already `waitForJobStatus` on 202. `clickTableRow` is the skill’s table helper (not `getByRole('row').filter`). No `eslint-disable` / `@ts-ignore` in the diff.

Nits:

- `test.setTimeout(180000)` is still on this test. After the SCM URL change, the slow path is the **UI** Sync + `#last-job-status` + 120s `Success` wait, not create. If local/CI is ~35s, drop or lower the 180s so a hang fails sooner.
- Flow still API-syncs, then clicks **Sync project** again. That’s correct if the AC is “launch from the UI”; just don’t treat create as the thing that used to burn the 180s.

No Currents link on this PR yet — `/run-playwright` with no trailing space when you want a live check.

playwright/tests/integration/automation-execution/jobs/jobs.spec.ts
```diff
@@ -177,40 +177,57 @@ test.describe('Jobs: Launch and Verify Output', () => {
...
         },
         page
       );
 
+      const syncResponsePromise = page.waitForResponse(
+        (response) =>
+          response.url().includes('/projects/') &&
+          response.url().includes('/update/') &&
+          response.request().method() === 'POST' &&
+          response.status() === 202
```
nixocio: Please match the project-update POST here, then assert `expect(syncResponse.status()).toBe(202)` after awaiting it. With `202` in the predicate, a real 4xx/5xx response is ignored and the test waits until timeout, hiding the API failure.

playwright/tests/integration/automation-execution/jobs/jobs.spec.ts
```diff
@@ -177,40 +177,57 @@ test.describe('Jobs: Launch and Verify Output', () => {
...
+          timeout: 120000,
+        },
+        page
+      );
+
+      // Navigate to the job output page for the sync we just triggered
+      await navigateTo(page, 'Automation Execution', 'Jobs');
+      await filterTable({ filterLabel: 'ID', filterValue: String(projectUpdate.id) }, page);
+      await expect(page.locator('tbody')).toBeVisible({ timeout: 10000 });
+      await page.locator('tbody').getByRole('link', { name: project.name }).first().click();
```
nixocio: Please reuse `clickTableRow({ text: project.name, filterLabel: "ID", filterValue: String(projectUpdate.id), pageTitle: "Jobs" }, page)` here. It already waits for one filtered row and clicks its link; raw `tbody` plus `.first()` bypasses the repository table helper and can select the wrong link if rendering changes.

playwright/tests/integration/automation-execution/jobs/jobs.spec.ts
```diff
@@ -177,40 +177,57 @@ test.describe('Jobs: Launch and Verify Output', () => {
...
-      // Wait for job to complete
-      await expect(page.getByText('Success', { exact: true }).first()).toBeVisible({
-        timeout: 120000,
-      });
-
-      // Cleanup
-      await Project.ui.delete(page, projectName);
-      await Organization.ui.delete(page, organizationName);
+      await Project.api.delete(page, project.id).catch(() => {});
+      await Organization.api.delete(page, organization.id).catch(() => {});
```
nixocio: Please put resource deletion in a `finally` block and do not swallow cleanup failures. As written, an assertion failure skips cleanup, while these catches can silently leave projects and organizations that destabilize later live tests.

acruzgon (commented):
Addressed all three review comments in 8cb9bfadd:

1. **Status assertion** — moved `202` out of `waitForResponse` predicate into `expect(syncResponse.status()).toBe(202)` so a 4xx/5xx fails immediately instead of timing out.
2. **clickTableRow** — replaced raw `tbody` + `.first()` with `clickTableRow({ text: project.name, filterLabel: 'ID', filterValue: String(projectUpdate.id), pageTitle: 'Jobs' }, page)`.
3. **finally block** — cleanup is now in `try/finally`.

keithjgrant (dismissed):

keithjgrant:
looks like that test is still failing in the latest playwright...?

playwright/tests/integration/automation-execution/jobs/jobs.spec.ts
```diff
@@ -177,40 +177,62 @@ test.describe('Jobs: Launch and Verify Output', () => {
     { tag: ['@not_mock'] },
     async ({ page }) => {
       test.setTimeout(180000);
```
lgalis: This 180s cap can be exceeded under slow CI: `Project.api.sync()` below waits up to 120s, and `waitForJobStatus()` in the test body waits another 120s (240s total). The sibling inventory sync test in this file uses `test.setTimeout(3 * 60 * 1000)` for the same pattern. Consider bumping the timeout here, or skipping the blocking setup sync if the project details page allows "Sync project" on a never-synced project.

playwright/tests/integration/automation-execution/jobs/jobs.spec.ts
```diff
@@ -177,40 +177,62 @@ test.describe('Jobs: Launch and Verify Output', () => {
...
+        );
 
-      // Cleanup
-      await Project.ui.delete(page, projectName);
-      await Organization.ui.delete(page, organizationName);
+        await expect(page).toHaveURL(/\/jobs\/project\/\d+\/output/);
+        await expect(page.getByText('Success', { exact: true }).first()).toBeVisible();
+      } finally {
+        await Project.api.delete(page, project.id).catch(() => {});
+        await Organization.api.delete(page, organization.id).catch(() => {});
```
lgalis: The `finally` block ensures cleanup runs on failure, but `.catch(() => {})` still swallows delete errors — orphaned projects/orgs can destabilize later live tests. The inventory sync test in this file deletes without catching. Consider dropping the `.catch()` handlers (or only ignoring a specific "already deleted" case).

lgalis (commented):
Two minor suggestions on timeout budgeting and cleanup error handling — see inline comments.

acruzgon:
Addressed @lgalis review feedback in 76a7d27f7:

1. **Timeout** — bumped to `5 * 60 * 1000` to cover stacked `Project.api.sync` + `waitForJobStatus` waits under slow CI.
2. **Cleanup** — removed `.catch(() => {})` from `finally` so delete failures surface.

/run-playwright

acruzgon:
Merged latest `devel` and fixed the remaining failure in `ecee34a71`.

**Root cause:** `clickTableRow` failed on the job output page with a strict-mode violation — the project name appears in both the page title (`page-title`) and the job summary heading.

**Fix:** Navigate directly to `/execution/jobs/project/{id}/output` after `waitForJobStatus`, matching the pattern in `workflowVisualizerOutput.spec.ts`.

/run-playwright

playwright/tests/integration/automation-execution/jobs/jobs.spec.ts
```diff
@@ -176,41 +176,63 @@ test.describe('Jobs: Launch and Verify Output', () => {
...
+          page
+        );
 
-      // Wait for job to complete
-      await expect(page.getByText('Success', { exact: true }).first()).toBeVisible({
-        timeout: 120000,
-      });
+        // Navigate to the job output page via the Jobs list (same pattern as inventory sync below)
+        await navigateTo(page, 'Automation Execution', 'Jobs');
+        await filterTable({ filterLabel: 'ID', filterValue: String(projectUpdate.id) }, page);
```
lgalis: After filtering by job ID, the row/link is clicked using project.name. On the Jobs list, the accessible name/link text may not exactly match the project name (unified job naming can differ). Filtering by ID narrows to one row, but the name-based click can still flake.

Consider one of:

clickTableRow after the ID filter (asserts exactly one row), using the job’s display name from the API response
Direct navigation by job ID, as in workflowVisualizerOutput.spec.ts:
page.goto(\${baseUrl}/execution/jobs/project/${projectUpdate.id}/output`)`
The inventory sync test below doesn’t use the Jobs list, so it’s not a great pattern match for this step.

lgalis (dismissed):
non-blocking comments

playwright/tests/integration/automation-execution/jobs/jobs.spec.ts
acruzgon: Thanks for the review — good catch on the flaky project.name click after filtering by ID.

I didn’t use page.goto here. Instead, I followed your first suggestion: after waitForJobStatus completes, the test navigates to Jobs and uses clickTableRow filtered by job ID, with text set to completedProjectUpdate.name from the API response. That avoids assuming the unified job name matches project.name, while still navigating through the UI.

lgalis (dismissed):
LGTM

keithjgrant:
looks like it's failing with `Error: strict mode violation: getByRole('heading', { name: 'E2E Project 8cef6c37' }) resolved to 2 elements:`

lgalis (approved):
LGTM

keithjgrant:
looking at your latest change, it looks like you had to not use the clickTableRow helper function... was that so you could increase the table load timeout? If so, maybe it would be better to add an optional timeout field to the options object for that helper?

nixocio (approved):
LGTM — converts flaky UI-based setup to reliable API-based setup with response interception. Uses `waitForJobStatus` helper, proper cleanup in `finally` block, follows Playwright skills.

PR #3528, opened 2026-09-01
fix(AAP-90404): paste into Monaco via clipboard to avoid JSON corruption

lgalis:
## Summary

`credential-types-crud.spec.ts` ("should create and edit a credential type") timed out waiting for **Generate extra vars**. That button only renders when React Hook Form `inputs` is a parsed object with `fields.length > 0`. The form gets that value from Monaco's `onDidChangeContent` handler.

`fillMonacoEditor` used `page.keyboard.type()`, which sends one keydown per character. Monaco's JSON language auto-closes `{`, `[`, and `"`, so the typed JSON was corrupted, the schema never parsed, the button never appeared, and the click hung until the 60s timeout.

The required fix is to fill Monaco without per-keystroke auto-closing, in a way that still updates the model / RHF.

## What is required for that failure

- **`fillMonacoEditor`**: clipboard paste (`Ctrl+V`) instead of `keyboard.type()`. Paste inserts the whole string, skips auto-closing, and fires content-change handlers so the form stays in sync.
- `keyboard.insertText()` is not used: it dispatches only an `input` event and may not update the Monaco 0.56 native-edit-context model / RHF (see #3468).

Clipboard permissions: this spec is `@not_mock` and runs on `live chromium`, which already had `clipboard-read` / `clipboard-write`. The global config grant is not what unblocks this test; it only matters if other projects call the same helper.

## What is not required for that failure

| Change | Why it is in this PR |
| --- | --- |
| 15s wait for "Generate extra vars" | Fail faster if fill is still wrong. Does not fix the timeout by itself. |
| Empty-string → Backspace | `remotes.spec.ts` calls `fillMonacoEditor(page, '')`. The credential type test never clears the editor that way. |
| Namespace + constructed-inventory `keyboard.type()` → helper | Same auto-pair class of bug on other specs. |
| `ScheduleEditWizard.test.tsx` | Unrelated Vitest flake that failed the `Vitest (awx) / views` CI gate. |
| Clipboard perms on all Playwright projects | Needed for paste on mock/other browsers; `live chromium` already had them. |

## Type of Change

- [x] Bug fix
- [ ] Enhancement
- [x] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

### Ephemeral E2E Tests

Once preliminary checks pass, comment `/run-playwright` on this PR.

### Manual Testing

- [ ] `cd playwright && npx playwright test tests/integration/automation-decisions/credential-types/credential-types-crud.spec.ts --project 'live chromium' --retries=0`
- [ ] Spot-check another `fillMonacoEditor` caller (e.g. `regular-inventory-crud.spec.ts` or `workflow-visualizer.spec.ts`)

### Why clipboard paste (and what we ruled out)

- `keyboard.type()` — auto-closes `{}` / `[]` / `""` and corrupts JSON/YAML.
- `locator.fill()` — fails on Monaco 0.56 `native-edit-context`.
- `keyboard.insertText()` — bulk insert, but may not update the model / RHF (#3468).
- Clipboard paste — bulk insert that still fires Monaco content-change handlers.

### DIFF
--- a/playwright/commands/fillMonacoEditor.ts
+++ b/playwright/commands/fillMonacoEditor.ts
@@ -8,8 +8,8 @@ import { Locator, Page } from '@playwright/test';
  * `<input>`, `<textarea>`, or `[contenteditable]` elements.
  *
  * Strategy varies by browser:
- * - Chromium/Firefox: Clipboard paste (Ctrl+V) bypasses auto-closing brackets
- *   and reliably triggers Monaco's content-change handlers.
+ * - Chromium/Firefox: Clipboard paste (ControlOrMeta+V) bypasses auto-closing
+ *   brackets and reliably triggers Monaco's content-change handlers.
  * - WebKit: Pastes via a temporary textarea (clipboard API is unavailable due
  *   to permission restrictions on non-Chromium browsers). The textarea is
  *   created, filled, copied to clipboard via JS, then pasted into Monaco.
@@ -28,7 +28,7 @@ import { Locator, Page } from '@playwright/test';
 export async function fillMonacoEditor(page: Page, text: string, editorLocator?: Locator) {
   const editor = editorLocator ?? page.getByRole('textbox', { name: 'Editor content' });
   await editor.click({ force: true });
-  await page.keyboard.press('Control+a');
+  await page.keyboard.press('ControlOrMeta+a');
   if (text === '') {
     await page.keyboard.press('Backspace');
     return;
@@ -49,13 +49,13 @@ export async function fillMonacoEditor(page: Page, text: string, editorLocator?:
     }, text);
     // textarea.select() steals focus; restore Monaco focus and selection before paste.
     await editor.click({ force: true });
-    await page.keyboard.press('Control+a');
-    await page.keyboard.press('Control+v');
+    await page.keyboard.press('ControlOrMeta+a');
+    await page.keyboard.press('ControlOrMeta+v');
   } else {
     // Chromium/Firefox: Use navigator.clipboard to paste without per-keystroke events.
     await page.evaluate(async (value: string) => {
       await navigator.clipboard.writeText(value);
     }, text);
-    await page.keyboard.press('Control+v');
+    await page.keyboard.press('ControlOrMeta+v');
   }
 }
--- a/playwright/commands/fillMonacoEditor.ts
+++ b/playwright/commands/fillMonacoEditor.ts
@@ -5,8 +5,19 @@ import { Locator, Page } from '@playwright/test';
  *
  * Monaco 0.56+ uses a `native-edit-context` element that is NOT a standard
  * `<textarea>` or `[contenteditable]`. Playwright's `.fill()` only works on
- * `<input>`, `<textarea>`, or `[contenteditable]` elements, so we use the
- * keyboard to select-all and type instead.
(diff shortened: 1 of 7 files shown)


playwright/commands/fillMonacoEditor.ts
```diff
@@ -17,5 +28,12 @@ export async function fillMonacoEditor(page: Page, text: string, editorLocator?:
...
   await page.keyboard.press('Control+a');
-  await page.keyboard.type(text);
+  if (text === '') {
+    await page.keyboard.press('Backspace');
+    return;
+  }
+  await page.evaluate(async (value: string) => {
+    await navigator.clipboard.writeText(value);
+  }, text);
+  await page.keyboard.press('Control+v');
```
acruzgon: **Minor (Functionality):** `Control+v` is not paste on macOS.

Playwright 1.57 supports `ControlOrMeta`, which resolves to Control on Linux/Windows and Meta on macOS. CI Linux is fine; local Mac runs of every Monaco fill will not insert text (tests fail loudly, not silently in production).

```suggestion
  await page.keyboard.press('ControlOrMeta+v');
```

Optionally also change the pre-existing `Control+a` to `ControlOrMeta+a` so select-all/empty-string works on Mac too.

playwright/playwright.config.ts
```diff
@@ -63,13 +63,16 @@ const config: PlaywrightTestConfig = {
     trace: 'retain-on-failure',
     screenshot: 'only-on-failure',
     video: 'retain-on-failure',
+    // fillMonacoEditor pastes via the clipboard (Ctrl+V). Chromium honors these
+    // permissions; other browsers ignore unknown entries.
+    permissions: ['clipboard-read', 'clipboard-write'],
```
acruzgon: **Nit (Security / least privilege):** `fillMonacoEditor` only needs `clipboard-write`. `clipboard-read` is unused by this helper.

The only `readText()` caller is the existing Hub copy-CLI test on live Chromium, which already had these permissions.

Consider:

```suggestion
    permissions: ['clipboard-write'],
```

…and keep `clipboard-read` on the `live chromium` project (or `test.use` on the copy-CLI spec). Not a merge blocker.

acruzgon (commented):
## Code Review: paste into Monaco via clipboard

**Score**: 9.8/10 (Functionality 9.5, Security 10, Quality 10)

No critical or major findings. Clipboard paste is the right root-cause fix for Monaco JSON auto-closing (`keyboard.type()`). See inline comments:

1. **Minor** — use `ControlOrMeta+v` so local macOS paste works (CI Linux is already fine).
2. **Nit** — grant only `clipboard-write` globally; keep `clipboard-read` on live Chromium.

Optional: Risk Analysis Low is defensible (test-only) but the shared helper + global permissions can also be read as Medium.

playwright/commands/fillMonacoEditor.ts
lgalis: Done. Paste and select-all now use `ControlOrMeta` so they work on macOS as well as Linux/Windows CI.

playwright/playwright.config.ts
lgalis: Done. Global grant is `clipboard-write` only. `live chromium` still has `clipboard-read` + `clipboard-write` for the Hub copy-CLI `readText()` specs.

acruzgon (approved):

PR #3540, opened 2026-09-03
Replace uuid dependency with native function

keithjgrant:
## Summary

Removes `uuid` dependency by using native uuid generating function. I'm doing this to render the dependabot PR #3427 unnecessary.

## Type of Change

- [ ] Bug fix
- [x] Enhancement
- [ ] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED


frontend/awx/analytics/components/Chart/convertApi.ts
```diff
@@ -12,7 +11,7 @@ export const convertGroupedByData = (data: GroupedApi): ChartDataSerie[] => {
         items[idx] = {
           serie: [],
           hidden: false,
-          name: v4(),
+          name: crypto.randomUUID(),
```
nixocio, line 14: Nice simplification. One compatibility concern before removing `uuid`: [`crypto.randomUUID()` is only available in secure contexts](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID). We have seen this cause a runtime failure in another UI when it was served over plain HTTP; this chart conversion would then throw.

Could we use a small shared UUID helper that falls back to [`crypto.getRandomValues()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) when `randomUUID()` is unavailable, and add a test for that fallback? That keeps the dependency removal while preserving HTTP compatibility.

lgalis, line 14: **Medium — `crypto.randomUUID()` fails on non-secure contexts (HTTP)**

[`crypto.randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) is only available in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API#security) (HTTPS or `localhost`). On plain HTTP — common in lab/dev AWX deployments — this throws `TypeError: crypto.randomUUID is not a function`.

That matters here because `convertApiToData()` runs inside a `useEffect` in the analytics `Chart` component, so a throw would break analytics chart rendering entirely. The old `uuid` package worked regardless of transport.

**Suggestion:** Add a small shared helper in `frontend/common/utils/` that prefers `crypto.randomUUID()` and falls back to an RFC 4122 v4 implementation via [`crypto.getRandomValues()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues), which works on HTTP.

frontend/awx/analytics/components/Chart/convertApi.ts
```diff
@@ -45,7 +44,7 @@ export const convertApiToData = (result: ApiReturnType): ChartData => {
         {
           serie: result.items || result.meta?.legend || [],
           hidden: false,
-          name: v4(),
+          name: crypto.randomUUID(),
```
lgalis, line 47: **Low — Inconsistent pattern elsewhere in the repo**

`playwright/commands/createE2EName.ts` also calls `crypto.randomUUID()` directly. E2E runs on `localhost` (secure context), so it is not affected today, but if you add a shared helper for the HTTP fallback (see comment on line 14), consider using it here too for consistency.

Optional follow-up — not a blocker on its own.

frontend/awx/analytics/components/Chart/convertApi.ts
lgalis, line 1: **Low — No unit tests for changed or fallback behavior**

`convertApi.ts` has no tests today, and SonarCloud reports 0% coverage on new code. For a utility-path change with environment-sensitive behavior, tests would add meaningful confidence:

- `convertGroupedByData` / `convertApiToData` produce series with unique `name` values
- Fallback path is used when `crypto.randomUUID` is undefined (mock/stub)
- Legend `childName` still maps correctly to series `name`

Consider adding `convertApi.test.ts` alongside this file.

lgalis (changes requested):
Thanks for the cleanup — removing a direct dependency with a single call site is a nice win.

I'm requesting changes primarily for the secure-context compatibility concern on `crypto.randomUUID()`. Please see inline comments for details on the HTTP fallback helper, test coverage, and optional shared-helper consistency.

keithjgrant:
Closing for later rework

PR #3511, opened 2026-08-27
[AAP-84126] fix(deps): drop unused leftover packages

nixocio:
## Summary

- Remove packages that are not imported, the fully migrated Cypress stack, obsolete Cypress workflows, and root webpack tooling.
- **Out of scope:** replacements for still-used `debounce` / `p-debounce`, `p-limit`, `uuid`, and `yaml`. Those land in a follow-up with local helpers.
- Hub Insights webpack (`frontend/hub/insights`) is unchanged.

## Type of Change

- [ ] Bug fix
- [x] Enhancement
- [ ] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

- [ ] **High** — Broad platform impact (e.g., SWR config, shared framework components, authentication, routing, API wrappers, build/bundler config). Changes affect multiple workspaces or could cause widespread regressions.
- [ ] **Medium** — Scoped but cross-cutting (e.g., shared utility functions, changes to multiple pages within one workspace, component library updates, test infrastructure). Limited blast radius but touches common code paths.
- [x] **Low** — Narrowly scoped (e.g., single page fix, styling tweak, documentation, test-only changes, copy/string updates). Minimal risk of unintended side effects.

No runtime UI behavior change. App `start`/`build` remain Vite. This only drops unused direct dependencies and retired toolchains.

| Area | Risk | Why | Residual / mitigation |
| --- | --- | --- | --- |
| Unused runtime leftovers (`dagre`, `set-value`, `klona`, `sift`, `canvas`, `marked`, `path-to-regexp`, `victory`, `chalk`) | **Low** | Zero app imports. Topology uses `@dagrejs/dagre` via PatternFly, not old `dagre` 0.8.5. Mesh viz uses `d3`. | `marked` / `yaml` / `victory` / `path-to-regexp` can still appear as **transitives** (Monaco, chart-builder, MSW). Direct CVE pins go away; parent advisories remain. |
| Cypress stack | **Low** | 0 `.cy.ts(x)` specs and no Cypress config. Unit tests are Vitest and E2E tests are Playwright. Removed the obsolete manual E2E/component workflows and Cypress cache paths. | PR checks run eslint / tsc / prettier / build / Vitest; use `/run-playwright` for ephemeral E2E. |
[...]

PR #3544, opened 2026-09-03
[AAP-84126] refactor(deps): replace debounce packages

nixocio:
## Summary

- replace the direct debounce and p-debounce dependencies with a shared debounce function and React useDebounce hook
- cancel delayed work when components unmount, and preserve trailing invocation behavior
- cover helper and hook behavior and run focused tests for every affected framework and AWX call site
- follows #3511; this draft intentionally targets its branch

## Type of Change

- [ ] Bug fix
- [x] Enhancement
- [x] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

- [ ] **High**: introduces breaking changes, significant regressions, or impacts critical user workflows.
- [ ] **Medium**: changes behavior in a way that could affect existing functionality; requires thorough testing and review.
- [x] **Low**: minor, isolated, low-impact change with minimal risk of regression.


framework/utils/debounce.ts
```diff
@@ -0,0 +1,26 @@
+export type DebouncedFunction<Args extends unknown[]> = {
+  (...args: Args): void;
+  clear: () => void;
+};
+
+export function debounce<Args extends unknown[]>(
```
nicklemmon: Some JSDoc descriptions would be useful here.

framework/hooks/useDebounce.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useMemo, useRef } from 'react';
+import { DebouncedFunction, debounce } from '../utils/debounce';
+
+export function useDebounce<Args extends unknown[]>(
```
nicklemmon: JSDoc style descriptions would be useful to keep this documented for later use.

nixocio:
Retargeted onto `devel` so Konflux can resolve an AAP version. The previous base (`aap-dep-minimize`) is not in `branch-aap-version-mapping`, which aborted E2E at `resolve-aap-version`. This PR no longer depends on #3511.

nicklemmon (approved):

PR #3549, opened 2026-09-08
Fix subscription config refresh race

nixocio:
## Summary
- wait for subscription config revalidation before completing a manifest upload
- expose the SWR refresh promise from the config context
- add regression coverage for refresh ordering

## Testing
- targeted Vitest tests
- ESLint
- Prettier
- AWX package TypeScript check

## Risk Analysis - REQUIRED

- [x] **Low** — Narrowly scoped subscription refresh fix

## Dependencies

None.

This change contains no customer, manifest, credential, escalation, or internal-ticket details.

### DIFF
--- a/frontend/awx/common/useAwxConfig.test.tsx
+++ b/frontend/awx/common/useAwxConfig.test.tsx
@@ -176,4 +176,35 @@ describe('useAwxConfig', () => {
       expect(result.current.serviceDownStatusCode).toBe(504);
     });
   });
+
+  test('refreshAwxConfig resolves after loading the latest config', async () => {
+    let requestCount = 0;
+    const refreshedConfig = {
+      ...mockConfig,
+      license_info: {
+        ...mockConfig.license_info,
+        time_remaining: 2000000,
+      },
+    };
+
+    server.use(
+      http.get(awxAPI`/config/`, () => {
+        requestCount += 1;
+        return HttpResponse.json(requestCount === 1 ? mockConfig : refreshedConfig);
+      })
+    );
+
+    const { result } = renderHook(() => useAwxConfigState(), { wrapper });
+
+    await waitFor(() => {
+      expect(result.current.awxConfig?.license_info.time_remaining).toBe(1000000);
+    });
+
+    await result.current.refreshAwxConfig?.();
+
+    await waitFor(() => {
+      expect(result.current.awxConfig?.license_info.time_remaining).toBe(2000000);
+    });
+    expect(requestCount).toBe(2);
+  });
 });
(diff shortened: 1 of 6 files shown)


lgalis (dismissed):
Non-blocking:

Findings (by severity)
Medium — Config refresh failure now blocks success after a successful upload

SubscriptionWizard.tsx
Ln 91–95
}
await refreshAwxConfig?.();
props.onSuccess();
Previously, refreshAwxConfig?.() was fire-and-forget, so onSuccess ran even if revalidation failed. Now a failed GET /config/ after a successful POST /config/ leaves the user on the wizard with a submit error and no navigation.

That is stricter and likely intentional, but it is a behavior change worth being aware of:

The manifest may already be applied on the backend while the UI still shows failure.
Retrying “Finish” may re-post the manifest.
PageWizard will surface the error (good), but consider whether the error message should distinguish “upload succeeded, config refresh failed” from a true upload failure. Not necessarily blocking if this tradeoff is accepted.

Low — Redundant second refresh in PlatformSubscription

PlatformSubscription.tsx
Ln 42–42
<SubscriptionWizard onSuccess={() => void refreshAwxConfig?.()} />
SubscriptionWizard now awaits refreshAwxConfig before calling onSuccess, so this callback triggers a second, unnecessary revalidation. Pre-existing pattern, but more obviously redundant after this PR.

Suggestion: Change onSuccess to a no-op or remove the refresh (e.g. onSuccess={() => {}}) since the wizard already handles it. The settings-route usage (navigate('/settings/subscription')) is the pattern that actually needs onSuccess.

Low — Refresh ordering only asserted for manifest flow
The manifest test correctly verifies:

expect(events).toEqual(['refresh-start', 'refresh-complete', 'success']);
The service-account test still only checks mockOnSuccess.toHaveBeenCalled(). Since await refreshAwxConfig?.() applies to all subscription types, the same ordering assertion would make the regression coverage symmetric.

Low — PlatformSubscription.test.tsx does not exercise the race fix
The new tests mock SubscriptionWizard as a button that calls onSuccess directly. They verify wiring, not the await-before-success behavior. That is covered in SubscriptionWizard.test.tsx, so acceptable, but the PlatformSubscription “wizard succeeds” test name overstates what it validates.

nixocio:
Addressed Laura’s review suggestions in commit 67cffcfdd and synced the same commit to both backports. Removed the redundant PlatformSubscription refresh callback, added refresh-before-success ordering coverage for service-account subscriptions, and kept the awaited refresh behavior so the UI does not navigate past a failed config revalidation after upload. All 25 targeted tests pass.

nixocio:
Follow-up: added a specific error when the backend upload succeeds but the subsequent subscription config refresh fails: “Subscription uploaded, but the subscription status could not be refreshed. Please try again.” This is also included in both 2.6 and 2.7 backports.

nixocio:
Sonar coverage follow-up is now fixed in commit 1a667eb8c. Added regression coverage for the manifest-upload-success/config-refresh-failure path; the targeted suite now passes 26 tests. The same test is included in both backports.

lgalis (approved):
LGTM

PR #3552, opened 2026-09-09
Feature/conditional survey

kenancakir:
# feat(ui): conditional surveys with categories, multi-page wizard, and a condition editor

## Summary

This PR adds **conditional (dynamic) surveys** to AWX Job Template and Workflow Job
Template launch flows. Survey questions can now be shown or hidden based on the
answers to *other* questions, grouped into **categories**, and split across
**multiple wizard pages**. The survey editor gets a UI to author these conditions
without touching raw JSON.

It is a direct, concrete take on the long-standing request in
[ansible/awx#20 — Programmable Surveys](https://github.com/ansible/awx/issues/20),
which explicitly asks for *conditional survey questions that depend on previous
answers*.

> **Transparency / provenance:** This feature was designed and implemented with the
> help of AI (Claude Code). It was built as a **self-contained, standalone
> solution** on top of the existing survey code. I'm opening it as a genuine
> contribution — but I'm equally happy for it to be treated as a **proof of concept
> or an idea to build on**. Feel free to merge it, cherry-pick parts of it, or use it
> purely as a reference for a "proper" implementation. No hard feelings whichever way
> it goes. I wanted to make the AI involvement explicit rather than hide it.

## Motivation

Today surveys are a flat list: every question is always visible and there's no way
to react to previous answers. In practice, launch forms often need branching
("only ask for the DB password if the user chose *external database*"), logical
grouping, and a less overwhelming layout for large surveys. Issue [#20](https://github.com/ansible/awx/issues/20) has tracked
this need for a long time. This PR is an attempt to cover the "conditional
questions" part of that request end-to-end in the new React/PatternFly UI.

## What it does

**Condition engine**
- New `evaluateConditions()` / `pruneSurveyByConditions()` helpers in
  `frontend/awx/common/useSurveyConditions.ts`.
- Operators: `eq`, `neq`, `in`, `notin`, `gt`, `lt`, `gte`, `lte`, `is_set`,
  `is_not_set`.
- Multiple conditions per question combined with **AND / OR** logic
  (`condition_logic`).
- Conditions are evaluated against the full answer set (survey answers merged over
  extra vars context), so visibility reacts live as the user types.

**Survey editor (`TemplateSurveyForm.tsx`)**
- A **condition editor** UI: add/remove condition rows, pick the driving variable
  (from other survey questions), pick an operator, and enter the expected value.
- Value input is hidden automatically for `is_set` / `is_not_set`.
- AND/OR selector appears once there is more than one condition.
- New per-question fields: **Category** and **Page**.

**Launch wizard (`TemplateLaunchWizard.tsx` / `SurveyStep.tsx`)**
- Questions with a `page` number are rendered as **separate wizard steps**;
  a single-page survey keeps the old single "Survey" step.
- Questions are grouped by **category** into titled PatternFly cards.
- Hidden questions are filtered out of the rendered form in real time.
- Page-aware value resolution: each variable is read from the wizard step that
  "owns" it, avoiding stale duplicate values leaking across pages.

**Review step & submission correctness**
- `pruneSurveyByConditions()` is applied before the **review step** and before the
  **launch payload** is built, so conditionally-hidden fields (which still carry a
  seeded default) are **not** submitted as extra vars and don't show up in review.

**Data model (`interfaces/Survey.ts`)**
- `Spec` extended with optional `category`, `page`, `conditions`, `condition_logic`.
- New `SurveyCondition` interface. All fields are optional, so existing surveys keep
  working unchanged.

## Scope / affected files

- `frontend/awx/common/useSurveyConditions.ts` (new)
- `frontend/awx/common/SurveyStep.tsx`
- `frontend/awx/interfaces/Survey.ts`
- `frontend/awx/resources/templates/TemplatePage/TemplateLaunchWizard.tsx`
- `frontend/awx/resources/templates/TemplatePage/TemplateSurveyForm.tsx`
- `frontend/awx/resources/templates/TemplatePage/steps/TemplateLaunchReviewStep.tsx`

## What it does NOT do (limitations / out of scope)

- **Frontend only, by design.** This PR ships only the UI. The condition metadata is
  stored inside the existing `survey_spec` structure, so on its own there is **no
  server-side enforcement** — a caller hitting the launch API directly could still
  submit a hidden field's value. This is UI-level branching, not a security boundary.
  - **Note:** I have also prototyped a **matching backend validator in AWX** that
    enforces these conditions server-side, but I deliberately left it out of this PR
    to keep the change small and self-contained and to get a read on the frontend
    direction first. If there's interest, I'm happy to open a **separate follow-up PR
    against `ansible/awx`** with the backend piece so the two sides line up.
- **No automated tests included yet.** No new Vitest/Playwright coverage was added
  for the condition engine, the editor UI, or the multi-page wizard. This is the
  biggest gap before a real merge.
- **No dynamic / API-populated fields.** Issue [#20](https://github.com/ansible/awx/issues/20) also asks for questions whose
  choices are fetched at runtime (e.g. reading options from an API / external data
  source, or populating a dropdown from live inventory, projects, credentials, etc.).
  This PR only branches on **statically authored** answers — it does not pull choice
  lists from any endpoint.
- **No computed / derived fields.** There's no way to auto-fill or calculate a
  question's value from other answers or from an external lookup — the user still
  enters every visible field manually.
- **No custom validation rules** beyond the existing survey field types (no
  regex/range/cross-field validation driven by the survey spec).

## Backwards compatibility

Surveys without the new fields behave exactly as before: no categories → no cards,
single page → single survey step, no conditions → always visible. The new `Spec`
fields are all optional.

## How to test manually

1. Open a Job Template with surveys enabled → **Survey** tab → add/edit a question.
2. Set a **Category** and a **Page**, then add one or more **Conditions** referencing
   another question (try AND vs OR, and `is_set` / `eq` / `in`).
3. Launch the template: verify questions appear/disappear as you change answers,
   that pages become separate wizard steps, and that categories render as cards.
4. On the **Review** step and in the launched job's **extra vars**, confirm that
   hidden questions are absent.

## Notes for reviewers

I'd especially appreciate feedback on: whether this belongs in the frontend at all
vs. a backend-backed survey spec, the shape of the `SurveyCondition` schema, and how
you'd want the multi-page/category concepts modeled if this were to be done
"officially". Again — happy for this to be merged, adapted, or just used as input for
[issue #20.](https://github.com/ansible/awx/issues/20)

### DIFF
--- a/frontend/awx/common/SurveyStep.tsx
+++ b/frontend/awx/common/SurveyStep.tsx
@@ -5,11 +5,13 @@ import { PageSelectOption } from '@ansible/ansible-ui-framework/PageInputs/PageS
 import { usePageWizard } from '@ansible/ansible-ui-framework/PageWizard/PageWizardProvider';
 import { useGet } from '@ansible/common-ui/crud/useGet';
 import { useEffect } from 'react';
-import { useFormContext } from 'react-hook-form';
+import { useFormContext, useWatch } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
+import { Card, CardBody, CardTitle } from '@patternfly/react-core';
 import { Spec, Survey } from '../interfaces/Survey';
 import { WizardFormValues } from '../resources/templates/WorkflowVisualizer/types';
 import { awxAPI } from './api/awx-utils';
+import { evaluateConditions } from './useSurveyConditions';
 
 function getJobType(resource: WizardFormValues['resource']) {
   if (!resource) return;
@@ -35,10 +37,14 @@ export function SurveyStep({
   templateId,
   jobType,
   singleColumn,
+  pageNumber,
+  surveySpecData,
 }: {
   templateId?: string;
   jobType?: string;
   singleColumn?: boolean;
+  pageNumber?: number;
+  surveySpecData?: Survey;
 }) {
   const { t } = useTranslation();
   const { wizardData, stepData } = usePageWizard();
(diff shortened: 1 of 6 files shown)


adpavlov:
I’m configuring surveys via survey spec json. Is it covered?

kenancakir:
> I’m configuring surveys via survey spec json. Is it covered?

Yes, that's covered. Since the API essentially just transmits plain JSON and the UI handles the rendering, you can also build the survey using the API. Essentially, you just need to add a few more fields: 

For example: https://github.com/kenancakir/awx/blob/feature/conditional-survey/vm_provisioning_survey.json

svenfroebus:
Hi kenancakir: This feature is brilliant and belongs in AAP. For some workflow job templates, the "standard surveys" simply aren't enough—or they are too long. In my opinion, Red Hat should integrate this as soon as possible. Thanks and best regards, Sven

PR #3561, opened 2026-09-10
[AAP-92380] Stabilize workflow visualizer save assertion

nixocio:
## Summary

`saveAndDismissToast` in the workflow visualizer template-switch spec waited only on the success toast. The toolbar creates that toast with a 5s auto-dismiss timeout, while saving issues a long sequence of node and credential requests, so on a live backend the outcome regularly landed outside the old 30s budget and the transient toast could come and go unobserved.

The helper also clicked Save by role. The toolbar renders two save controls ("Save" and "Save and exit"), and the toolbar button is disabled until the visualizer registers the pending edit, so the click could land on a dead button.

This change:

- targets the toolbar button by test id instead of by accessible name
- waits for it to become enabled rather than clicking a disabled control
- races the success toast against the failure alert, and surfaces the alert text

That last point matters for diagnosis: waiting only on the toast reported a failed save as an opaque `element(s) not found`, hiding the server error that explained it.

Test-only change; no product code is touched.

## Type of Change

- [x] Tests

## Risk Analysis - REQUIRED

### Manual Testing

Ran the full spec against a live AAP backend. 10 of 12 scenarios pass, including every scenario that exercises the save helper.

The two remaining failures (Scenarios 11 and 12) are unrelated to this change and do not reach the save path — both fail in `beforeEach`, in the shared `login()` command, on its 5s wait for the toolbar user button after 11+ consecutive logins in a serial run. That is a separate pre-existing issue in a shared command and is out of scope here.

### DIFF
--- a/playwright/tests/integration/automation-execution/workflow-visualizer/workflow-visualizer-template-switch.spec.ts
+++ b/playwright/tests/integration/automation-execution/workflow-visualizer/workflow-visualizer-template-switch.spec.ts
@@ -62,8 +62,9 @@ async function saveAndDismissToast(page: import('@playwright/test').Page) {
   // found" and hides the server error that explains it. Keep the budget under
   // the per-test timeout: this helper runs several times per test, so a larger
   // value would just trade the reported error for a generic test timeout.
-  const toast = page.getByText('Successfully saved workflow visualizer');
-  const errorAlert = page.getByText('Failed to save workflow job template');
+  const toaster = page.getByTestId('alert-toaster');
+  const toast = toaster.getByText('Successfully saved workflow visualizer');
+  const errorAlert = toaster.getByText('Failed to save workflow job template');
   await expect(toast.or(errorAlert).first()).toBeVisible({ timeout: 60000 });
 
   if (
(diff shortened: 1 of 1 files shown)


acruzgon:
## Code Review: [AAP-92380] Stabilize workflow visualizer save assertion

**Score**: 9.8/10 (Functionality 10, Security 10, Quality 9.5)

Reviewed via AAP SDLC (`review-pr-workflow` → `code-review`).

### Summary

No critical or major findings. This is a well-scoped, test-only fix that addresses real flake root causes in the shared `saveAndDismissToast` helper (called 24× in this spec):

- **Selector ambiguity** — replaced role-based "Save" with `getByTestId('workflow-visualizer-toolbar-save')` to avoid matching "Save and exit"
- **Disabled-button click** — waits for the toolbar Save to become enabled before clicking
- **Toast race / opaque failures** — races success vs. failure alerts inside `alert-toaster`, surfaces server error text on failure, and scopes assertions to avoid strict-mode collisions

The changes align with [AAP-92380](https://redhat.atlassian.net/browse/AAP-92380) and match product strings/test ids in `WorkflowVisualizerToolbar.tsx` and `PageAlertToaster.tsx`. CI is green; `e2e: done` label present.

### Minor (non-blocking)

- **Follow-up opportunity**: sibling specs (`workflow-visualizer.spec.ts`, `workflowVisualizerOutput.spec.ts`) still use the older unscoped toast pattern — consider applying the same helper in a future PR if they flake similarly.

### Recommendation

Approve from a code-review perspective. Nice iterative fixes across the three commits (test-id scoping, strict-mode guard, toaster containment).

---
*AI-assisted review — human maintainer approval still required.*

acruzgon (approved):

PR #3569, opened 2026-09-11
Enable unused local checks in TypeScript

nixocio:
## Summary

Enable TypeScript unused-local checking in both project configurations and remove the violations found by the new check.

The CommonJS configuration also now resolves the Monaco editor types, supports the existing `String.replaceAll` usage, and excludes Vite-only configuration from the publish build.

## Validation

- Root TypeScript check
- CommonJS TypeScript check
- Changed-file ESLint
- Prettier
- `git diff --check`

## TypeScript references

- [`noUnusedLocals`](https://www.typescriptlang.org/tsconfig/noUnusedLocals.html)

## Risk Analysis - REQUIRED


nixocio:
Addressed the actionable Chai Bot suggestions:

- Documented that `tsconfig.cjs.json` uses `ES2021` for type-library compatibility while retaining `target: ES5` for CJS output.
- Kept `noUnusedParameters: false` intentionally because interface and callback signatures can require parameters that are not used; this is now documented in both TypeScript configs (commit `337654814`).
- Confirmed the `ChatbotPortal.tsx` eslint-disable is pre-existing and outside this PR.

nixocio:
Thanks for the review. The remaining points are non-blocking and were verified:

- The ES2021 CJS lib change is scoped to the CJS/framework build and the current typecheck passes.
- `noUnusedParameters: false` remains intentional because interface and callback signatures can require parameters that are not used locally; this can be revisited separately.
- The `i18next/no-literal-string` suppression in ChatbotPortal.tsx is pre-existing and outside this PR.

No additional changes are needed for these suggestions.

tsconfig.cjs.json
```diff
@@ -2,16 +2,21 @@
   "compilerOptions": {
     "target": "ES5",
     "module": "CommonJS",
-    "lib": ["ES5", "DOM"],
+    "lib": ["ES2021", "DOM"],
```
nicklemmon, line 5: Why ES2021? Agreed that ES5 is way too old.

nicklemmon (approved):

tsconfig.cjs.json
nixocio: Good call — the old `"lib": ["ES5", "DOM"]` was too restrictive for type-checking.

This PR only updates which standard-library *types* are available during compile; it does not change emitted output. TypeScript separates `lib` and `target` ([handbook](https://www.typescriptlang.org/tsconfig#lib)):

- **`target: "ES5"`** stays. This config emits the published framework CJS bundle to `./publish/cjs`, and we are not changing runtime/emit behavior here.
- **`lib: ["ES2021", "DOM"]`** aligns framework CJS type-checking with the root `tsconfig.json`, so we get modern stdlib types without emitting ES2021 syntax — emit is still governed by `target: "ES5"`.

The main scope of this PR is enabling `noUnusedLocals` (with `noUnusedParameters` left off per our repo convention — see the inline note in this file). Bumping `lib` lets that check run against the same type surface as the rest of the repo.

PR #3570, opened 2026-09-11
Enable unused parameter checks in TypeScript

nixocio:
## Summary

Enable TypeScript unused-parameter checking in both project configurations and make the existing intentional unused parameters explicit.

The changes preserve callback signatures and remove one unused mock prop; they do not change application behavior.

## Validation

- Root TypeScript check
- CommonJS TypeScript check
- Changed-file ESLint
- Prettier
- `git diff --check`

## TypeScript references

- [`noUnusedParameters`](https://www.typescriptlang.org/tsconfig/noUnusedParameters.html)

## Risk Analysis - REQUIRED

- [ ] **High** — Broad platform impact (e.g., auth, RBAC, shared framework, routing, API client contracts).
- [ ] **Medium** — Cross-cutting but reversible (e.g., shared component behavior, form validation patterns).


h-kataria (approved):

nicklemmon (approved):

nicklemmon:
AI-assisted review comment:

`playwright-insights/tsconfig.json` is the only standalone TypeScript config that neither extends the root `tsconfig.json` nor enables `noUnusedLocals`/`noUnusedParameters`. Its expanded config leaves both options unset, so the new unused-parameter policy does not apply when this package is checked with its own config. Please either extend the root config (while preserving its standalone module/bundler settings as needed) or set both flags explicitly.

nixocio:
Addressed in 089298b5c — `playwright-insights/tsconfig.json` now extends the root config so `noUnusedLocals`/`noUnusedParameters` apply while keeping the bundler/module settings.

PR #3577, opened 2026-09-14
fix(survey): render question names containing colons

nixocio:
## Summary
Survey question names are user-provided values and must not be passed through i18next. A colon in a question name was interpreted as an i18next namespace separator, causing the label to render blank. This renders the question name directly and adds regression coverage.

## Risk Analysis - REQUIRED

- [ ] **High** — Broad platform impact.
- [ ] **Medium** — Scoped but cross-cutting.
- [x] **Low** — Narrowly scoped survey launch form label fix with minimal risk of unintended side effects.

## Dependencies

None.

## Validation
- Vitest: frontend/awx/common/SurveyStep.test.tsx
- ESLint on changed files
- Prettier check on changed files


frontend/awx/common/SurveyStep.test.tsx
```diff
@@ -63,6 +63,24 @@ const testSurveys = {
       },
     ],
   },
+  withColonInQuestionName: {
+    name: 'Colon in question name',
+    description: '',
+    spec: [
+      {
+        question_name: 'Markets:',
```
lgalis, line 72: Optional: `'Markets:'` covers the blank-label case (empty key after the separator). A mid-string colon (`'Markets: US'`) is a different i18next split (namespace + leftover key → truncated text, not empty). One extra name in this fixture would cover that, without needing another test.

Covering only `type: 'text'` is enough **if** the test above actually fails when `t(question_name)` returns. Right now it would not, and the same `label={t(...)}` pattern is repeated for all seven field types in `SurveyStep.tsx`.

frontend/awx/common/SurveyStep.test.tsx
```diff
@@ -153,6 +171,14 @@ describe('SurveyStep', () => {
     });
   });
 
+  test('renders a survey question name containing a colon', async () => {
+    renderSurveyStep(server, '999', testSurveys.withColonInQuestionName);
+
+    await waitFor(() => {
+      expect(screen.getByText('Markets:')).toBeInTheDocument();
```
lgalis: **Blocking:** this assertion does not lock the colon/i18next regression.

AWX Vitest mocks `t` as identity (`framework/vitest.i18n.ts`), so `t('Markets:')` still returns `'Markets:'`. This test would pass on `devel` with `label={t(element.question_name)}` still in place, and CI would stay green on a revert of the actual bug.

Please make the test fail if `t(question_name)` is reintroduced. Either:
1. Spy the shared `t` (`useTranslation()` uses the same `vi.fn`) and `expect(t).not.toHaveBeenCalledWith('Markets:')`, or
2. In this test only, stub `t` to mimic production `nsSeparator: ':'` so a colon name would render empty/truncated if wrapped again.

Also prefer `getByRole('textbox', { name: 'Markets:' })` over `getByText` so this asserts the accessible field label, not leftover copy.

lgalis:
## Rest of the review (not tied to a diff line)

**Product change looks good.** Dropping `t()` on operator-authored names is simpler and safer than a global `nsSeparator: false`. That global option would still send survey names through i18n (interpolation with `escapeValue: false`, pseudolocalization). `question_description` and choice strings were already rendered raw. The survey editor list UI (`useSurveyColumns`, `useManageSurveyQuestions`) was already raw too.

**No functional/security issue** in `SurveyStep.tsx`. `PageFormTextInput` forwards `label` without wrapping it in `t()` again.

**Out of scope / no change needed**
- E2E is not required for this size of change if the unit test actually fails on revert.
- CI on `ddb593d` is green (prettier, eslint, tsc, Vitest, Sonar).

**Suggested PR explanation** (optional to paste into the description):

> Launch-form survey question labels are rendered as the stored `question_name` instead of `t(question_name)`. Question names are user data; i18next treats `:` as a namespace separator, so a name like `Markets:` can resolve to an empty key and render blank. Static UI copy (`Select option`) still goes through `t()`.

lgalis (changes requested):
Requesting changes for the new test, not the product fix.

`label={element.question_name}` is the right change: survey names are user data, and runtime i18next still treats `:` as a namespace separator. Placeholders correctly stay behind `t()`.

The added Vitest case does not actually protect that fix — line comments on `SurveyStep.test.tsx` have the details. Please make that test fail if `t(element.question_name)` is put back before merge.

lgalis (dismissed):
Non-blocking
useTranslation() inside a test() callback (SurveyStep.test.tsx ~188). It works only because the setup mock is a vi.fn, not a real hook. Prefer vi.mocked(useTranslation)() so the test does not look like a component.

mockClear() on the process-wide stableT. Safe here (no concurrent, AWX common Vitest passed). Slightly less brittle: assert on t.mock.calls without clearing shared mock state.

Text-only coverage. Restoring t(question_name) on integer / float / password / textarea / multiplechoice / multiselect would not fail this test. Acceptable if you treat the seven branches as one pattern; a helper would make a one-branch revert harder. Not required.

frontend/awx/common/SurveyStep.test.tsx
nixocio: Addressed in d86c5e3d / 5155b0a1: added mid-string `Markets: US`, `getByRole('textbox', { name })` assertions, and `t` spy `not.toHaveBeenCalledWith` for both names.

frontend/awx/common/SurveyStep.test.tsx
nixocio: Addressed in d86c5e3d / 5155b0a1: spy the shared `t` mock and assert `not.toHaveBeenCalledWith('Markets:')` / `('Markets: US')`, plus accessible name via `getByRole('textbox', { name })`.

lgalis (dismissed):

lgalis (approved):
LGTM

PR #3583, opened 2026-09-14
chore(node): move CI and engines to Node 24 LTS

nixocio:
## Summary
- Node 20 reached EOL on 2026-03-24 (https://nodejs.org/en/about/previous-releases, https://nodejs.org/en/about/eol).
- Pin all GitHub Actions `setup-node` jobs to Node 24 (current LTS).
- Tighten `package.json` `engines.node` to `>=24.0.0` and add `.nvmrc`.
- Update Hub Insights setup docs that still said Node 20.x+.

## Risk Analysis - REQUIRED

- [x] **Low** — Toolchain/CI pin and engines metadata only; no product runtime behavior change.

## Test plan
- [ ] CI green on Node 24 (vitest, lint, build workflows)
- [ ] Confirm no workflow still references `node-version: 20`
- [ ] Local `nvm use` / `.nvmrc` resolves to 24


nixocio:
Addressed the review suggestions with verification:\n\n- The root `package.json` already requires `node >=24.0.0`.\n- Audited the repository for Node-20 Dockerfile pins; none are present.\n- Ran `npm test` under Node 24.21.0. Common tests passed (35 files / 301 tests), but the full run is not clean because EDA has existing localhost:3000/network-dependent timeouts (39 tests failed) and the framework TypeScript check hit a compiler stack overflow. These failures are unrelated to the Node-version configuration changes.\n\nNo additional source change is needed for these suggestions.

h-kataria (approved):

PR #3589, opened 2026-09-15
fix(automation-dashboard): exclude system jobs from report data (AAP-92027)

appuk:
## **Bug:** AAP-92027

The automation dashboard was including management jobs (Cleanup Activity Stream, Cleanup Expired Sessions, Cleanup Job Details) in job counts, charts, and the template table. These `system_job` records were ingested into the metrics-service `JobData` table via the hourly sync hook, which does not filter by job type.

### What's happening

The automation dashboard is powered by two endpoints in `ansible/metrics-service`:

| Endpoint | What it powers |
|---|---|
| `GET /api/metrics/v1/dashboard_reports/report/` | Template table (rows with run counts, costs, savings) |
| `GET /api/metrics/v1/dashboard_reports/report/details/` | Summary cards, charts, top users, top projects |

Both query the `dashboard_job_data` table, which has been polluted with `system_job` records — "Cleanup Activity Stream", "Cleanup Expired Sessions", "Cleanup Job Details".

### **UI fix (the PR)**

This UI-side fix fetches all real job template IDs from the `/templates/` endpoint (which `INNER JOIN`s `main_jobtemplate`, excluding `system_job_templates`) and injects them as `template=` query params into the `/report/` and `/report/details/` API calls when the user has no template filter active. When the user selects templates from the dropdown, their selection is used as-is since the dropdown already excludes `system_job_templates`.

Since a backend fix requires migrations and a release cycle, the PR does a frontend-only fix:

1. New hook `useJobTemplateIds` fetches all real template IDs from `/templates/` (already excludes `system_job_templates`).
2. When no user template filter is active, those IDs are injected as `template=` params into both `/report/` and `/report/details/` calls.
3. When the user selects templates from the dropdown, their selection is used as-is (already clean).

*5 files changed, no backend changes, no migrations.*

## Type of Change

- [x] Bug fix
- [ ] Enhancement
- [ ] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

## Dependencies

*None*

### Ephemeral E2E Tests

### Manual Testing

- Have an instance with mix of System Jobs and Regular Jobs
- Verify that the system jobs don't show up in Templates list in Dashboard `analytics/automation-dashboard`

### Screenshots
Before: All the jobs counts were incorrect, System jobs were being shown in the list and in calculculations for savings
image

After: All the job counts are correct, System jobs are being excluded from templates list and is not in calculations for savings
image


frontend/awx/analytics/automation-dashboard/views/useAutomationDashboardView.tsx
```diff
@@ -71,7 +87,12 @@ export function useAutomationDashboardView(options: {
     [mainTableViewBase, clearAllFilters]
   );
 
-  const detailsResponse = useGetReportDetails(toolbarFilters, filterState, QUERY_PARAMS);
+  const detailsResponse = useGetReportDetails(
+    toolbarFilters,
+    filterState,
+    QUERY_PARAMS,
+    systemJobExclusionParams
```
lgalis, line 90: **Blocking:** `systemJobExclusionParams` is wired into `/report/details/`, but `useExportCsv` below (line 109) still calls `/dashboard_reports/report/export/` without the same template-ID injection.

When no template filter is active, CSV export can still include system jobs (Cleanup Activity Stream, Cleanup Expired Sessions, etc.) while the on-screen table and KPI cards are fixed.

Please extend `useExportCsv` with the same exclusion params (and add a unit test), or explicitly confirm export is out of scope for AAP-92027 before merge.

frontend/awx/analytics/automation-dashboard/common/useJobTemplateIds.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * This endpoint queries the AWX DB's main_jobtemplate table via an INNER JOIN,
+ * so system_job_templates (Cleanup Activity Stream, etc.) are excluded.
+ *
+ * Returns template IDs as URLSearchParams entries (['template', '<id>'][])
+ * suitable for injection into API query strings to exclude system jobs from
+ * report results. Returns an empty array while loading or on error.
+ */
+export function useJobTemplateIds(): [string, string][] {
+  const url = metricsAPI`/dashboard_reports/templates/?page_size=1000`;
```
lgalis: **Fix or document + Jira:** This fetches only the first page (`page_size=1000`) with no pagination. On instances with >1000 job templates, jobs for templates beyond page 1 will be excluded from dashboard metrics (under-reporting), even though system jobs remain excluded.

Please either paginate through all template pages (similar to `useAwxGetAllPages`) or document this limitation in the PR/README and file a follow-up Jira story.

frontend/awx/analytics/automation-dashboard/common/useJobTemplateIds.ts
```diff
@@ -0,0 +1,34 @@
...
+  const fetcher = useFetcher();
+  const { data } = useSWR<AwxItemsResponse<TemplateRecord>>(url, fetcher, {
+    revalidateOnFocus: false,
+    dedupingInterval: 60_000,
+  });
+
+  return useMemo(() => {
+    if (!data?.results?.length) return [];
+    return data.results.map((t): [string, string] => ['template', t.id.toString()]);
```
lgalis: **Fix or document + Jira:** Injecting hundreds/thousands of `template=<id>` params can produce query strings of 10–20 KB, which may exceed proxy/server URL limits (~8 KB) and cause 414 errors on large deployments.

Please validate against a realistic template count in staging, or document this as a known limitation of the frontend workaround and file a Jira story for the backend fix.

frontend/awx/analytics/automation-dashboard/common/useJobTemplateIds.ts
```diff
@@ -0,0 +1,34 @@
...
+export function useJobTemplateIds(): [string, string][] {
+  const url = metricsAPI`/dashboard_reports/templates/?page_size=1000`;
+  const fetcher = useFetcher();
+  const { data } = useSWR<AwxItemsResponse<TemplateRecord>>(url, fetcher, {
+    revalidateOnFocus: false,
+    dedupingInterval: 60_000,
+  });
+
+  return useMemo(() => {
+    if (!data?.results?.length) return [];
```
lgalis: **Fix or document + Jira:** Returning `[]` while SWR is loading means report/details endpoints fetch once without exclusion params, then refetch after template IDs arrive. Users may briefly see incorrect KPIs/counts including system jobs.

Consider gating report fetches until template IDs are loaded, or document this flash-of-incorrect-data behavior and file a Jira story if deferring.

frontend/awx/analytics/automation-dashboard/common/useJobTemplateIds.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Fetches all job template IDs from the /dashboard_reports/templates/ endpoint.
+ *
+ * This endpoint queries the AWX DB's main_jobtemplate table via an INNER JOIN,
+ * so system_job_templates (Cleanup Activity Stream, etc.) are excluded.
+ *
+ * Returns template IDs as URLSearchParams entries (['template', '<id>'][])
+ * suitable for injection into API query strings to exclude system jobs from
+ * report results. Returns an empty array while loading or on error.
```
lgalis: **Fix or document + Jira:** On templates API error, this silently returns `[]` and exclusion is skipped — the original bug persists with no user-visible indication.

Acceptable as a temporary workaround, but please document this fallback behavior (PR description or code comment) and file a Jira story to track observability or the planned backend fix.

frontend/awx/analytics/automation-dashboard/views/useAutomationDashboardView.tsx
```diff
@@ -42,14 +43,29 @@ export function useAutomationDashboardView(options: {
...
 
   const { filterState, setFilterState } = mainTableViewBase;
 
+  // When the user has no template filter active, inject all known job template
+  // IDs so system_job records (Cleanup Activity Stream, etc.) are excluded from
+  // both the table and summary cards. When a template filter IS active, the
+  // user selected from the clean dropdown (which already excludes system_job
+  // templates), so no injection is needed.
+  const systemJobExclusionParams = useMemo(() => {
+    const hasTemplateFilter = filterState?.template && filterState.template.length > 0;
```
lgalis: **Fix:** The `hasTemplateFilter` check here duplicates the same logic in `useAutomationDashboardBaseView` (`hasUserTemplateFilter` at line 52). If one path changes without the other, table and details could desync.

Consider extracting a shared helper, e.g. `getSystemJobExclusionParams(filterState, templateIds)`, used by both hooks.

lgalis (changes requested):
## Review summary

The dashboard fix is well structured and tested, but CSV export is still missing the same system-job exclusion logic. Several workaround limitations should be documented and tracked in Jira if not fixed in this PR.

See inline comments for details.

appuk:
Comment 1 (CSV Export): Added extraSearchParams parameter to useExportCsv and wired systemJobExclusionParams from the view,
  so CSV exports now exclude system jobs.

  Comment 2 (Pagination): Replaced useSWR with useAwxGetAllPages in useJobTemplateIds, which paginates through all pages
  instead of fetching only the first 1000.

  Comment 3 (URL Length Risk): Added a JSDoc comment in useJobTemplateIds documenting this as a known frontend-workaround
  limitation pending a backend fix.

  Comment 4 (Race Condition): useJobTemplateIds now returns undefined while loading instead of []. Added
  isLoadingSystemJobExclusionIds option to useAutomationDashboardBaseView that gates SWR fetching (fetchUrl = null) until
  template IDs are ready.

  Comment 5 (Silent Error Handling): useJobTemplateIds now exposes error from SWR so callers can detect and handle API
  failures.

  Comment 6 (Duplicated Filter Logic): Extracted getSystemJobExclusionParams() into utils/queryString.ts and replaced the
  inline logic in both useAutomationDashboardView.tsx and useAutomationDashboardBaseView.ts.

frontend/awx/analytics/automation-dashboard/views/useAutomationDashboardView.tsx
```diff
@@ -42,14 +44,24 @@ export function useAutomationDashboardView(options: {
   toolbarFilters: IToolbarFilter[];
 }): IAutomationDashboardView {
   const { toolbarFilters } = options;
+
+  const { templateIds: allTemplateIds, isLoading: isLoadingTemplateIds } = useJobTemplateIds();
```
lgalis, line 48: **Templates API failure is ignored**

`useJobTemplateIds` exposes `error`, but only `templateIds` and `isLoading` are consumed here. If `/dashboard_reports/templates/` fails, `isLoading` becomes `false`, `templateIds` stays `undefined`, exclusion is skipped, and both `/report/` and `/report/details/` proceed unfiltered — the original system-job bug returns with no user-visible indication.

Consider blocking dashboard fetches (and surfacing an error) when `error` is set, similar to the loading gate on the table.

frontend/awx/analytics/automation-dashboard/views/useAutomationDashboardView.tsx
```diff
@@ -71,7 +83,12 @@ export function useAutomationDashboardView(options: {
     [mainTableViewBase, clearAllFilters]
   );
 
-  const detailsResponse = useGetReportDetails(toolbarFilters, filterState, QUERY_PARAMS);
+  const detailsResponse = useGetReportDetails(
```
lgalis, line 86: **Details fetch is not gated while template IDs load**

The table correctly waits via `isLoadingSystemJobExclusionIds`, but `useGetReportDetails` is called immediately. While `allTemplateIds` is still `undefined`, `systemJobExclusionParams` is `undefined` and `/report/details/` fires without exclusion params, then refetches once IDs arrive.

Summary cards/charts can briefly show inflated counts (system jobs included) while the template table is still waiting. Consider passing `isLoadingTemplateIds` into `useGetReportDetails` and setting `url = null` until IDs are ready, or defer rendering KPIs until exclusion params are available.

lgalis (commented):
Two remaining items from follow-up review — see inline comments.

frontend/awx/analytics/automation-dashboard/views/useAutomationDashboardView.tsx
appuk: Added isLoadingExclusionIds parameter. When true, the SWR fetch URL is set to null, preventing the details endpoint from firing before template IDs are ready.

frontend/awx/analytics/automation-dashboard/views/useAutomationDashboardView.tsx
appuk: Destructured error from useJobTemplateIds() as templateIdsError. Passed isLoadingTemplateIds to useGetReportDetails to gate it. Exposed templateIdsError in the returned view object.

appuk:
views/useGetReportDetails.tsx — Added isLoadingExclusionIds parameter. When
  true, the SWR fetch URL is set to null, preventing the details endpoint from
  firing before template IDs are ready. (Comment 8)

  views/useAutomationDashboardView.tsx — Destructured error from
  useJobTemplateIds() as templateIdsError. Passed isLoadingTemplateIds to
  useGetReportDetails to gate it. Exposed templateIdsError in the returned view
  object. (Comments 7 & 8)

  types/index.ts — Added templateIdsError: Error | undefined to
  IAutomationDashboardView.

  AutomationDashboard.tsx — Imported EmptyStateError. When templateIdsError is
  set, renders an error state ("Unable to load dashboard data") instead of the
  dashboard content, preventing unfiltered data from showing.

updated Tests

lgalis (commented):
Must-fix items from code review (inline comments below).

frontend/awx/analytics/automation-dashboard/common/useJobTemplateIds.ts
```diff
@@ -0,0 +1,71 @@
...
+  const error: Error | undefined = response.error;
+
+  useEffect(() => {
+    if (data && data.length === size && data[data.length - 1]?.next) {
+      void setSize(size + 1);
+    }
+  }, [data, size, setSize]);
+
+  const templateIds = useMemo(() => {
+    if (isLoading || data === undefined) return undefined;
```
lgalis: **Must fix:** `isLoading` from `useSWRInfinite` goes false after page 1, while later pages are still being fetched. On instances with **>200 templates**, `templateIds` is computed from partial data and the loading gate opens too early — report/table/details fetch with an incomplete `template=` list (under-reporting) until a refetch after pagination completes.

Gate until pagination is complete, e.g. keep `templateIds` / `isLoading` true while `data?.at(-1)?.next` is set. Add a test that asserts no partial IDs are exposed mid-pagination.

frontend/awx/analytics/automation-dashboard/views/useAutomationDashboardView.tsx
```diff
@@ -42,14 +44,28 @@ export function useAutomationDashboardView(options: {
...
+    isLoading: isLoadingTemplateIds,
+    error: templateIdsError,
+  } = useJobTemplateIds();
+
   const mainTableViewBase = useAutomationDashboardBaseView<IJobTemplate>({
     url: metricsAPI`/dashboard_reports/report/`,
     defaultFilters: DEFAULT_FILTERS,
     toolbarFilters,
+    systemJobExclusionTemplateIds: allTemplateIds,
+    isLoadingSystemJobExclusionIds: isLoadingTemplateIds,
```
lgalis: **Must fix:** Report fetches are gated only on `isLoadingTemplateIds`, not on `templateIdsError`. When `/dashboard_reports/templates/` fails, `isLoading` is false, exclusion is skipped, and `/report/` + `/report/details/` still fire unfiltered — the original system-job bug returns (UI error state hides it, but fetches proceed and export is unaffected).

Also, if pagination fails after page 1, `templateIds` can be a partial list while `error` is set.

Gate fetches on `templateIdsError` as well (e.g. `isLoadingSystemJobExclusionIds: isLoadingTemplateIds || !!templateIdsError`, same for `useGetReportDetails`), and return `undefined` from `useJobTemplateIds` when `error` is set.

lgalis (changes requested):
Two must-fix items posted inline:

1. **Pagination loading gate** (`useJobTemplateIds.ts`) — partial template IDs can be used before all pages load (>200 templates).
2. **Error gating** (`useAutomationDashboardView.tsx`) — template API failure does not block report fetches; system jobs can leak back into metrics.

Please address before merge.

appuk:
Issue 1 — Pagination loading gate (useJobTemplateIds.ts):
  - Added isStillPaginating check: !!data?.at(-1)?.next — detects when SWR has
  loaded page 1 but more pages remain
  - templateIds now returns undefined while paginating or on error (not just
  during initial isLoading)
  - Returned isLoading is now isLoading || isStillPaginating, keeping the gate
  closed until all pages resolve
  - Added test that delays page 2 response and asserts templateIds stays
  undefined / isLoading stays true until all pages arrive

  Issue 2 — Error gating (useAutomationDashboardView.tsx):
  - Introduced isTemplateIdsUnavailable = isLoadingTemplateIds ||
  !!templateIdsError
  - This gates both useAutomationDashboardBaseView and useGetReportDetails —
  when /templates/ fails, no downstream report/details fetch fires unfiltered
  - Added two new tests: one verifying isLoadingSystemJobExclusionIds: true when
   loading, another when there's an error

frontend/awx/analytics/automation-dashboard/AutomationDashboard.tsx
```diff
@@ -233,7 +234,14 @@ export function AutomationDashboard() {
...
           >
-            {isLoading ? <LoadingState /> : dashboardContent}
+            {isLoading && <LoadingState />}
+            {!isLoading && templateIdsError && (
+              <EmptyStateError
+                titleProp={t('Unable to load dashboard data')}
+                message={templateIdsError.message}
+              />
+            )}
+            {!isLoading && !templateIdsError && dashboardContent}
```
lgalis: **Non-blocking / Jira:** The grid waits on collection `isLoading` and `templateIdsError`, but not on template-ID pagination. While IDs are still loading, users can see `dashboardContent` (empty placeholders) even though report/details fetches are correctly deferred.

Please track a follow-up to also treat `isLoadingTemplateIds` as a loading state here (or expose it from the view hook).

frontend/awx/analytics/automation-dashboard/views/useAutomationDashboardView.tsx
```diff
@@ -85,7 +109,12 @@ export function useAutomationDashboardView(options: {
     }
   }, [mainTableView, detailsResponse]);
 
-  const exportCsvBase = useExportCsv(toolbarFilters, filterState, QUERY_PARAMS);
+  const exportCsvBase = useExportCsv(
+    toolbarFilters,
+    filterState,
+    QUERY_PARAMS,
+    systemJobExclusionParams
```
lgalis, line 118: **Non-blocking / Jira:** Report/details honor `isTemplateIdsUnavailable`, but CSV export does not — `systemJobExclusionParams` is `undefined` until `allTemplateIds` resolves, so a quick export right after load could omit exclusion params.

Follow-up: disable export (or no-op with toast) until template IDs are ready; same condition as SWR gating.

lgalis (commented):
Looks good to merge for AAP-92027 — the template-ID injection and fetch gating should fix system jobs polluting the default dashboard view.

**Non-blocking follow-ups:** please open Jira ticket(s) (linked to AAP-92027 or a metrics-service epic) for the items called out inline below. None of these should block this frontend workaround.

Suggested split:
1. **UX / export parity** — gate toolbar/export (and optionally the dashboard body) until job template IDs finish loading.
2. **Edge case** — behavior when `/templates/` returns an empty list (exclusion params omitted today).
3. **Platform** — backend `exclude_system_jobs` (or equivalent) to replace all-`template=` query params and avoid URL length limits at scale.

frontend/awx/analytics/automation-dashboard/utils/queryString.ts
```diff
@@ -142,6 +144,20 @@ export function getPeriodFilterParam(filterState: IFilterState, filter: IToolbar
...
+ * Returns system-job-exclusion search params when no user template filter
+ * is active. Returns `undefined` when the user has an explicit template
+ * filter or when template IDs are not yet available.
+ */
+export function getSystemJobExclusionParams(
+  filterState: IFilterState | undefined,
+  templateIds: [string, string][] | undefined
+): [string, string][] | undefined {
+  const hasUserTemplateFilter = filterState?.template && filterState.template.length > 0;
+  if (hasUserTemplateFilter || !templateIds?.length) return undefined;
```
lgalis: **Non-blocking / Jira:** When `templateIds` is `[]`, this returns `undefined` and report/details/export run without any `template=` filter — system jobs can still appear on instances with no job templates.

Worth a ticket to document product behavior or handle explicitly (don’t fetch unfiltered metrics when the exclusion list is empty).

frontend/awx/analytics/automation-dashboard/common/useJobTemplateIds.ts
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * Returns template IDs as URLSearchParams entries (['template', '<id>'][])
+ * suitable for injection into API query strings to exclude system jobs from
+ * report results. Returns `undefined` while loading or on error so callers
+ * can gate fetches until data is ready.
+ *
+ * NOTE: Injecting all template IDs as query params may exceed URL length
+ * limits (8-20 KB) for very large deployments (thousands of templates).
+ * This is a known limitation of the frontend workaround, pending a
+ * backend-side fix (e.g. an `exclude_system_jobs` query parameter).
```
lgalis: **Non-blocking / Jira:** Good callout in the comment. Please ensure a metrics-service / platform ticket exists for server-side exclusion (e.g. `exclude_system_jobs`) so large controllers aren’t limited by query-string size. Link from AAP-92027 when closing the UI workaround.

lgalis (dismissed):
LGTM - a Jira should be opened for the issues  in my last comment, especially Export before template IDs finish loading

appuk:
Comment 1 — Loading gate for template ID pagination (AutomationDashboard.tsx):
  - Exposed isLoadingTemplateIds from the view hook and added it to IAutomationDashboardView
  - Dashboard now shows <LoadingState> while template IDs are paginating, preventing empty placeholder flash

  Comment 2 — Export timing risk (DashboardMainTableCard.tsx):
  - Added isLoadingTemplateIds to the export button's isDisabled condition, preventing CSV export before exclusion params are ready

  Comment 3 — Empty template list edge case (queryString.ts, useAutomationDashboardView.tsx):
  - getSystemJobExclusionParams now returns [] (not undefined) when templateIds is empty, distinguishing "loaded but empty" from "still loading"
  - Added noTemplatesExist flag that gates report/details fetches when no templates exist, preventing unfiltered system job data

  Comment 4 — Backend ticket reference (useJobTemplateIds.ts):
  - Updated the existing comment to reference AAP-92027 as the tracking ticket for the server-side exclude_system_jobs fix

lgalis (commented):
## Non-blocking review notes

Nice iteration on the loading/error/pagination gates since the earlier threads — the core fix looks solid. A few follow-ups I’d like tracked (not required for approval):

1. **Export vs report gating** — CSV export is disabled on `isLoadingTemplateIds` only; report/details use the broader `isTemplateIdsUnavailable` (error + zero templates). Worth aligning export with that same flag so a future UI change can’t export without exclusion params.
2. **Large deployments** — The documented URL-length risk is real for controllers with thousands of templates. Please link a metrics-service / platform ticket from the PR description (or AAP-92027 sub-task) for server-side exclusion so we can retire the query-string workaround.
3. **Zero job templates** — When `allTemplateIds` is `[]`, fetches stay gated and the dashboard can read like “no jobs” rather than “no templates.” Fine if product agrees; otherwise a dedicated empty state would help.
4. **Test timing** — A few tests use fixed `setTimeout` to assert “no fetch”; consider `waitFor` with a short timeout to reduce flake risk.
5. **Latency** — Every dashboard open waits for a full paginated `/templates/` sweep before metrics load. Acceptable for the workaround; worth calling out in release notes for large instances.

frontend/awx/analytics/automation-dashboard/components/DashboardMainTableCard.tsx
```diff
@@ -263,7 +264,13 @@ export function DashboardMainTableCard(props: IAutomationDashboardView) {
       exportType={'csv'}
       title={t('Export as CSV')}
       icon={ExportIcon}
-      isDisabled={loading || !(mainTableView.itemCount ?? 0) || !costState || !filtersValid}
+      isDisabled={
```
lgalis, line 267: **Non-blocking:** Export is disabled when `isLoadingTemplateIds` is true, but report/details use `isTemplateIdsUnavailable` (error + zero templates). Consider reusing the same gate (e.g. expose `canLoadReportMetrics` from the view hook) so CSV export cannot diverge from metrics fetches if the UI changes later.

lgalis (changes requested):
## Blocking items (before approval)

Please address the two inline comments on `useJobTemplateIds.ts`:

1. **Remove the new `eslint-disable`** — type or narrow `response.error` instead of adding a suppression in new code.
2. **Document backend exit** — link a tracked metrics-service ticket for server-side system-job exclusion (URL-length workaround is not sustainable on large controllers).

Non-blocking notes are in the earlier COMMENT review + export gating thread.

Thanks — happy to re-review quickly once these are in.

frontend/awx/analytics/automation-dashboard/common/useJobTemplateIds.ts
```diff
@@ -0,0 +1,74 @@
...
+    },
+    [baseUrl]
+  );
+
+  const response = useSWRInfinite<AwxItemsResponse<TemplateRecord>>(getKey, fetcher, {
+    revalidateOnFocus: false,
+    dedupingInterval: 60_000,
+  });
+  const { data, isLoading, size, setSize } = response;
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
lgalis: **Blocking (review policy):** New `eslint-disable-next-line @typescript-eslint/no-unsafe-assignment` on `response.error`. We treat new ESLint suppressions in the diff as a merge blocker — please type the SWR error (or use the same narrow cast/helper as elsewhere) instead of silencing the rule in this new file.

frontend/awx/analytics/automation-dashboard/common/useJobTemplateIds.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * This endpoint queries the AWX DB's main_jobtemplate table via an INNER JOIN,
+ * so system_job_templates (Cleanup Activity Stream, etc.) are excluded.
+ *
+ * Returns template IDs as URLSearchParams entries (['template', '<id>'][])
+ * suitable for injection into API query strings to exclude system jobs from
+ * report results. Returns `undefined` while loading or on error so callers
+ * can gate fetches until data is ready.
+ *
+ * NOTE: Injecting all template IDs as query params may exceed URL length
```
lgalis, line 32: **Blocking (release risk):** The query-string workaround can hit URL/proxy limits (~8 KB) on controllers with many templates (414 / failed report loads). The in-code note is good — before merge please link an explicit metrics-service / platform follow-up (sub-task of AAP-92027 or separate ticket) in the PR description so we do not ship this without a tracked path to server-side exclusion (e.g. `exclude_system_jobs`).

lgalis (approved):
Approve  - these will be addressed at a later time:
Findings (by severity)
Medium — zero templates: blocked fetches without loading or error UI
In useAutomationDashboardView, an empty template list sets isTemplateIdsUnavailable (and isLoadingSystemJobExclusionIds: true) so report/details never fetch, but the page only treats isLoadingTemplateIds and templateIdsError as blocking states:

const noTemplatesExist =
  !isLoadingTemplateIds && !templateIdsError && allTemplateIds?.length === 0;
const isTemplateIdsUnavailable = isLoadingTemplateIds || !!templateIdsError || noTemplatesExist;
After a successful empty /templates/ response, isLoadingTemplateIds is false and templateIdsError is unset, so AutomationDashboard renders dashboardContent while underlying hooks keep fetchUrl === null. Users can see an idle shell (no spinner, no error) instead of a clear “no templates” or “no data” state.

Suggestion: Drive the shell from isTemplateIdsUnavailable (or a dedicated flag), or allow fetch with [] exclusion only when product accepts that edge case, with an explicit empty state.

Medium — URL length / proxy limits (documented, still operational risk)
useJobTemplateIds documents that thousands of template=<id> params can exceed ~8–20 KB URL limits. The same params apply to table, details, and CSV export GETs. Large AWX instances risk 414, truncated queries, or proxy failures—not a logic bug, but a real deployment ceiling for this workaround.

Suggestion: Keep AAP-92027 (or follow-up) for a backend exclude_system_jobs (or equivalent) flag; call out max tested template count in the PR test plan.

Medium — metrics for templates not in the current ID list
Exclusion is “only jobs whose template ID is in today’s full template list.” Jobs tied to deleted templates (or stale metrics rows) can disappear from totals vs. an unfiltered query. System jobs are fixed; historical/orphan job data may be under-reported. Worth a one-line PR note so support knows this is expected for the UI workaround.

Low — pagination key vs. API next URL
useSWRInfinite advances with page=${pageIndex + 1}&page_size=200 rather than following the API’s next URL verbatim. Tests match that contract; if metrics-service ever changes pagination shape, this hook is the fragile point. Align with how useAwxGetAllPages (or similar) works elsewhere if that exists.

Low — non-Error SWR failures
useJobTemplateIds only surfaces response.error instanceof Error. Other failure shapes become error: undefined with templateIds: undefined, which can block fetches without hitting EmptyStateError. Uncommon if the fetcher always throws Error/RequestError.

Nit
Comment typo in useAutomationDashboardBaseView: “backed API” → “backend API”.
useJobTemplateIds.test.ts has two near-duplicate API-failure tests; could merge.

PR #3600, opened 2026-09-17
[AAP-93178] fix(data-editor): prevent YAML block-scalar growth loop on invalid input

acruzgon:
## Summary

Fixes a critical browser-crash bug in YAML/JSON data editor fields across all AAP product UIs (Controller, PAH, EDA, Gateway).

### Root Cause

When `valueToObject()` received a string that failed both JSON and YAML parsing (e.g. `  ---\n  a: b` — an indented document separator copied from a code block), it silently discarded the return value of the inner `catchError()` helper and returned the raw string as-is. `objectToString()` then called `jsyaml.dump(rawString)` which serialised the string as a YAML block scalar:

```
# Input:  '  ---\n  a: b'
# After one jsyaml.dump() call:
'|2-\n    ---\n    a: b\n'
```

On the next parse–dump round-trip `jsyaml.load()` returned the string with an extra level of indentation, and `jsyaml.dump()` produced an even longer block scalar. **Each iteration added 2 characters**, so editor content grew unboundedly (length 21 → 23 → 25 → ...) until the browser tab consumed all available memory and crashed.

### Fix

Replace the catch-and-discard pattern with a nested try/catch that **propagates the YAML parse error** when both JSON and YAML parsing fail. `handleChange()` already has a catch block that calls `setError()` to display a validation message without updating the form value, so the growth loop can never start.

Before the fix:
```ts
const catchError = () => {
  try { value = safeLoad(value); } catch (err) { return new Error(err.message); }
};
// ...
try { value = parseJSONPreservingLargeInts(value); } catch {
  catchError(); // ← return value silently discarded!
}
```

After the fix:
```ts
try { value = parseJSONPreservingLargeInts(value); } catch {
  try { value = safeLoad(value); } catch (err) {
    throw err; // ← propagate so handleChange shows validation error
  }
}
```

Affects: https://redhat.atlassian.net/browse/AAP-93178

## Type of Change

- [x] Bug fix

## Risk Analysis - REQUIRED

The change is confined to the `valueToObject()` function in `PageFormDataEditor.tsx`. It only alters behavior for inputs that are already unparseable as either JSON or YAML (previously: silently returned raw string; now: throws a parse error that the existing `handleChange` catch block already handles). All valid inputs are unaffected.

### Manual Testing

1. Open any YAML/JSON editor field (e.g. Extra Variables in a Job Template).
2. Paste or type `  ---\n  a: b` (2 spaces before `---`).
3. **Before fix**: text immediately starts "flying" — content grows on each render until the tab crashes.
4. **After fix**: an inline validation error is shown; content remains stable.

Same test applies to JSON mode with ` '\n` (space + unclosed single quote).

### Unit Tests

Three regression tests were added to `PageFormDataEditor.test.tsx`:

- `throws for indented YAML document separator (leading spaces before ---)`
- `throws for unclosed single-quoted scalar in JSON mode`
- `round-trip of invalid input does not grow unboundedly`

Assisted-by: Claude Sonnet 4.6 (Anthropic)


framework/PageForm/Inputs/PageFormDataEditor.tsx
```diff
@@ -535,10 +558,30 @@ export function valueToObject(
...
+          throw new YAMLException('invalid or incomplete YAML/JSON content');
+        }
+        value = parsed as object;
+      } catch (err) {
+        // Both JSON and YAML parsing failed. Throw so the caller (handleChange)
+        // treats this as a validation error rather than silently returning the
+        // raw string. Returning the raw string would cause jsyaml.dump() to
+        // serialise it as a YAML block scalar that grows unboundedly on each
+        // parse-dump round-trip, crashing the browser tab (AAP-93178).
+        if (err instanceof Error || err instanceof YAMLException) {
```
lgalis, line 683: **Blocking:** Throwing here is correct for `handleChange`, but other call sites still invoke `valueToObject` unguarded:

- `getDisplayValue` (~134) and the `useLayoutEffect` / `watchValue` effects that sync `dataEditorValue` (~168–195)
- `onDrop` (~149), `handleCopy` / `handleDownload` (~142–166), undo/reset (~342–363)

If RHF already holds unparseable text (API/defaultValues) or the user drops an invalid file, mount/sync can throw and white-screen the page. Please reuse `formatDetailCodeEditorValue` (or shared try/catch) for **display-only** paths and surface errors for drop/copy failures.

framework/PageForm/Inputs/PageFormDataEditor.tsx
```diff
@@ -362,7 +376,27 @@ export function PageFormDataEditor<
...
+
+          type ValidateRecord = Record<
+            string,
+            Validate<PathValue<TFieldValues, TFieldName>, TFieldValues>
+          >;
+
+          if (!props.validate) return { parseFormat } as ValidateRecord;
+          if (typeof props.validate === 'function')
+            return { parseFormat, custom: props.validate } as ValidateRecord;
+          return { parseFormat, ...props.validate } as ValidateRecord;
```
lgalis: Nit: `{ parseFormat, ...props.validate }` lets a consumer supply a `parseFormat` key and override the submit-blocking rule. Safer order: `{ ...props.validate, parseFormat }`.

framework/PageForm/Inputs/PageFormDataEditor.tsx
```diff
@@ -535,10 +558,30 @@ export function valueToObject(
...
     try {
       value = parseJSONPreservingLargeInts(value) as object;
     } catch {
-      catchError();
+      // JSON parsing failed — try YAML
+      try {
+        const parsed = safeLoad(rawString);
+        // Whitespace-only strings (e.g. a lone space while typing in JSON mode) load as
+        // `undefined` but must not commit as empty extra_vars and wipe saved content.
+        if (parsed === undefined && rawString !== '') {
```
lgalis: Good guard for YAML `undefined` (`' '`, `'\n'`). js-yaml also loads some whitespace as `null` (e.g. `'\t'`, `'---'`), which can still normalize to `''` via `objectToString` and wipe saved vars — consider rejecting `parsed == null && rawString.trim() !== ''` too.

framework/PageDetails/PageDetailCodeEditor.tsx
```diff
@@ -15,6 +15,19 @@ import { DataEditorLanguages } from '../components/DataEditor';
 import { useClipboard } from '../hooks/useClipboard';
 import { PageDetail } from './PageDetail';
 
+/** Read-only display: keep original text when value is not structured YAML/JSON (e.g. job stdout). */
+export function formatDetailCodeEditorValue(
```
lgalis, line 18: Nice extraction. Worth reusing (or moving to a shared helper beside `valueToObject`) for `PageFormDataEditor` display sync / drop / copy so every caller gets the same fail-safe behavior.

framework/PageForm/Inputs/PageFormDataEditor.test.tsx
```diff
@@ -394,4 +452,194 @@ debug_mode: true         # Enable debugging`;
...
+
+  // Regression test for AAP-93178.
+  //
+  // The DataEditor mock renders a controlled textarea (value={props.value}).
+  // When handleChange() throws on an invalid YAML string, the form value does
+  // not update, so the textarea resets to its previous value after every
+  // re-render.  This prevents userEvent.type from accumulating multi-character
+  // text, making fireEvent.change the only reliable way to fire the onChange
+  // callback with the full invalid string in one shot.
+  test('should display an inline parse error when invalid YAML is entered (AAP-93178)', async () => {
```
lgalis, line 578: Coverage gap: tests cover typing invalid YAML, but not **mounting** with invalid preloaded values. Suggest `defaultValue={{ vars: '  ---\n  a: b' }}` — expect inline error and no throw (same class of bug as job stdout in `PageDetailCodeEditor`).

playwright/tests/integration/automation-execution/templates/job-template-validation.spec.ts
```diff
@@ -121,4 +122,53 @@ test.describe('Job Template Form - Validation', () => {
...
+      // Blur the editor to trigger the watchValue update cycle.
+      await page.getByPlaceholder('Enter job template name').click();
+
+      // The editor must NOT show the YAML block scalar prefix `|2-`.
+      // That prefix was the signature of the growing-content bug — each
+      // parse → dump cycle produced `|2-\n    ---\n    a: b\n` with one more
+      // level of indentation, eventually crashing the tab.
+      // After the fix the editor reverts to its default (empty) value on blur.
+      const editorLines = page.getByTestId('extra-vars').locator('.view-lines');
+      await expect(editorLines).not.toContainText('|2-', { timeout: 3000 });
```
lgalis, line 172: Useful regression, but `not.toContainText('|2-')` also passes when the raw invalid string is shown (no growth). Stronger checks: visible inline validation error after blur, and Save blocked until content is valid.

lgalis (dismissed):
Thanks for the thorough AAP-93178 fix — throwing from `valueToObject` and gating submit with `parseFormat` stops the growth loop on the typing path, and `formatDetailCodeEditorValue` is the right read-only pattern.

Requesting changes before merge: several non-`handleChange` paths still call `valueToObject` without try/catch (see inline on `valueToObject` / `onDrop`-related note), so forms that **load** unparseable values can throw during effects instead of showing a field error.

lgalis (commented):
Review comments added.

framework/PageForm/Inputs/PageFormDataEditor.tsx
acruzgon: Addressed in 30a6e637d: added `formatEditorDisplayValue` / `getValueToObjectParseError`, wired display sync (effects, undo/reset), and drop/copy/download with toasts + field errors where appropriate. `handleChange` remains the strict throw path.

framework/PageForm/Inputs/PageFormDataEditor.tsx
acruzgon: Good catch — swapped to `{ ...props.validate, parseFormat }` (and `custom` + `parseFormat` for function validators) so consumers cannot override `parseFormat`.

framework/PageForm/Inputs/PageFormDataEditor.tsx
acruzgon: Agreed — now reject `(parsed === undefined || parsed === null) && rawString.length > 0` with tests for `---` and `\t`.

framework/PageDetails/PageDetailCodeEditor.tsx
acruzgon: Moved shared logic to `formatEditorDisplayValue` in `PageFormDataEditor.tsx`; `PageDetailCodeEditor` delegates to it.

framework/PageForm/Inputs/PageFormDataEditor.test.tsx
acruzgon: Added mount test with `defaultValue={{ vars: '  ---\\n  a: b' }}` expecting inline error (no throw).

playwright/tests/integration/automation-execution/templates/job-template-validation.spec.ts
acruzgon: Strengthened E2E: assert inline validation message after blur, fill template name, click Create, and confirm we stay on the form with the error still visible (not just `|2-` absent).

keithjgrant (dismissed):

lgalis (approved):

lgalis (dismissed):
1. valueToObject() now throws — external validate rules may throw instead of returning messages

valueToObject no longer returns the raw string on parse failure; it throws. Any field validate that calls valueToObject without try/catch can blow up RHF validation instead of returning a friendly string.

Concrete call site on this branch:

// InventoryForm.tsx — constructed inventory `source_vars`
validate={(item) => {
  const obj = valueToObject(item) as { ... };
  // ...
}}
While the user is editing invalid YAML, this can throw during validate rather than surfacing a controlled message. Suggestion: use getValueToObjectParseError() in custom validators, or wrap valueToObject in try/catch and return the parse message.

2. Risk rating understates blast radius

The PR marks Low, but PageFormDataEditor is shared across AWX, Hub, EDA, and Gateway. Behavior changes for:

invalid parse handling (throw vs silent string),
whitespace / --- empty markers,
copy/download/drop on invalid content,
new parseFormat rule blocking submit.
That is Medium cross-cutting framework impact even if the diff is mostly one module.

PR #3610, opened 2026-09-18
ci: enforce Knip unused-code checks

nixocio:
## Summary

- Make the Knip report a blocking PR check instead of an advisory warning.
- Enable `pipefail` so a failing Knip command cannot be masked by `tee`.
- Keep the report summary and artifact available when findings fail the job.

## Dependencies

Merge after #3608 and #3609 so the cleanup and false-positive configuration changes are present.

## Risk Analysis - REQUIRED

- [x] **Medium** — Changes the PR CI merge gate by enforcing the existing Knip report.


.github/workflows/pull-request.yml
```diff
@@ -80,15 +80,13 @@ jobs:
...
       - run: npm ci
         if: steps.cache.outputs.cache-hit != 'true'
-      # Advisory: reports unused files/deps, does not block merge.
-      # Remove continue-on-error when the report is small enough to enforce.
+      # Enforced: actionable unused-code findings block the required check.
       - name: knip report
-        continue-on-error: true
-        run: npm run knip | tee knip-report.txt
+        run: |
+          set -o pipefail
```
nicklemmon: AI-assisted review note: `set -o pipefail` is intentional here because the pipeline ends with `tee`. Without it, a non-zero exit from `npm run knip` (for example, findings) could be masked by `tee` successfully writing `knip-report.txt`, so the blocking check would incorrectly pass. GitHub Actions uses `bash -e {0}` when the shell is unspecified; specifying `shell: bash` would enable `pipefail` automatically. Could you add a short comment in the workflow explaining this reasoning (and, if retaining the explicit `set`, why it is preferred over `shell: bash`)?

This review note was prepared with AI assistance.

nicklemmon: (I know AI wrote this but I swear this was my idea!) - a human in the loop

nixocio: Added the inline comment explaining that `tee` can mask Knip failures and why `set -o pipefail` remains explicit. Pushed in `05a0455f3`.

.github/workflows/pull-request.yml
```diff
@@ -80,15 +80,13 @@ jobs:
...
+          npm run knip | tee knip-report.txt
       # Surface the report on the run's Summary page so findings are visible
-      # without expanding logs or downloading the artifact. Advisory only.
-      # continue-on-error so the summary can never fail the job (which would
-      # trip the "All checks passed" gate). Advisory only.
+      # without expanding logs or downloading the artifact.
```
nicklemmon: AI-assisted review note: the `|| echo "- No findings 🎉"` fallback in the summary below will not run when `grep` finds no matches. This workflow uses the default `bash -e` shell without `pipefail`, and `sed` is the last command in the pipeline, so the pipeline still exits 0. On a clean Knip report, the summary would therefore show a blank findings section rather than `No findings`. Could we test the `grep` result separately or enable `pipefail` for this summary pipeline?

nicklemmon (dismissed):

.github/workflows/pull-request.yml
nixocio: Addressed in `1a8f72134`: the summary now captures `grep` in an explicit `if`, so a clean report reliably prints `No findings` even with the default `bash -e` shell.

PR #3613, opened 2026-09-18
chore(types): pilot implicit override checks

nixocio:
## Summary

Enable `noImplicitOverride` for the `framework` TypeScript project and add the
required `override` modifiers to the existing `ErrorBoundary` members. This is
the next small compiler-safety pilot in the staged TypeScript rollout.

## Type of Change

- [ ] Bug fix
- [x] Enhancement
- [ ] Tests
- [ ] Documentation
- [ ] Other (please specify)

## Risk Analysis - REQUIRED

## Dependencies

This pilot is independent of the repository-wide unused-locals and
unused-parameters changes. It does not enable additional compiler flags
monorepo-wide.

## Testing

- `npx tsc --noEmit --pretty false -p framework/tsconfig.json`
- `npx eslint framework/components/ErrorBoundary.tsx`
- `npx prettier --check framework/tsconfig.json framework/components/ErrorBoundary.tsx`
- `npx tsc --showConfig -p framework/tsconfig.json` confirms the option is enabled.

## TypeScript references

- [`noImplicitOverride`](https://www.typescriptlang.org/tsconfig/noImplicitOverride.html)

No UI behavior changed, so Playwright is not required.


h-kataria (approved):

framework/components/ErrorBoundary.tsx
```diff
@@ -19,12 +19,12 @@ export class ErrorBoundary extends Component<Props, State> {
...
   }
 
-  public componentDidCatch(error: Error, errorInfo: ErrorInfo) {
+  public override componentDidCatch(error: Error, errorInfo: ErrorInfo) {
     // eslint-disable-next-line no-console
     console.error('Uncaught error:', error, errorInfo);
   }
 
-  public render() {
+  public override render() {
```
nicklemmon, line 27: This doesn't come up very often because we don't use classes/class components these days!

framework/tsconfig.json
```diff
@@ -1,5 +1,8 @@
 {
   "extends": "../tsconfig.json",
```
nicklemmon, line 2: This could be enabled in the root config, though I don't think it will have much of an impact due to the lack of use of classes

framework/components/ErrorBoundary.tsx
nixocio: Agreed — `ErrorBoundary` is the only hand-written `Component` subclass in the framework package, so the blast radius here is tiny. The `override` annotations are what `noImplicitOverride` requires for the remaining lifecycle members; no broader class-component migration is in scope for this pilot.

framework/tsconfig.json
nixocio: Good point — we could enable `noImplicitOverride` in the root `tsconfig.json`, and you're right that class usage is sparse outside `framework`.

This PR intentionally keeps the pilot scoped to `framework/tsconfig.json`, matching the staged TypeScript compiler rollout (#3569, #3570, #3612): enable one flag in one workspace, fix fallout, then widen. Here that means adding `override` on the existing `ErrorBoundary` members and confirming `tsc -p framework` is clean before touching the monorepo root.

If you'd like it flipped on at the root next, I'm happy to do that in a follow-up — given limited class usage it should be low impact.

nixocio:
Closing: deferring noImplicitOverride pilot — the only React class component in the repo is ErrorBoundary; the codebase is hooks-first, so enabling this compiler flag standalone is not worth the churn per review feedback. No behavior change needed.

PR #3617, opened 2026-09-21
[AAP-82489] fix: show detailed error message on template launch failure

keithjgrant:
## Summary

When launching a Job/Workflow Job Template through the "Prompt on Launch"
wizard, a rejected launch (e.g. swapping a prompted credential for one of a
different type) showed only the generic HTTP reason phrase — "Bad Request" —
in the failure toast. The wizard's submit handler caught the `RequestError`
itself and displayed `err.message` only, discarding the API's parsed JSON
body (e.g. `{"credentials": ["Cannot assign a Credential of kind \`token\`", ...]}`)
that contains the actual validation detail.

This fixes the toast to route the error through the existing
`useAwxErrorMessageParser()` / `awxErrorAdapter`, so the specific API
validation message is shown instead of the generic status text.

### Root cause

Per AAP-82489, changing a credential to one of a different type at launch is
**intentional backend behavior** (documented since 2018 in
`docs/prompting.md` — a same-type replacement is required so a relaunch uses
the up-to-date credential). Engineering confirmed this in the Jira comments,
and the customer's original complaint about the block itself was later
withdrawn. The remaining, still-valid complaint was that the UI didn't
surface *why* the launch failed — just "Bad Request" — even though the API
already returns a specific, actionable message.

`frontend/awx/resources/templates/TemplatePage/TemplateLaunchWizard.tsx`'s
`handleSubmit` wraps the launch `postRequest` in its own `try/catch` and
shows a toast built from `err.message` alone. Since it catches locally, the
error never reaches `PageWizardProvider`'s own `submitError` state and its
`RequestErrorAlert`, and it also never runs the (JSON-aware) error adapter.

## Type of Change

- [x] Bug fix

## Risk Analysis - REQUIRED

- [x] **Low** — Narrowly scoped: one launch-failure toast handler in one wizard.

### Manual Testing

1. Create a Job Template with "Prompt on launch" enabled for Credentials, with a default credential set.
2. Launch the template and, on the Credentials prompt, swap the credential for one of a different credential type.
3. Submit the wizard.
4. Before this fix: the failure toast shows only "Bad Request". After this fix: the toast shows the specific validation message returned by the API (e.g. `Cannot assign a Credential of kind \`token\`.`).

### Regression test

Added a test to `TemplateLaunchWizard.test.tsx`: mocks the launch POST returning a 400
with a `credentials` validation array (the exact body from the Jira report)
and asserts the toast's `children` is the specific message, not the generic
status text.

**Mutation testing:** the workflow's automated mutation-test script errored
on reverse-applying the fix patch (unrelated tooling issue), so I verified
manually — checked out the pre-fix version of the source file and confirmed
the new test fails, then restored the fix and confirmed it (and the other 19
tests in the file) pass.

**Similar bug detection:** the same pattern (`error instanceof Error &&
error.message` swallowing a `RequestError`'s JSON body in a toast) appears
in ~29 other call sites across awx/eda/hub, e.g.
`frontend/awx/resources/templates/hooks/useLaunchTemplate.tsx:52` and
`frontend/awx/resources/templates/TemplatePage/RelaunchTemplateWithPasswords.tsx:65`.
Out of scope for this fix — noting for future cleanup.

Screenshot 2026-09-21 at 10 18 10 AM

Assisted-by: Claude Code / Sonnet 5 (Anthropic)


keithjgrant:
only 5 lines changed. that 80% coverage is because one of those lines has a ternary that is only half exercised in test. I'm not sure it's worth adding a whole test for that last line (it's just a fallback to a generic text string)

frontend/awx/resources/templates/TemplatePage/TemplateLaunchWizard.tsx
```diff
@@ -195,10 +199,12 @@ export function LaunchTemplate({ jobType }: { jobType: string }) {
...
         }
       } catch (err) {
+        const failureTitle = t('Failure to launch');
         alertToaster.addAlert({
           variant: 'danger',
-          title: t('Failure to launch'),
-          children: err instanceof Error && err.message,
+          title: failureTitle,
+          children:
+            err instanceof Error ? parseErrorMessage(err, failureTitle).message : failureTitle,
```
lgalis: `useAwxErrorMessageParser().message` only contains API text when `parsedErrors.length === 1`. For multi-field 400s (e.g. `{ credentials: [...], inventory: [...] }`), `.message` is the second argument — here `failureTitle` — so the toast body becomes "Failure to launch" and drops every validation string.

EDA already renders all parsed lines in toasts, e.g. `useCopyCredential.tsx`:
```tsx
children: <>{errorResults.parsedErrors.map((e) => e.message)}</>,
```

Suggest using `parsedErrors` for `children` (with `failureTitle` only as the alert title), and fall back to `err.message` when `parsedErrors` is empty (e.g. empty `{}` JSON body) so generic 500s keep the HTTP reason instead of duplicating the title.

frontend/awx/resources/templates/TemplatePage/TemplateLaunchWizard.test.tsx
```diff
@@ -523,6 +523,50 @@ describe('TemplateLaunchWizard', () => {
...
+    it(
+      'should show the API validation error when a credential type mismatch is rejected',
+      { timeout: 15000 },
+      async () => {
+        mockAddAlert.mockClear();
+        server.use(
+          http.post(awxAPI`/job_templates/1/launch/`, () =>
+            HttpResponse.json(
+              {
+                credentials: [
```
lgalis: The mock includes two `credentials[]` strings from the Jira report, but the assertion only checks the first. That matches current `awxErrorAdapter` behavior (it keeps `value[0]` for field arrays), so the second "without replacement" line is never shown.

Worth either:
- trimming the mock to a single string so the test matches what we actually surface, or
- if we switch the toast to `parsedErrors` and later join array values in the adapter, extending the assertion.

Optional follow-up: a case with two field keys (`credentials` + `inventory`) would catch the `.message` multi-error regression above.

lgalis (commented):
Thanks for the focused fix and the AAP-82489 regression test — routing launch failures through the AWX adapter is the right direction.

Left two inline notes on items worth addressing before merge: use `parsedErrors` in the toast (not `.message`) so multi-field 400s don't collapse to the title, and align the test mock with what we actually display for multi-string `credentials` arrays.

frontend/awx/resources/templates/TemplatePage/TemplateLaunchWizard.tsx
keithjgrant: Good catch — fixed in e553e65a7: the toast now renders every parsed error message (one per line, matching PageWizardBody's RequestErrorAlert convention) instead of collapsing to the title when there's more than one field error, with a fallback to err.message when nothing was parsed. Also trimmed the existing test's mock to match what awxErrorAdapter actually surfaces (first message only) and added a regression test for the multi-field case. Thanks for the catch!

frontend/awx/resources/templates/TemplatePage/TemplateLaunchWizard.test.tsx
keithjgrant: Trimmed the mock to a single string — matches what awxErrorAdapter actually surfaces since it still only keeps the first array element per field. Also added a new test for two rejected fields (credentials + inventory) to directly cover the multi-error regression from the other comment.

lgalis (approved):
LGTM

